PR #3, opened 2026-05-18
ci: add SSH deploy workflow for production

emanueleDiVizio:
## Summary
- Add `.github/workflows/deploy.yml` that runs on push to `main` (and manual `workflow_dispatch`)
- It SSHes to the production host using the `DEPLOY_SSH_KEY` / `DEPLOY_HOST` / `DEPLOY_USER` / `DEPLOY_PATH` repo secrets, then runs `./deploy.sh` in `$DEPLOY_PATH`
- `concurrency: production-deploy` so overlapping pushes queue rather than race

## Why
Wire continuous deploy to the production host so merges to `main` go straight to the live stack via the existing `deploy.sh` smart-rebuild script. Host/path/user are all opaque secrets.

## Test plan
- [ ] Merge this PR → GH Actions "Deploy to production" workflow runs
- [ ] SSH step connects and `deploy.sh` reports either "No new commits" or a successful build
- [ ] Manual trigger via Actions tab also works (`workflow_dispatch`)

## Build artifacts
- [ ] Build macOS .dmg
- [ ] Build Windows .exe


src/platform/web.ts
```diff
@@ -609,12 +609,26 @@ class WebServerApi implements IServerApi {
...
-    if (this.ws) {
-      this.ws.close();
-      this.ws = null;
-    }
+    const ws = this.ws;
+    this.ws = null;
     this.relayUrl = null;
     this.serverPassword = null;
+    if (!ws || ws.readyState === WebSocket.CLOSED) return;
+    // Wait for the actual close event before resolving. Resolving on
```
fedepoi, line 617: useless comment remove plis

src/platform/web.ts
```diff
@@ -609,12 +609,26 @@ class WebServerApi implements IServerApi {
...
+    // ws.close() alone races against the server still cleaning up the
+    // session, so Settings "Save & Reconnect" (disconnect → connect)
+    // would race the new authenticate against the previous session and
+    // forge-server rejected it as a duplicate.
+    await new Promise<void>((resolve) => {
+      const done = () => {
+        ws.removeEventListener("close", done);
```
fedepoi, line 624: magic strings here?

fedepoi (approved):

src/platform/web.ts
emanueleDiVizio: i think it's convention but can have a look

src/platform/web.ts
emanueleDiVizio: will do in follow up pr so we can test the autodeploy

PR #12, opened 2026-05-19
ci(deploy): trigger deploy via github actions over ssh

fedepoi:
## Summary

- Add Discord bot notifications to the deploy workflow (`POST /channels/{id}/messages`), with a green/red embed posted in the community channel after every deploy.
- Capture the deploy log on the runner so the failure path posts useful context (not just a workflow-run link).
- Slim `deploy.sh`'s header comment + clean up the dead n8n / private-repo references — public-repo aware. Keep main's optional `GITHUB_TOKEN` PAT block (no-op when unset; useful purely as a rate-limit avoidance on the server's `git pull`).
- Document the bot + secrets setup in `docs/DEPLOY.md`.

Built on top of #3 (Emanuele's SSH deploy workflow) — adopts `DEPLOY_HOST` / `DEPLOY_USER` / `DEPLOY_SSH_KEY` / `DEPLOY_PATH` / `production-deploy` concurrency group as-is.

## Why

The repo recently went public and the old n8n-webhook deploy trigger no longer fits. #3 already migrated to GitHub Actions over SSH; this PR adds the missing piece — visible deploy notifications in the community Discord channel — and uses the project's existing bot identity rather than a channel webhook.

Posting from the workflow runner (not from `deploy.sh` on the server) means a failed SSH still surfaces as a red embed instead of silence.

## Test plan

One-time secret setup (operator adds these in GitHub repo → Settings → Secrets → Actions):

- [ ] `DEPLOY_HOST`, `DEPLOY_USER`, `DEPLOY_PATH`, `DEPLOY_SSH_KEY` — already needed by #3, may already exist.
- [ ] `DISCORD_BOT_TOKEN` — bot token from the Discord Developer Portal.
- [ ] `DISCORD_DEPLOY_CHANNEL_ID` — right-click the target channel (Developer Mode on) → Copy Channel ID.
- [ ] In the target channel, confirm the bot's role has **Send Messages** + **Embed Links**.

End-to-end verification after merge:

- [ ] Manually dispatch (`Actions → Deploy to production → Run workflow → main`); green "Deploy complete" embed lands in Discord posted by the bot.
- [ ] Push a small change to a Web file (`src/...`); embed shows `Changed: Web` and `Services rebuilt: - manabrew`.
- [ ] Push a no-op change; deploy.sh prints "No new commits / No Java/Rust/infra changes" and the Action still posts a (success) embed with that line.
[...]

PR #13, opened 2026-05-19
ci(deploy): debug harness — pull_request trigger + verbose SSH logs

fedepoi:
## Summary

Temporary harness so we can debug the SSH-handshake failure from #12 (`client_loop: send disconnect: Broken pipe`, exit 255) without spamming the community Discord channel or risking a real deploy on every push.

- Adds `pull_request` trigger so this PR's pushes fire the workflow.
- Adds a diagnostic block ahead of `./deploy.sh`:
  - runner egress IP + `ssh -V`
  - validate the `DEPLOY_SSH_KEY` parses with `ssh-keygen -y` (mangled-newline secrets are the #1 cause of immediate disconnects — this catches them locally)
  - probe TCP reachability of `host:port`
  - `ssh-keyscan -v` with explicit port + 5s timeout
  - `ssh -v` with a trivial remote command (`whoami; hostname; pwd; ls -la $DEPLOY_PATH/deploy.sh`)
- On `pull_request` events the workflow **stops after diagnostics** — `deploy.sh` is never invoked, so production is safe.
- Discord notifications are suppressed on PR events.
- Optional new secret `DEPLOY_SSH_PORT` (defaults to 22) in case sshd listens elsewhere.

## Why

The current deploy job dies in 0s with `client_loop: send disconnect: Broken pipe`, which is the local SSH client reporting the server closed the connection before any output flowed. The existing workflow doesn't expose enough to tell whether it's:

- a mangled key in `DEPLOY_SSH_KEY`,
- wrong port in `ssh-keyscan` (host-key verification mismatch under `StrictHostKeyChecking=yes`),
- `~/.ssh` perms on the server,
- `fail2ban` or `AllowUsers` in `sshd_config`.

This PR adds enough signal in one workflow run to nail it down.

## Test plan

- [ ] Workflow fires on this PR. (If it doesn't trigger, the test fails before we even start — that's a signal too.)
- [ ] `Validate SSH private key format` group prints a public key and "private key parses". If it says `FAILED to parse`, the secret has corrupted newlines — fix by re-pasting the key with hard line breaks preserved.
- [ ] `TCP reachability` group reports the port open. If "closed/filtered", `DEPLOY_HOST` is wrong or a firewall is blocking the runner.
- [ ] `ssh-keyscan` writes ≥1 line to known_hosts.
- [ ] `Verbose SSH connect` succeeds; the trailing `DIAGNOSTICS_OK` appears in the log.
- [ ] No Discord embed appears in the community channel (PR events are gated out).

Once it's green, we revert the diagnostic / PR-trigger bits and merge #12 normally. If `DEPLOY_SSH_PORT` ends up needed, that part stays.

## Build artifacts


fedepoi:
test succeeded closing this

PR #23, opened 2026-05-19
fix(web): support N-player multiplayer in WASM engine

### LINKED ISSUE #10: 4-player Commander on web shows a 1v1 board for 2 players and a blank screen for the other 2
okay so there's a brekaage in prod. if i make a 4-player commander room on manabrew.app and we start the game, 2 players see what looks like a 1v1 battlefield and the other 2 see literally nothing — blank waiting screen forever.

i dug into it. the wasm multiplayer engine is hardcoded to 2 players. when the 4 of us start a game, the host's worker silently drops decks 3 and 4 and runs a 2-player game with decks 1 and 2. that's exactly why the symptom looks the way it does:

- host (player-0) + player-1 are inside the engine → they see a 1v1 board. player-1 gets it via the one remote-prompt SAB the host relays.
- player-2 and player-3 don't exist as far as the engine is concerned. no `game:remote_prompt` ever arrives with `forPlayer === myPlayerSlot`, and `useGameEventListeners.ts:104` only patches metadata onto an *existing* gameView, so they stay on the waiting screen forever.

the 4 places that pin this at 2:

1. `forge-engine/crates/forge-wasm/src/wasm_api.rs:293` — `run_multiplayer_game` takes `player0_deck_json` + `player1_deck_json`, builds 2 `PromptAgent`s for `PlayerId(0)` / `PlayerId(1)`. that's it.
2. `forge-engine/crates/forge-wasm/src/game_runner.rs:120` — `WasmGame::new` accepts `human_deck` + `ai_deck`, builds a 2-player `GameState`, `GameLoop::new(2)` is literal.
3. `src/workers/game-engine.worker.ts:380` — worker reads `decks[0]` / `decks[1]` and errors if they're not both there. decks 2 and 3 are silently dropped before they ever reach wasm.
4. `src/platform/web.ts:395` — `setRemotePlayerSlot` flips between `player-0` / `player-1`. no slot for player-2 / player-3.

## "this used to work" — i was wrong about that

i thought we'd had 4-player commander working on web before. git says no:

- the original web multiplayer commit was explicit about it: `838bb1bf2 feat(web): Enable WASM multiplayer hosting with dual SAB relay` (#351, apr 3). the commit body literally says *"Currently supports 2 players (1 local + 1 remote)."*
- the tauri-native side got generalized to N players in `4b5cdab71 Add Java Forge backend bridge` (#407, apr 21) — that's the `for i in 0..num_players` loop in `src-tauri/src/engine_backend/rust_backend.rs:136` that works correctly today on the desktop build.
- the wasm side was never brought up to match.

so: not a regression. just a long-standing TODO that nobody noticed because we hadn't let people make 4-player commander rooms on the web until recently. the lobby/room UI shows all 4 seats so it *looks* like it's starting, and the host's view of a 1v1 looks superficially fine.

i must've been remembering desktop commander games, or just imagining it. either way, prod has never been able to do this.

## fix sketch

extend the wasm path to N players, mirror what `rust_backend::run_multiplayer_game` already does on the desktop side:

1. `WasmGame::new` takes `&[Vec<DeckCard>]` instead of the 2-tuple. pass `num_players` to `GameLoop::new`.
2. `run_multiplayer_game` (wasm) takes `Vec<WireDeck>` + per-player commander names + one SAB per non-host player (i'd keep them separate rather than multiplexing one SAB — `Atomics.wait` is already per-SAB and the parallel-block protocol stays simple).
3. worker: allocate N−1 remote SABs, post each one to the main thread tagged with its `player-N`. main thread routes each remote SAB to the WebSocket relay.
4. `WebGameApi`: kill `setRemotePlayerSlot(string | null)` and replace it with a slot list. `respond` and prompt routing both need to know *all* non-host slots, not just one.
5. drop the `decks[0] / decks[1]` guard in the worker.

## can we just have one game loop?

asked claude to think about whether we can dedup wasm and tauri into a single source of truth. answer is yes for *most* of it, but with a caveat.

today we have two near-identical setup-and-run loops:

- `src-tauri/src/engine_backend/rust_backend.rs:136` (`run_multiplayer_game`) — N-player, uses `TauriTransport` (mpsc)
- `forge-engine/crates/forge-wasm/src/wasm_api.rs:293` (`run_multiplayer_game`) — 2-player, uses `WasmTransport` (SharedArrayBuffer)

both do the same dance: prepare players from decks, build the `GameState`, register the token DB onto the `GameLoop`, build one `PromptAgent` per slot, run the loop, emit final per-player views.

the transports themselves can't be unified, and i don't want to pretend otherwise: `MpscTransport` is `std::sync::mpsc` which can't cross a wasm worker boundary, and `WasmTransport` is `Atomics.wait` on SABs which is web-only. that's a real hardware-level divergence.

but the setup loop *can* be unified. proposal: add something like `forge_game_runtime::host_multiplayer_game` taking

- `decks: &[Vec<CardIdentity>]`
- `commander_names: &[Option<String>]`
- `starting_life: i32`
- `engine_player_index: usize`
- `agent_factory: impl FnMut(PlayerId, bool /* is_local */) -> Box<dyn PlayerAgent>`

and owning the boilerplate (prepare players, register tokens, build agents via the factory, run the loop, emit per-player game-over views). then `rust_backend.rs` and `wasm_api.rs` each shrink to ~20 lines that just supply their transport-specific factory.

this also pulls the duplicated game-over emission (lines 233-263 in rust_backend, lines 387-397 in wasm_api) into one place. that's the most likely place for the two backends to drift out of sync as we touch the engine, so it'd genuinely be worth doing.

caveat: wasm still has setup work that doesn't fit the factory pattern — allocating N SABs, posting them to the main thread, and the AI-in-same-worker thing (`WasmAiTransport` is sync; `MpscTransport` AI uses a separate responder thread). but the *game-state setup + game-loop run + game-over emission* part is genuinely shared and worth pulling out.

## priority

this is prod-blocking for 4-player commander on web. workaround until fixed: clamp `max_players` to 2 on the web client when format is commander, or just hide commander from the web format picker. desktop build keeps working as-is.

emanueleDiVizio:
## Summary

- Extract the multiplayer setup-and-run dance into `forge_game_runtime::host_runtime::run_hosted_multiplayer_game`; Tauri and WASM both call it via transport/token-DB/final-view closures.
- WASM `run_multiplayer_game` becomes N-player — takes `Vec<WireDeck>` + matching commander names + an `Array<SharedArrayBuffer>` of size `num_players - 1`.
- Worker allocates one remote SAB per non-host seat, tagged with its `player-N` slot; `WorkerBridge` keeps a slot-keyed map and routes incoming WebSocket responses by `fromPlayer`.

## Why

Fixes #10. The web path was hardcoded to 2 players in four places (`wasm_api.rs`, `game_runner.rs`, `game-engine.worker.ts`, `web.ts`) — so any 4-player commander room on manabrew.app silently ran a 2-player engine with the first two decks. Host + player-1 saw a 1v1 board, player-2 and player-3 stayed on the blank waiting screen.

Tauri's native path was already N-player (since PR fedepoi/bardidinaXmageUI#407) but its setup loop and WASM's were near-identical copies of the same logic — that's how WASM fell behind without anyone noticing. The shared `run_hosted_multiplayer_game` puts the game-state setup + token registration + agent construction + game loop run + per-player game-over emission in one place. Transports stay separate (mpsc vs SAB is a real hardware-level divergence across the worker boundary), but everything else lives here exactly once.

WASM also now emits a final GameOver prompt per seat post-loop, matching Tauri — previously the WASM clients sometimes stalled on the last mid-game prompt instead of transitioning to the game-over screen.

## Review feedback addressed

Each item below is its own commit on the branch:

1. **Tauri discarded `HostedGameOutcome`** — now `let outcome = …` with an explicit `if outcome.aborted { return; }` so future post-game cleanup is gated correctly.
2. **`force_commander_by_name` ordering** — traced: `deck_to_identities` tags the `commanders` pile, `prepare_registered_player` routes it to the command zone, and `force_commander_by_name` early-returns if already present (idempotent, no double-set). It exists as a fallback for the lobby supplying the commander out-of-band. Documented on the function + both call sites.
3. **`pollForRemotePromptsSeat` one-frame race** — added a per-seat `cancelled` flag that `terminate()` flips before clearing the map.
4. **`ensureRemoteResponseListener` race** — listener now installed eagerly in the constructor; no-ops on an empty seat map. `terminate()` no longer tears it down (would break a 2nd game on the singleton bridge).
5. **RNG threading** — `run_hosted_multiplayer_game` now takes `rng: &mut StdRng`; both callers build it with `from_entropy()`. Added a `wasm-bindgen` runtime smoke test (`tests/rng_smoke.rs`) that constructs the RNG on wasm32 — passes under `wasm-pack test --node`. getrandom's `js` feature confirmed enabled. `forge-game-runtime` keeps `rand` (signature names `StdRng`, which isn't in `rand_core`).
6. **`FORGE_ENGINE_GAME_LOG` inverted convention** — documented (ON by default, set `0`/`false`/`off` to disable); behavior unchanged. Rename to `_DISABLE` left as a follow-up.
7. **Lost per-seat card-count logging** — restored as `cards per seat=[…]` in the WASM startup log.
8. **`max_turns` magic 5000** — pulled into `host_runtime::DEFAULT_MAX_TURNS`, referenced at both call sites.

Follow-ups noted, not done here: rename `FORGE_ENGINE_GAME_LOG` → `_DISABLE`; wire the wasm RNG test into CI; add `WorkerBridge` slot-routing unit tests.

## Test plan

- [x] `cargo build -p manabrew` clean
- [x] `cargo build -p forge-wasm --target wasm32-unknown-unknown` clean
- [x] `cargo clippy` on touched crates clean
- [x] `tsc --noEmit` + `eslint` on touched TS clean
- [x] `cargo fmt --check` on touched crates clean
- [x] `wasm-pack test --node` — RNG-on-wasm32 smoke test passes
- [ ] **Manual smoke — 4-player commander (needs human):** local forge-server, 4 tabs, commander room, ready up, start. All 4 see a populated board; prompts route per-seat; GameOver shows for all when someone wins.
- [ ] **Manual smoke — 3-player, local=player-1 (middle seat):** exercises the SAB-array skip in `sab_by_player` construction. End-seat hosting can work while middle-seat hosting breaks silently if the iterator advance is wrong — this is the scenario that catches it.
- [ ] **Non-host client on a 4-player game:** this PR removed the `setRemotePlayerSlot` call from `WebGameApi.startMultiplayerGame`'s host branch. Confirm the non-host branch (not in this diff) doesn't reference the removed setter and still routes prompts correctly.
- [ ] **Tauri regression:** 2-player vs AI on desktop — `run_multiplayer_game` body changed (refactor + RNG threading) even though behavior should be identical.
- [ ] **1v1 web commander regression check.**

## Build artifacts


emanueleDiVizio:
stable now, testing manually

emanueleDiVizio:
seems to be working on a 2player vs 2 ai matchup. it's working! jaco will take over and do a bit of further testing

emanueleDiVizio:
Nice work Jaco. i'd approve but cant approve my own PR lol

JacopoMadaluni (dismissed):

JacopoMadaluni (approved):

PR #25, opened 2026-05-20
feat(web): configure hosted forge runtime

emanueleDiVizio:
## Summary
Brings PR #554 (private mirror) onto public `main` and makes web "Play vs AI" optionally run server-side on the **canonical Java Forge engine** via a self-hosted node — fully playable end-to-end, but **off by default**.

**Tauri runtime bundle** — stage the Java Forge harness + res as a generated Tauri bundle with a Forge-compatible `cardsfolder.zip` (`scripts/harness.mjs`, `src-tauri/build.rs`, `lib.rs`, `tauri.conf.json`).

**Web config** — relay endpoint + the hosted-engine availability flag are baked at build time via `Dockerfile.web` args (`VITE_RELAY_*`, `VITE_HOSTED_AI_ENABLED`); Caddy serves the static bundle (prod `ops/Caddyfile`, dev `ops/Caddyfile.local`).

**Self-hosted-node (opt-in)** — a headless node (`forge-engine/crates/self-hosted-node/`, new Dockerfile) hosts a relay room and runs the **Java** engine in-process via j4rs (a Rust facade over `forge.harness.ManaBrewEngineAdapter`), with a Rust `forge_bot` opponent that actually plays. Web "Play vs AI" routes through it only when `MANABREW_HOSTED_AI_ENABLED=true`.

## Why
- packaged Tauri builds should own the harness/res as bundle assets; web deploys need the relay + hosted-AI toggle configurable at container start
- hosting the canonical Java engine gives parity-correct AI games without waiting on the Rust engine port

## Default behavior (important)
- **`MANABREW_HOSTED_AI_ENABLED` defaults to `false`.** 1v1 "Play vs AI" runs the engine **client-side (WASM)** and does **not** depend on the node.
- The node is **opt-in** behind a `hosted-ai` compose profile (`docker compose --profile hosted-ai up`).
- The node's room is `hosted`, so it's **filtered out of the lobby's human room list** — the node can stay live without advertising. The Play-vs-AI flow still discovers it server-side.

## What it took to make the hosted Java game playable
[...]

PR #32, opened 2026-05-21
feat(node): self-play soak harness (java + rust) for the hosted prompt path

### LINKED ISSUE #31: hosted java play stalls on chooseTargetAny (target id round-trip)
found by the new java self-play harness (#3a / `SELF_HOSTED_NODE_JAVA_SELF_PLAY`).

repro: bot-vs-bot java game, turn 5, lightning bolt resolves and asks for a target (`chooseTargetAny`). the normalized prompt looks fine — `validCardIds: [engine-card-63]`, `validPlayerIds: [player-0, player-1]`, `hostile: true`. simpleai picks the first valid card and sends `TargetAny { card_id: engine-card-63 }`. java never resolves it and re-emits the same prompt → stall.

looks like a target-id round-trip gap: the normalizer rewrites java's card id to `engine-card-N` for the client outbound, but `translate_java_action_value` doesn't map it back to java's original id when building the action, so java sees an unknown target and ignores it.

where to look:
- `forge-agent-interface/src/java_prompt_normalizer.rs` — `translate_java_action_value`, and how target card ids get normalized outbound
- the `chooseTargetAny` / `TargetAny` path specifically (player targets might be fine, card targets aren't)

the harness dumps the full artifact (raw prompt + normalized + snapshot) at `target/self-play-stuck-<ts>.json` so it's easy to repro.

emanueleDiVizio:
## Summary

self-play soak harness for the hosted prompt path — drives a forge game with simpleai on every seat and checks it carries a full game end to end, plus a client-half check that the UI can ingest every prompt the engine produces.

**server half — node self-play (two backends):**
- `SELF_HOSTED_NODE_JAVA_SELF_PLAY` (`--features java-forge`): every prompt goes `normalize_java_prompt` → `AgentPrompt` → bot → `translate_java_action_value` → java. polls to game over; on a normalize failure / rejected action / repeated-prompt stall / prompt-cap it dumps `target/self-play-stuck-<ts>.json`.
- `SELF_HOSTED_NODE_RUST_SELF_PLAY` (default build, no jvm): runs the rust engine via `run_hosted_multiplayer_game` with a bot per seat. no normalizer — rust emits `AgentPrompt` natively — so it covers the DTO + bot + engine path wasm/tauri use. ~1s, CI-able.
- configurable: `SELF_HOSTED_NODE_SELF_PLAY_DECKS` (preset/parity deck ids), `_PLAYERS`, `_FORMAT` (`commander` → 40 life), `_SEED`. caps via `_PROMPTS` / `_TURNS`. `PRESET_DECKS_DIR` points at any deck dir (e.g. `parity_decks`).
- **seeded**: rust is fully deterministic given the seed (same seed → byte-identical game); java is seeded (drives the game) but not bit-deterministic — forge has residual non-determinism (hashmap order / hosted-session timing) that parity tames with deterministic collections, which the hosted path doesn't replicate. still makes failures far more reproducible.

**client half — `yarn test` (vitest):**
- `src/stores/gameStore.constants.test.ts` runs the rust `emit_prompt_fixtures` bin **on demand** (one example of every `AgentPromptInner` variant — 51 types, sourced from the real types so it can't drift) and replays them through the UI's real `applyPrompt`, asserting every type is known + handled + routes + ingests without throwing. no committed fixture, no browser.

## Why

the `relay → normalize_java_prompt → AgentPrompt → bot/ui → translate_java_action_value` round-trip is where every hand-found hosted bug lives, and the only way to test it was hand-playing a game until it broke. now: server-side self-play runs the game and dumps where it sticks; the client check confirms the UI consumes every prompt the engine emits.

### "doesn't parity already cover this?" — no, and here's the proof

parity diffs the **engine rules** (rust trace vs java trace), driving both engines through its own deterministic controllers — it never touches `normalize_java_prompt`, the `AgentPrompt` DTO, `translate_java_action_value`, the bot, or the UI's `applyPrompt`. that whole transport/normalization layer is a blind spot for parity by construction. proof: this PR fixes #31, a target-id round-trip stall in `java_prompt_normalizer.rs` — **parity never flagged it because parity never executes that code**; the harness caught it on its first full game.

parity = "is the rust port faithful to java"; self-play = "does hosted play actually carry a game". both can be green while the other is red. complementary, not a replacement — it leans on parity for rules-correctness and only asserts the hosted path carries a game + the UI ingests every prompt.

note: java emits a *subset* (the ~27 normalizer types); the rust engine / emitter cover the full 51-variant `AgentPrompt` contract the wasm/tauri client must handle — so the client check is a superset of java's emissions, which is the point.

## what it found

ran the full parity-deck matrix (7 decks, 21 pairs, both backends): **rust 19/21, java 16/21**. surfaced 4 bug classes, filed separately — #36 (engine selector divergence), #37 (harness snapshot CME), #38 (NoSuchElementException), #39 (choose_attackers stall). #31 fixed here as the demonstration; the rest are follow-ups so this PR stays the tool.

## Test plan

- `cargo check -p self-hosted-node` + `--features java-forge` clean, clippy clean, fmt clean
- rust self-play: 2p/4p run to a winner; deterministic across same-seed runs
- java self-play: full games after the #31 + game-over-detection fixes (4/4)
- `yarn test`: vitest runs the emitter on demand, replays all 51 prompt types through `applyPrompt` — green
- default build excludes the java mode (all `#[cfg(feature = "java-forge")]`)

## Build artifacts

- [ ] Build macOS .dmg
- [ ] Build Windows .exe

---
rebased onto main after #25 (cb94dd6b5) landed — the diff is just the self-play harness.


emanueleDiVizio:
fixed #31 here (bcd50ecfb) — the loop earning its keep.

root cause: `normalize_card_id` rewrites java's `java-card-<n>` → `engine-card-<n>` on the way out to the client, but actions came back unflipped. java matches a submitted target id against its own `javaCardId` (`java-card-<n>`), so `engine-card-<n>` never matched — java just silently dropped the action and re-issued the same prompt → stall. players (`player-<n>`) and stack ids have no `engine-card-` prefix so they round-tripped fine, which is exactly why `chooseTargetPlayer` worked and `chooseTargetAny` on a card didn't. fix is the inverse (`denormalize_card_id`) over every id in an outbound action — covers targeting, combat, discards, the lot, not just the one prompt.

loop result: before, every java game stalled at turn 5 on chooseTargetAny. after, it plays through targeting + combat and the next wall is **turn 14, a `priority` prompt in CombatDamage** — a separate bug. so the harness is doing its job: peel one, fix one, see the next. i'll file the turn-14 one separately rather than scope-creep this pr.

rust path remains 5/5 to a winner; this only touched the java action translation.

emanueleDiVizio:
ran the harness across the full parity-deck matrix — 7 decks (red_burn, green_stompy, white_aggro, black_control, comprehensive_test, trigger_expanded, staticability_test), all 21 pairs, both backends.

- **rust: 19/21 complete**
- **java: 16/21 complete**

it earned its keep — surfaced 4 distinct real bug classes, filed separately:
- #36 — engine: compiled card selector diverges from string matcher (`valid_filter.rs:509`) — core engine / parity, rust panic
- #37 — harness: ConcurrentModificationException in `InteractiveSnapshotExtractor` snapshot
- #38 — java self-play: NoSuchElementException mid-game
- #39 — hosted: stall on `choose_attackers` at declare-attackers (always vs staticability_test)

and #31 (chooseTargetAny target-id round-trip) is fixed *in this pr* as the demonstration. the four above are follow-ups — leaving them out so this pr stays the tool, not a grab-bag of engine fixes.

so this pr now carries: java + rust self-play (configurable decks/players/format), the client-half `yarn check:ui-prompts`, prompt recording, and the #31 fix. ready for review/merge (stacked on #25 — retarget to main once that lands).

emanueleDiVizio:
tidied up after jaco's notes + a couple of follow-on questions:

- **seeded** both backends (`SELF_HOSTED_NODE_SELF_PLAY_SEED`). rust is fully deterministic given the seed (same seed → byte-identical game); java is seeded — drives the game — but not bit-deterministic, forge has residual non-determinism (hashmap order / hosted-session thread timing) that parity only tames with its deterministic collections + controller, which the hosted path doesn't replicate. so a found stall is reproducible on rust, and close-enough on java.

- **client check is now a proper vitest test** (`yarn test`), not a loose script in `scripts/`. it can't be rust (it runs the TS `applyPrompt`), but the fixture now comes from a **rust emitter** (`emit_prompt_fixtures` bin) that serializes one of *every* `AgentPromptInner` variant — 51 types, vs the ~6 a recorded game happened to hit. a new variant won't compile until it's added, so coverage tracks the enum.

- **fixture is generated on demand** by the test (runs the emitter via child_process), so there's no committed artifact to drift. dropped the `SELF_HOSTED_NODE_SELF_PLAY_RECORD` game-dump path entirely — the emitter supersedes it, and java self-play already deserializes the normalizer output as a valid `AgentPrompt` so shape bugs surface there.

added vitest (overdue). all 51 prompt types pass through `applyPrompt`, both feature states + clippy + tsc clean.

forge-engine/crates/forge-agent-interface/src/java_prompt_normalizer.rs
```diff
@@ -697,7 +697,11 @@ pub fn translate_java_action_value(action_value: &Value) -> Value {
         return action_value.clone();
     }
     serde_json::from_value::<PlayerAction>(action_value.clone())
-        .map(|action| translate_java_player_action(&action))
```
emanueleDiVizio, line 700: Actual fix is to change 

/forge-harness/src/main/java/forge/harness/SnapshotExtractor.java

 return "java-card-" + ParityCardMap.parityId(c);
  
to return engine-card

forge-engine/crates/self-hosted-node/src/engine_backend/rust_backend.rs
```diff
@@ -92,6 +94,62 @@ pub fn run_hosted_engine_game(
...
+        let mut prepared =
+            prepare_registered_player(format!("Self-Play {}", i + 1), get_card_db(), &identities);
+        prepared.registered.starting_life = starting_life;
+        if let Some(ref commander) = seat.commander_name {
+            if !force_commander_by_name(&mut prepared, commander) {
+                warn!(commander, "commander name not found in self-play deck");
+            }
+        }
+        prepared_players.push(prepared);
+    }
```
emanueleDiVizio: Follow up: this is duplicated within rust_backend.rs and proabbyl elsewehre in teh code (tauri/wasm). coudl be extracted

forge-engine/crates/forge-agent-interface/src/java_prompt_normalizer.rs
emanueleDiVizio: done in cd7708560 — javaCardId now emits `engine-card-` so there's no rename either direction. dropped normalize_card_id + the denormalize i'd added (and a dup optional_string). it's hosted-only (parity's trace uses ParityCardMap.parityId directly) so parity stays green — verified HEALTHY, and java self-play still plays through targeting to game over.

forge-engine/crates/self-hosted-node/src/engine_backend/rust_backend.rs
emanueleDiVizio: agreed — it's the deck->identities->RegisteredPlayer prep, also in run_hosted_engine_game here and probably the tauri/wasm hosts. filed #47 to pull it into a shared forge-game-runtime helper so this stays scoped.

emanueleDiVizio: actually did it inline in 6b810f98b — extracted `forge_game_runtime::deck::prepare_players` and routed all four hosts through it (node run_self_play + run_hosted_engine_game, wasm single-player + multiplayer). closing #47.

emanueleDiVizio:
java path is now crash-clean. drove the self-play harness across the parity decks after a couple of fixes:

- **#37 / #38 (CME + NoSuchElementException)** — same root cause: the game-over poll and java `submitAction` both built a full snapshot on the j4rs caller thread while the forge game thread was mutating state. fixed with a cheap `getGameOver` query (`game.isGameOver()`, no iteration) for the poll, and `submitAction` no longer returns a snapshot (it raced the thread it just unblocked — caller polls instead). only snapshots left are built on the game thread (prompts) or the error-path dump.

results:
- core matrix (7 decks, 21 pairs): **19/21 OK**, zero crashes — the 2 are `did not reach game over within 1500`, i.e. slow bot-vs-bot games that hit the prompt cap (one ended with a player at 3 life). that's bot aggression / soak budget, reported cleanly, not a hosted-path bug.
- broader soak (planeswalkers, poison, tokens, replacements, keywords, phyrexian, payment, triggers vs red_burn, 2 seeds each): **22/22 OK**.

so for the product — a human vs the AI — the java path no longer crashes, targeting round-trips (#31), and games the human drives conclude. #37 + #38 are fixed by the commits here.

JacopoMadaluni (approved):

PR #33, opened 2026-05-21
fix(relay): app-level keepalive so idle clients aren't dropped

emanueleDiVizio:
## Problem
Players report being unable to stay connected — kicked from the lobby and dropped mid-game. The relay logs show:

```
[recv] idle timeout from 'Jaco21' (no frames for 90s)
[cleanup] 'Jaco21' removed (was not in a room)
```

`forge-server` drops any connection with no inbound frame for 90s (`READ_IDLE_TIMEOUT`). It sends a WS Ping every 25s and relies on the browser's automatic Pong to reset the read timer — but those WebSocket **control frames don't reliably survive the reverse-proxy hop**, so a client that isn't actively sending app frames (sitting in the lobby, or waiting through a long opponent turn) hits the 90s timeout and gets disconnected. Active in-game clients survive because their real messages reset the timer.

## Fix
Application-level heartbeat on the Text-frame path the rest of the protocol already uses (auth/game messages flow through the proxy fine):
- `protocol`: add `ClientMessage::Ping`
- `forge-server`: handle it as a no-op — its arrival alone resets the read timer
- web client: send `{type:"Ping"}` every 30s while the socket is open; cleared on disconnect

The existing WS-level ping/pong stays; we just stop depending on control-frame plumbing for liveness. 30s interval gives 3 beats inside the 90s window.

## Test plan
- [ ] Deploy → sit in the lobby > 90s without interacting → stay connected
- [ ] In a 2-player game, let one player idle through the other's long turn → no drop
- [ ] forge-server logs no longer show `idle timeout` for connected clients

## Build artifacts
- [ ] Build macOS .dmg
- [ ] Build Windows .exe

### DIFF
--- a/src/platform/web.ts
+++ b/src/platform/web.ts
@@ -547,6 +547,7 @@ class WebServerApi implements IServerApi {
   private serverPassword: string | null = null;
   private bots = new Map<string, BotEntry>();
   private wasmReady: Promise<typeof import("@/wasm/forge_wasm")> | null = null;
+  private keepaliveTimer: ReturnType<typeof setInterval> | null = null;
 
   constructor(eventBus: WebEventBus) {
     this.eventBus = eventBus;
@@ -585,6 +586,7 @@ class WebServerApi implements IServerApi {
           username: params.username,
           password: params.password,
         });
+        this.startKeepalive();
         resolve();
       };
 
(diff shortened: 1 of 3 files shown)


src/platform/web.ts
```diff
@@ -585,6 +586,11 @@ class WebServerApi implements IServerApi {
           username: params.username,
           password: params.password,
         });
+        // Application-level keepalive: forge-server drops a connection after
```
fedepoi: remove please

fedepoi (dismissed):

fedepoi (approved):

PR #41, opened 2026-05-21
fix(server): clean up relay sessions when websocket writes fail

anacletos:
## Build artifacts


emanueleDiVizio (approved):
Nice work

PR #53, opened 2026-05-22
fix(multiplayer): broadcast game-over and unstick lobby return on concede

### LINKED ISSUE #45: bug: Multiplayer concede does not reliably return the client to lobby

fedepoi:
## Summary
- Wire `PlayerAction::Concede` through the engine: new `EnginePlayerAction::Concede` variant; priority loop calls `crate::player::concede(game, priority_player)`, advances priority, lets SBA pick up `alive_players() ≤ 1` on the next iteration. Game-end then flows through the engine's existing prompt-broadcast path — no special host-side broadcast needed.
- `GameViewDto.conceded_player_ids` (mirrored as `concededPlayerIds` in TS); `GameOverScreen` renders "You conceded" / "{name} conceded" instead of "Draw!" when game-end is concede-driven.
- Client cleanup: `useServerStore.leaveRoom` clears `gameStarted`/`playerOrder`/`playerDecks`/`startingLife` so users don't strand in lobby with stale flags; `useGameStore.endGame` resets state synchronously then races `runtime.api.endGame()` + `leaveRoom()` against a 2 s timeout so a dead relay socket can't hang the UI.
- Delete the host-side concede shortcut in `src-tauri/src/game_manager.rs` (`perform_concede`, `broadcast_game_over_to_remotes`, `local_engine_player_index`, the Concede intercepts in `respond()` and `route_remote_response()`).
- `src/game/runtimeRegistry.ts` no longer forces `concedeBehavior: "end-session"` on the web platform — the WASM transport already deserializes `{"type":"concede"}` into `PlayerAction::Concede`, so web now flows concede through the engine like Tauri. Manual-tabletop keeps `end-session` (no engine).
- Extract `DEFAULT_STARTING_LIFE = 20` in `useServerStore.ts`; replace 6 literal sites (5 in `useServerStore`, 1 fallback in `useGameStore`).

## Why
Fixes #45.

The reported symptoms had three collaborating root causes:

1. **Engine never handled concede.** The agent-interface had `PlayerAction::Concede`, but the engine's internal `PlayerAction` had no `Concede` variant. `PromptAgent::choose_action` therefore mapped any inbound concede to the `_ => PassPriority` catch-all, so conceding through the engine did nothing. The Tauri host worked around this with a ~100-line shortcut that synthesized a `GameOver` prompt and aborted the session — single-player AI semantics for `winner_id`, no relay broadcast, and the original abort path skipped `GameNotification::GameOver`.
2. **Remote concedes were silently dropped.** A Concede arriving at the host via `route_remote_response` was forwarded to the engine's per-player response channel where the same `_ => PassPriority` mapping applied. Conceding as a non-host had no effect at all.
3. **Client cleanup was incomplete.** `leaveRoom()` cleared `currentRoom` but left `gameStarted=true`, so `Lobby.tsx`'s effect short-circuited and stranded the user. `endGame()` awaited network teardown *before* resetting state, so a dead relay socket could freeze `GameOverScreen` ("Returning to menu…") indefinitely.

After the fix, concede flows through the same code path as every other game-end. `host_runtime::run_hosted_multiplayer_game` emits `GameNotification::GameOver` to all agents via the existing prompt channel; `spawn_remote_prompt_forwarder` broadcasts it to relay peers via the same `StateEnvelope::Prompt` envelope that every other prompt uses. The engine's SBA picks the correct `winner_id` for any room shape — including 3+ player games, which now keep going until `alive_players ≤ 1` instead of ending on any single concede.

## Known limitation (flagged for follow-up)

**Concede only takes effect at the conceder's next priority window.** The engine's priority loop intercepts `PlayerAction::Concede` returned from `choose_action`, so the action has to make it through the conceder's `recv_action` for the engine to see it. If the user clicks Concede while the engine is waiting on them for a *non-priority* prompt (`choose_blockers`, `choose_targets`, a cost prompt, etc.), that prompt's `recv_action` consumes the action and falls into its own catch-all — the engine moves on as if no choice was made, and the concede only takes effect at the next priority window for that player.

In practice the engine re-enters priority after almost every meaningful event (spell resolves, phase transitions, etc.) so the delay is usually sub-second. But under heavy mid-combat or mid-cost prompts the user can see a brief unresponsiveness. The proper fix is a side-channel "concede pending" flag checked in SBA so concede is always atomic — separate refactor, tracking issue to follow.

Still out of scope here: translating a relay `PlayerDisconnected` event into a Concede on the host's engine. When a web tab closes mid-game without an explicit concede, the host's engine only sees the per-remote channel's `recv_timeout` (120 s) and returns Pass, not Concede — SBA never sees a loser and the game continues with that player idle. The proper fix is to surface relay disconnect events at `route_remote_response` and inject `PlayerAction::Concede` into the dropped player's channel. Separate PR.

## Test plan
Automated:
- `cargo check -p manabrew` — clean.
- `cargo check -p forge-engine-core` and `cargo check -p forge-agent-interface` — clean.
- `cargo fmt --all -- --check` — clean on changed files.
- `eslint` + `prettier --check` on edited TS files — clean.
- Pre-existing `yarn lint:all` failures (14 prettier files in `components/game/{dice,modals}` + `components/limited`, 51 clippy errors in `forge-engine-core`, a tsc error on `gameStore.constants.test.ts`) confirmed present on `main` via `git stash` round-trip — untouched by this PR.

Manual (recommended before merge):
- [ ] Single-player concede vs AI → `GameOverScreen` shows "You conceded", auto-returns ≤3 s; `gameStarted`/`currentRoom` clean in Zustand devtools.
- [ ] Tauri 2-player relay multiplayer: host concedes → host sees "You conceded", remote sees "You Win!" with correct `winnerId`, both return to lobby ≤3 s.
- [ ] Tauri 2-player relay multiplayer: non-host concedes → same expectation in mirror.
- [ ] **Web 2-player relay multiplayer: host concedes (new path)** — host sees "You conceded", remote sees "You Win!". Validates the WASM-transport concede serialization end-to-end.
- [ ] **Web 2-player relay multiplayer: non-host concedes (new path)** — same expectation in mirror.
- [ ] **Tauri 3-player relay multiplayer: one player concedes → game CONTINUES for remaining two players, conceder leaves to lobby.** This is the SBA path; if you see the game end for everyone, that's a regression.
- [ ] Mid-game relay kill (`pkill -f forge-server`) → both clients exit `Game.tsx` ≤5 s (`server:disconnected` handler clears `gameStarted`) and land on lobby.
- [ ] `GameOverScreen` "Return to Menu" button is a hard exit even with the relay socket closed (state reset is synchronous in `endGame`).
- [ ] Parity smoke (`yarn parity`) — verify the new `EnginePlayerAction::Concede` variant doesn't cascade into regressions across normal gameplay (concede isn't exercised in card-script parity, but the priority-loop edit lives on the hot path).

## Build artifacts


emanueleDiVizio (dismissed):

emanueleDiVizio (dismissed):

emanueleDiVizio (dismissed):
top notch

emanueleDiVizio (dismissed):
lgtm

emanueleDiVizio (approved):
Looks good!! I'd say let's merge and deploy so we cna playtest

PR #54, opened 2026-05-22
feat(node): hosted java-forge play-vs-ai — espresso concurrency, lobby, engine select

emanueleDiVizio:
## Summary

the full hosted **Java Forge "Play vs AI"** feature for the web client. started as the espresso concurrency spike; the stacked lobby redesign (#78) is now merged in here, so this is the complete thing.

**concurrency (the original blocker).** forge keeps process-global statics (`MyRandom`, the `maxId` counters), so a second `startGame` corrupts the first live game. each game now runs in its own graalvm **espresso** context — isolated statics, one shared jvm, no AOT and no forge changes. `java-espresso-host` (`ManaBrewEspressoAdapter`) is the host router (same json api), with a pre-warmed pool + background replenish; single-use by default, in-place reuse opt-in. a `JavaEngine` actor owns the jvm and services session-tagged commands so games run concurrently. ~260 mb/context.

**room lifecycle.** node hosts games through the shared actor (not a per-game jvm), advertises `capacity` on the heartbeat, and recycles rooms: game-over / player-leave / disconnect-grace tear the game down and reset the room (`EndGame` → `end_game_sync`); cancelled games abort-close their context (never reuse a context whose guest thread may still be blocked).

**hosted lobby redesign (from #78).**
- **any-format single pool** — rooms are created format-agnostic (`GameFormat::Any`); the format is chosen at game start (rides on the `startGame` payload) and resets to Any on reuse. one pool for play-vs-ai + human tables.
- **engine selection** — `RoomInfo`/`CreateRoom` carry an `EngineKind` (Wasm|Java); lobby shows every Lobby room with an engine badge (Forge/Rust) + a ManaBrew badge for hosted rooms, and hides in-progress rooms. play-vs-ai opens a **modal** to pick Forge (hosted) vs Rust (wasm) — replaces the old experimental settings toggle.

**prompt rendering.** prompt cards resolve their art via the DTO (setCode/cardNumber/owner) like the rust path — not a name hack; reveal modal capped + scrollable; battlefield target candidates clickable.

## Why

infra won't allow unbounded node processes, so in-process isolation (espresso) is the path to concurrent hosted play. rooms shouldn't be pinned to a format at creation, and hosted play should share one pool with the engine surfaced, not a hidden per-format set.

## Test plan

verified end-to-end on the local docker stack (relay + espresso node):
- `hosted-smoke` harness: discovers the Any pool, joins, locks to a format at start, plays to game-over — `1/1 games completed`. ✓
- concurrency: 2 games concurrently through the `JavaEngine` actor ✓; in-place reuse (one context, two games, no corruption) ✓.
- **browser, end-to-end** ✓: lobby shows the Any pool with Forge/ManaBrew badges; play-vs-ai opens the engine modal; Forge routes to the hosted node; format locks at start and resets to Any after; card art + targeting render.
- `cargo check` / `tsc` clean.

macos note: espresso embedding is broken on macos (libjava/libjvm split-cache) — all of this runs in a linux container; tested via docker, not a native macos build.

## Known limitations — Java engine polish (follow-ups, not blockers)

the hosted Java path is functionally complete but the java→web prompt/action bridge needs a polish pass, all filed:
- **#70** graveyard/zone targeting (the zone-target half; battlefield targeting works)
- **#74** play from alternate zones (graveyard land click doesn't submit)
- **#76** dynamic P/T (Knight of the Reliquary) — web render of the gameView power
- **#77** AI not blocking lethal → combat stall
- **#73** abandoned-context reuse (interrupt-and-reuse instead of force-close)

these are the "polish on dto / prompt / java normalizer" tail; the core feature (concurrent hosted play, lobby, engine selection) is done and verified.

## Build artifacts

no installers — server-side + web only, dmg/exe unchecked.


emanueleDiVizio:
follow-up / long-term shape (not this PR): **true on-demand room provisioning**.

right now the relay ties **host = room creator**, so the model is: the node pre-hosts rooms (one per play-vs-ai format) and the web *joins* one by format. that's what this work does (+ a room→lobby reset so they recycle).

the proper long-term shape, once we're balancing across **multiple nodes**, is to decouple the engine-host role from the creator: the web *creates* the play-vs-ai room in its chosen format, and the relay *assigns an available node* to host it on demand (real matchmaking). that's a relay redesign (host model + node-assignment + capacity-aware routing) — bigger, and only worth it at multi-node scale. the node already advertises `capacity` on its heartbeat, which is the hook for it.

parking it here so we pick it up when node count > 1.

emanueleDiVizio:
capacity / scale notes (staging box = **3 GB RAM, 2 vCPU**):

- each **concurrent** game = its own espresso context ≈ **260 MB** (the card DB; can't be shared without losing per-game isolation). this is the hard wall.
- a room hosts games **sequentially** (the EndGame reset recycles it), not concurrently — concurrent games = separate rooms.
- so this box realistically does **~3 concurrent games**, not more. pre-warm is set to **3** to match. it was already near-full (`available: 0`), so 3 pre-warmed contexts is about its ceiling — watch for OOM; the box wants more RAM even for this.
- **CPU is not the limit** — 2 cores handle many human-paced (think-time-bound) games; RAM is.
- **for ~50 concurrent players**: needs either a **~16 GB box** (→ ~50 contexts) or **multiple nodes** with the relay routing by advertised `capacity` (the on-demand provisioning follow-up above). don't pre-saturate 50 on a 3 GB box — it OOMs instantly.

tl;dr: fine for a low-concurrency friend alpha (the reset handles turnover); needs a bigger box or multi-node before any real concurrent load.

emanueleDiVizio:
pulled the `deploy-preview` label — can't validate the hosted java engine on the staging box.

what happens: the 3 gb box has to **build the image in place** (rust release + maven harness + graalvm) *and* run one espresso jvm (~1.25 gb + ~260 mb card-db) for the staging node, on top of prod's web+relay. the build + the jvm together exhaust ram → swap thrash → ssh unresponsive → deploy wedges. pool size doesn't fix it; it's the build+jvm combo on too small a box.

plan:
- preview deploy drops the self-hosted-node (keep it for cheap web/relay ui previews)
- java engine e2e → fat linux ci runner (hosted-smoke harness) + local docker for manual playtest
- **need a bigger staging/preview box** before this can be previewed at all — capacity wall arriving early

branch is good to review otherwise: per-format rooms + reset, espresso context pool, drop-guard against context leaks on panic/cancel, pool-exhaustion warn-log, InGame guard on end_game.

emanueleDiVizio:
note on espresso context reuse (added in this pr): **finished** games reuse their warm context (instant restart, no ~50s rebuild). **abandoned** games (leave/disconnect) force-close their context — safe, since the guest game thread may still be blocked and would poison a reused one — so rapid start-and-leave churns the pool back to a cold ~50s build. fine for real play (people finish games); only bites repeated abandons. making abandons cheap too (interrupt-and-reuse the blocked thread) is tracked in #73, out of scope here.

emanueleDiVizio:
merged the stacked lobby pr (#78) in here, so the hosted **java forge play-vs-ai** is actually complete now — espresso concurrency, room lifecycle (reuse + abandon teardown), any-format single pool, engine-select modal, and prompt art via the dto.

remaining is **polish on the dto / prompt / java-normalizer bridge**, all filed and functional-but-rough:
- #70 graveyard/zone targeting (battlefield targeting works)
- #74 play from alternate zones
- #76 dynamic p/t render (knight of the reliquary)
- #77 ai not blocking lethal
- #73 abandoned-context interrupt-and-reuse

core feature is done + verified end-to-end (browser + smoke). also re-signed the 7 commits the rebase-merge dropped signatures on.

forge-engine/crates/forge-agent-interface/src/protocol.rs
```diff
@@ -114,7 +116,16 @@ pub enum ClientMessage {
         commander_name: Option<String>,
     },
 
-    StartGame,
+    StartGame {
+        /// The format to lock the game into. Required when the room is `Any`;
+        /// ignored if the room already has a concrete format.
```
anacletos: The kind of comment I'd avoid

forge-engine/crates/forge-agent-interface/src/protocol.rs
```diff
@@ -114,7 +116,16 @@ pub enum ClientMessage {
...
+        #[serde(default)]
+        format: Option<GameFormat>,
+    },
+
+    /// Host signals a finished game so the relay returns the room to the lobby
+    /// (drops players, status back to Lobby) and it can be reused.
```
anacletos: comment

forge-engine/crates/forge-agent-interface/src/protocol.rs
```diff
@@ -239,6 +252,9 @@ pub enum RoomStatus {
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub enum GameFormat {
+    /// A room that hasn't committed to a format yet — the concrete format is chosen
+    /// at game start (see `StartGame`). Rooms are created as `Any`.
```
anacletos: comment

forge-engine/crates/forge-agent-interface/src/protocol.rs
```diff
@@ -251,3 +267,13 @@ pub enum GameFormat {
     Draft,
     Sealed,
 }
+
+/// Which engine runs a room's game. Drives the lobby's engine badge.
```
anacletos: comment

forge-engine/crates/forge-agent-interface/src/protocol.rs
```diff
@@ -251,3 +267,13 @@ pub enum GameFormat {
...
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub enum EngineKind {
+    /// In-browser Rust engine (wasm). The default for human-hosted rooms.
+    #[default]
+    Wasm,
+    /// Java Forge running on a hosted node.
```
anacletos: comments

forge-engine/crates/forge-agent-interface/src/protocol.rs
```diff
@@ -251,3 +267,13 @@ pub enum GameFormat {
     Draft,
     Sealed,
 }
+
+/// Which engine runs a room's game. Drives the lobby's engine badge.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub enum EngineKind {
+    /// In-browser Rust engine (wasm). The default for human-hosted rooms.
+    #[default]
+    Wasm,
```
anacletos, line 268: Is this supposed to be `Rust`?

forge-engine/crates/forge-server/src/lobby.rs
```diff
@@ -262,10 +265,20 @@ pub fn start_game_sync(
...
         }
 
+        // Lock in the format at start: a room created with a concrete format keeps it;
+        // an `Any` room takes the format supplied with StartGame (required).
+        if room.format == GameFormat::Any {
+            match format {
+                Some(chosen) if chosen != GameFormat::Any => room.format = chosen,
+                _ => return Err(ServerError::FormatNotChosen),
+            }
+        }
```
anacletos, line 275: Does the client set the format, or does it provide at start?
The host should be setting the format as they enter, otherwise other clients can't properly filter valid decks on selection

forge-engine/crates/forge-server/src/lobby.rs
```diff
@@ -281,3 +294,60 @@ pub fn start_game_sync(
 
     Ok((room_id, player_order, player_decks, starting_life))
 }
+
+// Return a finished hosted game to the lobby so the room is reusable: drop the
+// players (the human + the AI bot) and reset status; the non-playing host (node)
+// stays as an observer, so findHostedRoom sees an open Lobby room again.
```
anacletos, line 300: comment

forge-engine/crates/self-hosted-node/src/bin/hosted_smoke.rs
```diff
@@ -0,0 +1,368 @@
+//! Headless integration smoke for hosted Play-vs-AI through the relay.
```
anacletos: This is a smoke test right? Should we add it in CI?

forge-engine/crates/self-hosted-node/src/bin/hosted_smoke.rs
```diff
@@ -0,0 +1,368 @@
...
+//!
+//! Drives N concurrent games end-to-end against a running relay + self-hosted-node:
```
anacletos: im not sure I understand this file. It looks like a lot of duplicated logic, which makes the test itself not very useful?

forge-engine/crates/self-hosted-node/src/config.rs
```diff
@@ -14,7 +14,10 @@ pub struct Config {
     pub room_id: Option<String>,
     pub room_name: String,
     pub max_players: u8,
+    pub max_games: usize,
     pub format: GameFormat,
+    // One hosted room is created per format here, so every play-vs-AI format the
+    // web offers has a room to join. Defaults to just `format`.
```
anacletos: Is this comment even relevant anymore?

forge-harness/src/main/java/forge/harness/ManaBrewInteractiveSession.java
```diff
@@ -75,6 +75,17 @@ public void close() {
         if (game != null && !game.isGameOver()) {
             game.setGameOver(forge.game.GameEndReason.Draw);
         }
+        // Wait for the game thread to fully unwind before returning: callers that
+        // pool/reuse the JVM (or Espresso context) must not start a new game while
+        // this one's thread is still finishing match.startGame.
```
anacletos: Comment

forge-harness/src/main/java/forge/harness/ManaBrewInteractiveSession.java
```diff
@@ -788,6 +799,20 @@ private void publishPriorityPrompt(final int playerId, final List<SpellAbility>
         latestPromptJson = prompt.toString();
     }
 
+    // Carry the printing + owner on a prompt-card option so the client resolves the
+    // card's art from its deck the same way the rust engine's gameView DTO does,
+    // instead of a by-name fallback.
```
anacletos: Comment

forge-harness/src/main/java/forge/harness/ManaBrewInteractiveSession.java
```diff
@@ -949,8 +971,15 @@ private void publishRevealCardViewsPrompt(
             final CardView card = cards.get(i);
             JsonObject option = new JsonObject();
             option.addProperty("index", i);
-            option.addProperty("id", "java-card-view-" + card.getId());
-            option.addProperty("label", card.getName());
+            // Resolve the view back to its real card so the option carries printing +
+            // owner (art resolution); fall back to the bare view if it's gone.
```
anacletos: comment

fedepoi (dismissed):

src/views/Play.tsx
```diff
@@ -83,10 +95,32 @@ export default function Play() {
...
         <DeckVsSelector
           onStart={(playerDeck, opponentDeck, formatId, commanderName) => {
-            startGame(playerDeck, formatId, commanderName, opponentDeck);
+            // On web with a hosted node available, let the player pick the engine;
+            // otherwise (Tauri, or no hosted node) just run the local Rust engine.
+            if (getPlatform().type === "web" && isHostedEngineAvailable()) {
+              setPendingAiStart({ playerDeck, opponentDeck, formatId, commanderName });
+            } else {
+              startGame(playerDeck, formatId, commanderName, opponentDeck, "Wasm");
+            }
```
anacletos: For clarity, I actually think it should always show, and if hosted is not availble, it will say so and won't let you select the option. 
Otherwise, a user will go to play expecting to play with java, and he'll be sent to a game without warning, actually using the wasm.

forge-engine/crates/forge-agent-interface/src/protocol.rs
emanueleDiVizio: Yeah good point. we can add another one as Rust (as in, Rust hosted from Tauri). i have no strong opinion here, but i think Wasm is explanatory enough and not really worth another slot and threading it throuh

forge-engine/crates/forge-server/src/lobby.rs
emanueleDiVizio: The client does see the format yeah as part of the playload (like before)

forge-engine/crates/self-hosted-node/src/bin/hosted_smoke.rs
emanueleDiVizio: Yeah, can be an integration test

forge-engine/crates/self-hosted-node/src/bin/hosted_smoke.rs
emanueleDiVizio: It is meant to be a headless smoke test to run , locally or in ci. there is a bit duplicated logic but there's gotta be for this kind of test

forge-engine/crates/self-hosted-node/src/config.rs
emanueleDiVizio: no removing

src/views/Play.tsx
emanueleDiVizio: Yup fair point, although right now it resolves to a build env var -> 

https://github.com/witchesofthehill/manabrew/blob/5aeeed7518ea00ef44d98d70eb0f94decf08da19/src/config/webRuntimeConfig.ts#L7C1-L16C2
```
/** Whether this deployment offers the hosted (server-side Java) engine — i.e.
 *  a self-hosted-node is available. Gates the Settings engine toggle; actual
 *  routing also requires the per-user `preferHostedEngine` preference. Baked at
 *  build time via VITE_HOSTED_AI_ENABLED. */
export function isHostedEngineAvailable(): boolean {
  return ["1", "true", "yes", "on"].includes(
    (import.meta.env.VITE_HOSTED_AI_ENABLED ?? "").toLowerCase(),
  );
}
```

I'll tweak this

emanueleDiVizio:
pushed the review round (rebased onto main, all commits signed):

- stripped the over-added comments — went aggressive per the feedback, only kept three that stop a real silent break: the 64mb engine stack size, the smoke deck shape (40+20 so games end in combat not a deck-out recursion stall), and the idle-ping pong (unanswered relay pings drop the connection). everything else is gone.
- play.tsx now always shows the engine modal on web instead of silently falling back to wasm — when the build has no hosted node the forge option is disabled with a "not available in this build" note.
- moved the hosted smoke out of `src/bin` into `tests/hosted_smoke.rs` as an `#[ignore]`d integration test (asserts every game reaches game-over), plus a label/nightly gh actions job that boots the compose stack and runs it. also dropped the dead `SELF_HOSTED_NODE_FORMATS` env.
- format-on-enter needs a new `SetFormat` message + handler + lobby ui and play-vs-ai already works with format-at-start, so i split it out as #81 to keep this focused.

PR #80, opened 2026-05-27
feat(limited): multiplayer booster draft (+ image hydration / DraftCardDto cleanup)

fedepoi:
## Summary

This branch grew from a small gauntlet-WASM-commands fix into a full multiplayer draft feature on top of the limited / image-hydration / DraftCardDto cleanup.

### Multiplayer draft (commits 31a9a7a..160ae7df)

End-to-end pod draft over the existing room-relay channel. Host runs the engine, peers receive per-seat state broadcasts and submit picks back. Tournament play is explicitly out of scope (deferred to a future PR).

- **Engine** — `BoosterDraft::with_human_seats(humans: &[(usize, String)])` supports an arbitrary set of human seats; `submit_human_pick_for(seat_idx, card)` queues picks per seat; `tick()` blocks until every human seat has decided. Single-player flow unchanged.
- **Wire** — three new wasm/Tauri commands: `limited_start_multiplayer_draft`, `limited_submit_pick`, `limited_get_seat_state`. `DraftStateDto::from_engine_for_seat` renders any seat's POV.
- **Relay** — `draft-v1` protocol over `RoomRelayEnvelope`. Server stays a dumb relay. `DraftStart` / `DraftStateBroadcast` / `DraftPick` / `DraftComplete` payloads in `src/game/draftRelay.ts`.
- **Host** — `src/game/draftHost.ts` starts the engine session with seat assignments, applies peer picks through a serialised `pendingChain` promise (so concurrent picks don't race on broadcast), broadcasts per-peer state after every pick, detects completion and broadcasts final pools, then tears down. Disconnect handler in `useServerStore` tears the host session down too.
- **Peer** — `src/game/draftPeer.ts` attaches on `AuthResult` (keyed by username — matches server's `from_player` stamping), flips the multiplayer draft store on `start`, applies targeted `stateUpdate`s, and sends `pick` envelopes through `submitPeerPick`.
- **UI** — new `MultiplayerDraft` view; reuses the single-player `DraftingView` for the pack/pile layout during drafting and the shared `LimitedDeckBuilder` for the post-draft build phase. `StartMultiplayerDraftDialog` configures set + rounds + picks-per-pass + bot-fill via the extracted `SetPicker` (shared with the Limited hub). TablesList host CTA swaps to "Start Draft" on draft-format rooms. `useMultiplayerDraftStore` drives both host and peer. Auto-navigate from Lobby when the store flips to drafting. End-of-draft → builder with the local pool prefilled → "Save to My Decks" persists the deck (with leftover picks parked in the maybeboard for a lossless snapshot; synthesised basics from "fix manabase" are excluded).

### Limited cleanup (commits 6c553346..639b14b0)

The original gauntlet/cardset/draft-image work — shipping in the same PR because the multiplayer draft depends on it (image hydration, `CardIdentity` wire shape, etc.):

- Worker dispatched gauntlet WASM commands (Play Match / Sideboard fixes)
- Content-hashed cardset archive (`cardset.<sha>.rkyv` + manifest); fixes the stale Cache API entry that hid the editions data
- Checksum-gated `yarn build:wasm` via `scripts/ensure-wasm.mjs`
- Edition parser accepts `DraftBooster=` as a fallback (unblocks Strixhaven, Innistrad Double Feature)
- DraftCard collapsed to identity ref; image hydration routed through `useScryfallStore`; `DraftCardDto` folded into the canonical `CardIdentity`
- Lobby auto-reconnect tightened so a dead relay can't fire "Insufficient resources"

### LimitedDeckBuilder save (applies to draft + sealed)

`LimitedDeckBuilder.handleSaveToMyDecks` now writes leftover pool cards to `deck.maybeboard` so the saved deck is a lossless snapshot of every pick. Applies to both booster draft and sealed (the component is shared). Synthesised basics (from "fix manabase") are excluded — they weren't drafted in the first place.

## Test plan

- [x] `yarn typecheck`
- [x] `cargo check -p forge-wasm -p forge-agent-interface -p manabrew`
- [x] `cargo test -p forge-limited` — 37 tests pass (BoosterDraft multi-human refactor preserved single-player behavior)
- [x] Prettier clean across every file this PR touches
- [x] First run of `ensure-wasm.mjs` rebuilds + stamps; second run reports `up to date — skipping build:wasm`
- [ ] **Manual** (two-tab dry run pending): host creates Draft-format room, peer joins, host opens Start Draft dialog → set + bots → click Start → both navigate to `/draft/multiplayer`, packs render with images, picks round-trip, draft completes, both clients land on `LimitedDeckBuilder` with their own pool, Save to My Decks persists a deck with main/side and (if leftovers) maybeboard
- [ ] Manual: open a Strixhaven (STX) set on the Limited hub → confirm booster recipe shows the real slots instead of the generic fallback
- [ ] Manual: pick a Sealed pool → open gauntlet → "Play Match" → in-game cards render with real artwork
- [ ] Manual: with the relay server stopped, navigate to Lobby → single failed connect, no console spam

## Out of scope (future PRs)

- Tournament / pairings runner
- Observer mode for non-seated peers (`UI already has an empty-state placeholder for `myPool.length === 0`)
- Host-disconnect recovery (mirrors the existing multiplayer-game model — host dies, session dies)
- Username uniqueness gate (two tabs same username — same model as existing multiplayer game; needs server-side handling)

## Build artifacts


khaliostr (commented):
Rebase looks clean, but I think the model didn't migrate with it — flagging before this lands.

#82 changed three things this PR was written against the pre-version of:
- Rooms are created `GameFormat::Any`; format is chosen at `startGame` and resets to `Any` on reuse
- Every room carries an `EngineKind` (`Wasm` | `Java`), surfaced via `EngineChoiceModal` in the play flow
- Room lifecycle resets to `Any` on game-over / leave / disconnect-grace

This branch still has "Draft-format rooms only" in its TS comments, and `TablesList` gates the "Start Draft" CTA on `currentRoom.format === "Draft"` specifically. Worth migrating before merge:

1. **Drop "Draft-format rooms."** Rooms should be created `GameFormat::Any`; `StartMultiplayerDraftDialog` rides on a `startGame`-style payload that carries the draft format/set at start time. Same room-model pattern #82 introduced for play-vs-ai, applied from the lobby rather than the Play view. Otherwise the lobby contradicts what just landed in main.
2. **Carry `EngineKind` on the draft path.** Even if only `Wasm` is implemented here, declaring it on the room leaves a clean seam for a future `Java` draft host. A disabled "Forge (hosted)" entry in the start-draft dialog would also surface the limitation to users.
3. **Reconcile draft lifecycle with room recycling.** Define what `DraftComplete` and draft-host disconnect-mid-pack do to the room — reset to `Any`, or something else? Right now `draftHost.ts` teardown runs independently of #82's lifecycle.
4. **Hosted-smoke parity.** The TablesList/Lobby changes here overlap with what `hosted-smoke` exercises. Worth confirming the harness still passes, or extending it with a draft variant.

Non-blocking follow-up worth filing as an issue: the self-hosted-node has no handlers for the new `draft-v1` envelopes, so it sits idle in a Draft room while the wasm client hosts. Worth defining the node-present behavior (ignore? race the client for host?) even if node-host draft is deferred — silent host-role conflicts in a relayed protocol are painful to debug after the fact. Note that the Java backend isn't actually needed for drafting itself (pure card-selection, all in `forge-limited`); the missing wiring is node-as-draft-host + hand-off from the completed pool into hosted-AI / hosted-multiplayer match. Coherent follow-up shape.

khaliostr:
Sharpening the framing on asks 1+2 above, since I don't think I called it out clearly: those aren't independent cleanups — together they're the **precondition** for the obvious next feature (multiplayer-draft → multiplayer-Java-match in the same room).

The wasm host can't run Java, so any post-draft Java-backend match has to be node-hosted. For a Java-capable node to be in the room when the match starts, the room has to declare `EngineKind::Java` from creation — which is impossible under "Draft-format rooms" (no engine slot on the room) but natural under post-#82 `Any` + `EngineKind`. So migrating the room model isn't just consistency with #82, it's what unblocks the continuity feature.

Filed as #84 with a fuller design sketch so this PR isn't blocked on it. Happy to keep #80 scoped to the wasm-hosted slice + the room-model migration; the node-side passive-during-draft behavior + match-handover is the follow-up.

Not blocking — just wanted the rationale linked from the PR for whoever picks #84 up.

khaliostr (dismissed):
Migration looks clean — asks 1-3 from the earlier review all addressed:
- Rooms drop `Draft` from creation; `StartGame("Draft")` flips `Any` → `Draft` server-side (`910996de`)
- `EngineKind` flows through `CreateRoomParams` with the disabled "Forge (hosted)" placeholder waiting on `isHostedEngineAvailable()` (`910996de` + `ddbbc08c`)
- `ClientMessage::EndGame` on `DraftComplete` / host-exit → room resets to `Lobby` + `Any`, mirroring play-vs-ai lifecycle (`910996de`)

Two bits beyond the asks worth calling out:
- `11af65b0` baking `DraftConfig` into the room at creation is a nicer UX than the original second-dialog flow — joining players see what they're getting into before they ready up
- `9be366fb` hardening `start_game_sync` to reject `Draft`/`Sealed` starts with no config closes a real peer-stranding vector (malicious or just buggy `StartGame` → InGame transition with no envelopes coming). Good catch.

On ask 4 (hosted-smoke parity): no change here and the harness doesn't appear in this PR's CI rollup — assuming it runs on main post-merge. Worth a sibling smoke for the Draft format-flip → `EndGame` → `Any` reset cycle eventually, but not a blocker for this PR.

Happy to take #84 from here — node-side `draft-v1` no-op + match handover. The `EngineKind` seam through `CreateRoomParams` + the disabled "Forge (hosted)" UI placeholder mean the frontend's already wired for the engine flip; this becomes a node-only follow-up. Will pick it up once this lands.

Approving.

forge-engine/crates/forge-agent-interface/src/deck_dto.rs
```diff
@@ -6,14 +6,24 @@
...
 
-/// Pure printing identity. The single source of truth — the TS mirror
+/// Pure printing identity, the single source of truth shared across the
+/// engine ↔ UI boundary. TS mirror at `src/types/manabrew.ts`. `id` is
+/// only set for live game-state cards (UUID for zone tracking) and
+/// omitted from the wire for static identity refs (saved decks, draft
+/// pools). `foil` is the only per-copy runtime flag — rarity / colors /
+/// type-line / images all come from Scryfall on the UI side and from
+/// the engine's registry on the Rust side, never echoed through this
+/// struct.
```
anacletos: Comment

forge-engine/crates/forge-agent-interface/src/deck_dto.rs
```diff
@@ -6,14 +6,24 @@
...
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CardIdentity {
+    #[serde(default, skip_serializing_if = "String::is_empty")]
```
anacletos, line 12: Concerning? Can id be empty? I don't think it should ever be

forge-engine/crates/forge-foundation/src/edition/card_edition.rs
```diff
@@ -91,13 +92,15 @@ impl CardEdition {
     }
 
     pub fn to_sealed_template_named(&self, variant: Option<&str>) -> Option<SealedTemplate> {
+        // Forge legacy: a handful of edition files (Strixhaven, Innistrad
+        // Double Feature) stash the draft recipe under `DraftBooster=`
+        // instead of `Booster=`. Treat it as the primary booster when no
+        // explicit `Booster=` is present so those sets still resolve a
+        // template instead of falling back to the generic recipe.
```
anacletos: Comment

forge-engine/crates/forge-limited/src/booster_draft.rs
```diff
@@ -65,23 +65,70 @@ struct SeatSnapshot {
 }
 
 impl BoosterDraft {
+    /// Single-human draft (seat 0 is the human, the rest are AI). Matches
+    /// the original Forge model and the existing single-player UI.
```
anacletos: Comment

forge-engine/crates/forge-limited/src/booster_draft.rs
```diff
@@ -65,23 +65,70 @@ struct SeatSnapshot {
...
+            color_of,
+            &[(0, "You".to_string())],
+        )
+    }
+
+    /// Multi-human draft. `humans` lists every seat that should be a
+    /// human player along with its display name; every other seat in
+    /// `0..pod_size` becomes an AI seat. Used by the multiplayer draft
+    /// host to seat each connected player at their own index.
+    pub fn with_human_seats(
```
anacletos, line 87: In general, I wonder if this is the right layer to be concerned about humans / AI.
From `forge-limited` point of view, my intuition says it shouldn't. I'd rather have a list of "agents". Then the runtime resolves them to the right one (human, AI, network, etc..).

forge-engine/crates/forge-limited/src/booster_draft.rs
```diff
@@ -65,23 +65,70 @@ struct SeatSnapshot {
...
+        let mut seats = Vec::with_capacity(pod_size);
+        let mut ai_iter = BoosterDraftAI::build_ai_seats(
+            pod_size - humans.len(),
+            0, // numbering reassigned below
             ranker,
             color_of,
-        ));
+        )
+        .into_iter();
+        for seat_idx in 0..pod_size {
```
anacletos: Kind of like the same thing, "building AI seats" is likely a bad abstraction. The draft player should be a generic draft agent, regardless of the internals of it. Having branch logic here depending on agent type is not wise imo

forge-engine/crates/forge-server/src/connection.rs
```diff
@@ -573,6 +582,8 @@ fn handle_client_message(
                 format,
                 hosted,
                 engine,
+                draft_config,
+                sealed_config,
```
anacletos, line 586: We can discuss about this. But what is the meaning of "engine" in a draft room?

forge-engine/crates/forge-server/src/connection.rs
```diff
@@ -752,20 +763,29 @@ fn handle_client_message(
...
+                Ok(started) => {
                     info!(
                         "[game] game started in room {} | order: {:?}",
-                        &room_id[..8],
-                        player_order
+                        &started.room_id[..8],
+                        started.player_order
                     );
+                    // RoomUpdate first — clients keyed on room.format
+                    // (mp draft nav) need the flip before GameStarted.
```
anacletos: Comment

forge-engine/crates/forge-server/src/lobby.rs
```diff
@@ -226,11 +236,21 @@ pub fn set_deck_selection_sync(
     Ok(room_id)
 }
 
+pub struct StartedGame {
+    pub room_id: String,
+    pub player_order: Vec<String>,
+    pub player_decks: Vec<PlayerDeckInfo>,
+    pub starting_life: i32,
+    /// Post-flip snapshot — broadcast as `RoomUpdate` so clients see
+    /// the resolved format (`Any` → `Draft`) without polling.
```
anacletos: This comment confuses more than it helps

forge-engine/crates/forge-server/src/room.rs
```diff
@@ -93,7 +100,12 @@ impl Room {
     }
 
     pub fn all_ready(&self) -> bool {
-        self.players.len() >= 2 && self.players.iter().all(|p| p.ready)
+        // Draft/Sealed allow a solo human + bot fill.
+        let min_players = match self.format {
+            GameFormat::Draft | GameFormat::Sealed => 1,
+            _ => 2,
+        };
+        self.players.len() >= min_players && self.players.iter().all(|p| p.ready)
```
anacletos: I'm not sure I get this change
The previous code ensures you're not "alone" in a lobby.
The comment says this allows human + bot, but the previous code did that just fine.

forge-engine/crates/forge-server/src/connection.rs
khaliostr: Fair question. The semantic: `engine` on a draft room is the engine for the **match phase** that follows the draft, not for the draft itself.

The draft is a multiplayer protocol over the relay — wasm `forge-limited::BoosterDraft` runs on the draft host (always a wasm client), and picks ride as `draft-v1` envelopes. None of that touches the room's `engine` field; during the draft phase it's effectively dormant.

When the draft completes, the room recycles to `Lobby` + `Any` per the post-#82 lifecycle (`910996de`), and players hit Play Match with their drafted decks. **That's** when `engine` matters: a `Java` room had a hosted node attached at creation time, so the post-draft match is node-hosted; a `Wasm` room is client-hosted, same as today's play-vs-ai.

For this PR the "Forge (hosted)" option is a disabled placeholder, so in practice `engine === Wasm` on draft rooms. The field is there as the seam for #84.

Maybe cleanest framing: a draft room is a regular play room whose lifetime happens to include a draft phase first. `engine` is a room-level property; "draft" is one of the things that can happen in its lifetime (similar to how Sealed is now also a phase a room can go through). The engine determines who hosts when actual match play begins.

forge-engine/crates/forge-agent-interface/src/deck_dto.rs
fedepoi: Intentional. `CardIdentity` carries two wire shapes:

- **Static identity ref** (saved decks, draft pools, sideboard entries) — no `id`, because it isn't tied to a live game instance. This is the printing identity, period.
- **Live game-state card** — `id` is a UUID the engine assigns for zone tracking.

`skip_serializing_if = "String::is_empty"` keeps both shapes on one struct. A type-level split (`CardIdentityRef` vs `CardIdentityLive`) is cleaner but ripples through every wire boundary (deck IO, draft DTO, game-state DTO, TS mirror) — out of scope for this PR. Happy to file as a follow-up if you'd rather have it broken out.

forge-engine/crates/forge-limited/src/booster_draft.rs
fedepoi: Agree. The right shape is `with_seats(seats: Vec<LimitedPlayer>)` with the caller wiring whichever agent each seat needs — `forge-limited` shouldn't know about human-vs-AI. Mirrors Forge's `IGameEntity` flow.

Bigger refactor than this PR's scope (single-player path + `BoosterDraftAI::build_ai_seats` move out + 37 test migrations). Filed as #87 with the proposed API and migration sketch; not blocking this one.

forge-engine/crates/forge-limited/src/booster_draft.rs
fedepoi: Same theme — covered by #87.

forge-engine/crates/forge-server/src/room.rs
fedepoi: Bots aren't seated in `self.players` — they're synthesised at draft start from the room's `bot_fill` config (see `BoosterDraft::with_human_seats`). So a solo human + bot-fill Draft pod is `players.len() == 1` here, and the old `>= 2` gate would block the Ready transition for that case.

The relaxed minimum unblocks solo + bots; total seat count for the actual draft is enforced separately at draft start. The misleading comment is gone in 0cac6a205 — code stands on its own.

khaliostr (dismissed):

khaliostr (approved):

PR #82, opened 2026-05-27
feat(node): hosted java-forge play-vs-ai — espresso concurrency, lobby, engine select

emanueleDiVizio:
> recreated from #54 — a force-push wedged the old head pointer so it wouldn't reopen. same branch, same commits.

## Summary

the full hosted **Java Forge "Play vs AI"** feature for the web client. started as the espresso concurrency spike; the stacked lobby redesign (#78) is now merged in here, so this is the complete thing.

**concurrency (the original blocker).** forge keeps process-global statics (`MyRandom`, the `maxId` counters), so a second `startGame` corrupts the first live game. each game now runs in its own graalvm **espresso** context — isolated statics, one shared jvm, no AOT and no forge changes. `java-espresso-host` (`ManaBrewEspressoAdapter`) is the host router (same json api), with a pre-warmed pool + background replenish; single-use by default, in-place reuse opt-in. a `JavaEngine` actor owns the jvm and services session-tagged commands so games run concurrently. ~260 mb/context.

**room lifecycle.** node hosts games through the shared actor (not a per-game jvm), advertises `capacity` on the heartbeat, and recycles rooms: game-over / player-leave / disconnect-grace tear the game down and reset the room (`EndGame` → `end_game_sync`); cancelled games abort-close their context (never reuse a context whose guest thread may still be blocked).

**hosted lobby redesign (from #78).**
- **any-format single pool** — rooms are created format-agnostic (`GameFormat::Any`); the format is chosen at game start (rides on the `startGame` payload) and resets to Any on reuse. one pool for play-vs-ai + human tables.
- **engine selection** — `RoomInfo`/`CreateRoom` carry an `EngineKind` (Wasm|Java); lobby shows every Lobby room with an engine badge (Forge/Rust) + a ManaBrew badge for hosted rooms, and hides in-progress rooms. play-vs-ai opens a **modal** to pick Forge (hosted) vs Rust (wasm) — replaces the old experimental settings toggle.

**prompt rendering.** prompt cards resolve their art via the DTO (setCode/cardNumber/owner) like the rust path — not a name hack; reveal modal capped + scrollable; battlefield target candidates clickable.

## Why

infra won't allow unbounded node processes, so in-process isolation (espresso) is the path to concurrent hosted play. rooms shouldn't be pinned to a format at creation, and hosted play should share one pool with the engine surfaced, not a hidden per-format set.

## Test plan

verified end-to-end on the local docker stack (relay + espresso node):
[...]

PR #88, opened 2026-05-30
refactor(node): drop espresso, run java-forge on bare hotspot

fedepoi:
## Summary

- Replace GraalVM/Espresso-based per-game context isolation with a pool of bare-HotSpot `java` subprocesses (line-delimited JSON-RPC over stdin/stdout, modelled on `forge-parity`'s `--server` pool). Each room owns one JVM for its lifetime, so Forge's static counters and the per-game `ParityReset` can't race across games. One node hosts `SELF_HOSTED_NODE_MAX_GAMES` concurrent rooms.
- Add `--interactive-server` mode to `forge-harness/Main.java`: dispatches each JSON request to the existing `ManaBrewEngineAdapter` (startGame / submitAction / getPrompt / getSnapshot / getGameOver / endGame / abortGame) plus `reset` (ParityReset + ImageKeys.clearCaches + System.gc) and `quit`. Reuses the same `System.out → null sink` redirect the parity `--server` mode uses.
- Add `ManaBrewEngineAdapter.abortGameJson` to mirror the hard-abort path.
- Move `MyRandom.setRandom` inside `ManaBrewInteractiveSession`'s game-thread runnable so the seeded RNG actually reaches the thread that runs the match (the `ThreadLocal<Random>` in this fork made the previous caller-thread set a no-op).
- New `SubprocessBridge` + `JavaEnginePool` in the Rust backend. Stderr drain thread, stdout reader thread feeding an mpsc channel with `recv_timeout` so a hung java child can't block a room thread, graceful shutdown (quit + 5s grace, kill-on-timeout), session-id-collision warn + release, defensive non-JSON stdout skip.
- `JavaEngineHandle` keeps the same public surface as before — internals swap the j4rs worker-thread + mpsc model for direct calls on a per-session `Arc<Mutex<SubprocessBridge>>`. Per-room calls no longer serialise through one worker.
- Delete `java-espresso-host/`, the `java-espresso` Cargo feature, the espresso classpath/pool/reuse helpers, all `SELF_HOSTED_NODE_ESPRESSO_*` env vars.
- Drop the `j4rs` Cargo dep + the `default-jdk-headless` install from the `rust-builder` Dockerfile stage. Runtime stays Temurin 21 JRE because we still spawn `java` for each game.
- Compose: `SELF_HOSTED_NODE_MAX_GAMES` defaults to 2 locally; staging stays at 1 for the 3 GB box.

## Why

Espresso was chosen in #82 to give every game its own copy of Forge's process-global statics (per `a1ea74cf8`), so multiple rooms could share one JVM. The cost was running every card script on the Truffle interpreter (~10× slower than HotSpot) and ~50s `FModel.initialize` + ~260MB RSS per cold context — bad enough that staging is pinned at `SELF_HOSTED_NODE_MAX_GAMES=1`.

The cleaner answer (per khaliostr's review on this PR) is to mirror what `forge-parity` already does: spawn one bare-HotSpot `java` subprocess per game from a warm pool. Process isolation means Forge's static counters and `ParityReset.resetAllIdCounters` (called per `startGame` at `ManaBrewEngineAdapter:77`) can't corrupt a neighbour, no `forge-game` patch is needed, and one node hosts N concurrent rooms safely. Memory per game is one JVM + one cardDB; staging stays at 1 process and is identical to the current footprint. Larger boxes can lift `SELF_HOSTED_NODE_MAX_GAMES` and pay one extra warm JVM per concurrent slot.

## Test plan

- `cargo check -p self-hosted-node` (no features) — pass.
- `cargo check -p self-hosted-node --features java-forge` — pass.
- `cargo fmt --check -p self-hosted-node` — clean.
- `mvn -pl forge-harness -am -DskipTests -Dcheckstyle.skip compile` — pass.
- Concurrent self-play smoke: `SELF_HOSTED_NODE_JAVA_CONCURRENT_GAMES=2` ran two games end-to-end in two separate JVMs (each pre-warmed) in ~13s wall, both completed clean. Re-ran after the timeout/lifecycle fixes — same result.
- Hosted-smoke workflow run — TODO before merge (label `hosted-smoke` on this PR).

Pre-existing clippy errors in `forge-engine/src/svar/mod.rs` and `forge-engine/src/staticability/layer.rs` are unchanged by this diff; `yarn lint:rust` would surface them on `main` too.

## Build artifacts


khaliostr:
**Great rollback — espresso teardown is clean and the `MyRandom` thread fix is exactly right.** Before this lands I'd like to make the case for a different concurrency model, because I think thread-per-session has a latent hazard that process-per-game sidesteps entirely — and we mostly already have the code for it.

**The hazard:** `ManaBrewEngineAdapter.startGame` still calls `ParityReset.resetAllIdCounters()` (line 77). `AtomicInteger` makes `nextId()` race-free, but the *reset* is global — when game B starts and resets `maxId` to 0 while game A is still live, both games hand out overlapping IDs. Atomicity doesn't cover a global reset. That's the `MAX_GAMES=2` case still marked TODO in your test plan, and I suspect it bites there. Keeping thread-per-session means either dropping the per-game reset (→ non-deterministic, unbounded IDs) *and* proving nothing cross-game is keyed by those IDs — and re-proving it on every upstream Forge merge, since any new gameplay-touched mutable static reintroduces the race.

**The shape I'd suggest instead: one bare-JVM *process* per game, from a warm pool. This is exactly what the parity harness already does** — `forge-parity` spawns bare-JVM `--server` processes, talks line-delimited JSON over stdin/stdout, and pools them in `JavaServerPool`. Process isolation makes the whole shared-static question disappear: each process owns its counters, resets freely (deterministic per game), can't corrupt a neighbor, and crashes in isolation. No `forge-game` changes at all, so no submodule PR to carry across upstream rebases. Cost is memory (N× `FModel`), bounded by a pool sized to `MAX_GAMES` — staging stays at 1, so identical footprint there.

It slots in behind the `JavaBridge` trait you already have, so `run_hosted_engine_game`'s loop doesn't change — only what's behind the trait. And most of the moving parts are liftable from `forge-parity`:

- **`JavaServer::spawn` + `resolve_java_bin` + the stderr-drain thread** (`forge-parity/src/java_bridge.rs`) — the process-spawn and stdio plumbing is already written; reuse as-is.
- **`JavaServerPool`** (`forge-parity/src/runtime.rs`) — warm pre-spawn, checkout-a-free-one, return on completion; the pool is already there.
- **A `SubprocessBridge` impl of `JavaBridge`** — thin: each of the 8 trait methods is one JSON-RPC round-trip (replacing the j4rs JNI call); quit+kill on `Drop`. Also drops the `j4rs` build-time JDK dependency.
- **The only genuinely new bit on the Java side** is an `--interactive-server` mode: same stdout-capture trick as your `--server`, but the stdin loop dispatches to the existing `ManaBrewEngineAdapter` methods (`startGameJson` / `getPrompt` / `submitAction` / `isGameOver` / `endGame` / `reset`) instead of parity's deterministic one-shot. FModel loads once per process; `reset` between games is safe because the process is idle and owns its counters.

Your espresso teardown + temurin image carry over wholesale either way. WDYT — worth going process-per-game while we're already in here, given the harness gives us most of it for free?

fedepoi:
Pushed the process-per-game pool migration on top of the earlier commits, per your review.

**Java**
- Added `--interactive-server` to `forge-harness/Main.java`: line-delimited JSON-RPC over stdin/stdout dispatching to the existing `ManaBrewEngineAdapter`. Reuses the same `System.out → null sink` redirect the parity `--server` mode uses so Forge's stray `println`s can't corrupt the protocol. `reset` runs `ParityReset.resetAllIdCounters` + `ImageKeys.clearCaches` + `System.gc`. `quit` exits.
- Added `ManaBrewEngineAdapter.abortGameJson` to mirror the hard-abort path.

**Rust**
- New `SubprocessBridge` implements the existing `JavaBridge` trait by spawning `java -jar harness --interactive-server` and round-tripping one JSON-RPC request per method. Stderr drain thread, graceful shutdown (quit then kill-on-timeout), `is_alive` check — mirrors `forge-parity/src/java_bridge.rs::JavaServer`.
- New `JavaEnginePool` pre-spawns `MAX_GAMES` bridges, hands one out per `startGame` for the lifetime of the session, and on `endGame` runs `reset` then returns it to the pool. Dead bridges drop (Drop kills the child) and the pool respawns to keep size.
- `JavaEngineHandle` keeps the same public surface as before — internals swap the j4rs worker-thread + mpsc model for direct calls on a per-session `Arc<Mutex<SubprocessBridge>>`. Per-room calls no longer serialise through one worker.
- Dropped the `j4rs` Cargo dep entirely and the `default-jdk-headless` install from the `rust-builder` Dockerfile stage. Runtime stays Temurin 21 JRE.
- Defensive: `call()` skips non-JSON stdout lines so a stray Forge `println` can't kill a session.

**Concurrency**
- `SELF_HOSTED_NODE_MAX_GAMES` is now the pool size (2 in local, 1 in staging until the box sizes up). The `MAX_GAMES=1` pin can go away once you're comfortable.

**Smoke**
Ran two concurrent self-play games on the pool with `SELF_HOSTED_NODE_JAVA_CONCURRENT_GAMES=2`; both completed end-to-end in two separate JVMs in ~7s after the pool warm-up.

Latest commit: `afa1e423d`.

khaliostr (dismissed):
This is exactly the migration I was hoping for — really nicely done.

The concurrency hazard is gone by construction now: each game owns its own process and static counters, and `reset` only runs when the bridge is idle on release, so the `resetAllIdCounters` collision simply can't happen. Love that it let you **drop the `forge` submodule `AtomicInteger` patch entirely** — no engine changes to carry across upstream merges.

The pool is clean (`free`/`in_use`, health-checked `release` with respawn, `MAX_GAMES` = pool size so it can't oversubscribe), and the `recv_timeout` + graceful `quit`→kill lifecycle means a wedged JVM fails safe instead of stalling a room. Parity green at 346/346 and the two-game concurrent smoke is the right proof.

LGTM 🚀

anacletos (approved):

Issue #77, opened 2026-05-26
hosted engine: game soft-locks when AI doesn't block lethal (combat flow stalls)


emanueleDiVizio:
playing vs ai (hosted java), attacked for lethal and the ai didn't declare blocks — the game then **stuck** instead of resolving combat / ending. either SimpleAi isn't returning a block decision, or the declare-blockers step in the hosted path doesn't advance when the bot passes, leaving the game waiting on a decision that never comes. soft-locks the game.

severe (it halts the game). part of the hosted-engine fidelity cluster (#69/#70/#74/#75/#76). found playtesting hearthhull commander locally vs the node bot.

anacletos:
Fixed

PR #89, opened 2026-05-30
fix(node): pass instead of stalling when a bot can't answer a prompt

khaliostr:
## Summary

- The relay bot (`forge-bot`) silently dropped any prompt it couldn't turn into an action — `AgentPrompt` deserialization failure, agent returning no decision, or action serialization failure each `return`ed without sending a response.
- All three paths now log at `error` level (including the offending prompt JSON on a deserialize miss) and fall back to a pass response both the rust engine and the java harness accept.

## Why

Fixes the soft-lock in #77. When a bot can't answer a prompt, the hosted Java engine blocks forever on that seat — the game hangs with no way forward. The reported repro is attacking a hosted-AI opponent for lethal: the AI's `declareBlockers` prompt went unanswered and combat never resolved.

This is a **defensive fix, not a root-cause fix**. I traced the full declare-blockers path (Java `declareBlockers` → `awaitBlockers` → `normalize_java_prompt` → bot `handle_envelope` → `SimpleAi` → `translate_java_player_action`) and it's correct on inspection — `SimpleAi` always returns `Some(DeclareBlockers { .. })`, so the most likely culprit is a silent `AgentPrompt` deserialize failure in `handle_envelope` that the old code swallowed with only a `warn!` and an empty return. This patch turns that (and the other two silent drops) into a loud log plus a game-advancing pass, so the game no longer hangs and the failing prompt is captured for a targeted follow-up.

## Test plan

- `cargo fmt --check` on the changed file — clean
- `cargo clippy -p forge-bot` — no new warnings from this change (the remaining warnings are pre-existing, in `forge-foundation` / `forge-carddb` / engine effect resolvers)
- **Manual (pending):** reproduce #77 (attack hosted-AI for lethal), confirm the game advances instead of hanging, and capture any `error!`-level bot log to pin the underlying prompt-shape failure for the root-cause PR.

## Build artifacts

### DIFF
--- a/forge-engine/crates/forge-bot/src/state.rs
+++ b/forge-engine/crates/forge-bot/src/state.rs
@@ -4,7 +4,7 @@ use forge_agent_interface::prompt::AgentPrompt;
 use forge_agent_interface::protocol::{ClientMessage, ServerMessage, StateEnvelope};
 use serde::{Deserialize, Serialize};
 use serde_json::json;
-use tracing::{debug, warn};
+use tracing::{debug, error, warn};
 
 use crate::agent::{AgentKind, BotAgent};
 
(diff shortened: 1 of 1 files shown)


anacletos (changes requested):
Dangerous cos this shouldn't be possible to happen. With this we wouldn't even notice

khaliostr:
fair, but it is happening and haven't found root cause nor logs. this would at least log it!

anacletos (approved):

anacletos (changes requested):
But this works? If it passes the prompt can still be invalid and crash the engine later?

khaliostr:
it would basically ignore it, log it and pass. anyhow closing this hoping it's fixed after latest dto changes. can pick up again if needed

PR #94, opened 2026-06-01
fix: hosted play-vs-ai — let the user set format and start, not the node

khaliostr:
## Summary

Make hosted "Play vs AI" driven by the **user**, not the node:

- **relay** (`lobby.rs`): `set_format_sync` gates on `is_controller` instead of `is_host`, matching `start_game_sync`.
- **node** (`self-hosted-node/main.rs`): drop the `startGame` relay-message handler — the node no longer mediates start.
- **web** (`hostedAiPlay.ts`): the human (the controller) calls `server.startGame({ format })` directly instead of routing a relay envelope through the node.

## Why

The first seated user is the **controller** (drives format/start); the engine holder is the **host**. They coincide in a self-created room but diverge in a `hosted` (self-hosted-node) room: the node holds the engine as a *non-seated observer* (host), and the first human to join is the controller. So:

- `set_format` required `is_host` → the human couldn't set the format (`not_host`).
- start went through the node, which sent `StartGame` as a non-controller → `not_host`.

Now the controlling player sets the format and starts the game directly, and the node just reacts to `GameStarted` to spin up the engine. Keeps Anacleto's "first user is the controller" model; normal rooms unaffected (controller == host there). The node's `auto_start` self-play path is untouched (gated by `config.auto_start`).

## Test plan

- `cargo test -p forge-server` — `first_user_is_controller_in_hosted_room` passes.
- `cargo check -p self-hosted-node` — 0 errors.
- `tsc` — clean.
- Manual: hosted Play-vs-AI — change format in lobby, start, play a game to completion.

## Build artifacts

### DIFF
--- a/forge-engine/crates/forge-server/src/lobby.rs
+++ b/forge-engine/crates/forge-server/src/lobby.rs
@@ -279,7 +279,7 @@ pub fn set_format_sync(
             .get_mut(&room_id)
             .ok_or_else(|| ServerError::RoomNotFound(room_id.clone()))?;
 
-        if !room.is_host(player_id) {
+        if !room.is_controller(player_id) {
             return Err(ServerError::NotHost);
         }
 
--- a/forge-engine/crates/forge-server/src/room.rs
+++ b/forge-engine/crates/forge-server/src/room.rs
@@ -310,3 +310,34 @@ impl Room {
         }
     }
 }
+
+#[cfg(test)]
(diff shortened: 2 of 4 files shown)


anacletos (dismissed):

anacletos (dismissed):

anacletos (approved):
lgtm

PR #103, opened 2026-06-01
fix(ci): keep windows build green when discord rejects embed

fedepoi:
## Summary
- Switch Windows `Build changelog` step from `pwsh` + `Out-File` to bash, mirroring the macOS step byte-for-byte — removes CRLF/Out-File divergence between the two jobs.
- Replace `curl -fsS` in all three Discord-notify steps (mac, windows, release) with a `post_discord` bash helper that retries 3× with exponential backoff, captures the response body via `-o`/`-w`, logs it on every non-2xx via `::warning::`, and always returns 0.

## Why
Run [`26769518240`](https://github.com/witchesofthehill/manabrew/actions/runs/26769518240) (workflow_dispatch on main) failed the whole `windows-exe` job at the "Post Discord embed (Windows artifact)" step with `curl: (22) The requested URL returned error: 400`, even though the `.exe` and `.msi` had already uploaded successfully. `curl -fsS` suppressed Discord's response body so the log only showed `error: 400`.

The macOS job ran the same code against the same channel from the same bot 8 min earlier and succeeded — full 200 JSON in the log. The only meaningful divergence is the changelog step's shell (`pwsh` on windows, `bash` on mac). PowerShell's `Out-File -Encoding utf8 -Append` writes CRLF terminators, which can leak `\r` into the multi-line env var delivered to the next step → ends up inside the JSON `description` → Discord rejects control chars other than `\n`/`\t`. Aligning Windows on bash removes that hazard.

Independent of that root cause: a Discord notification failing should never fail the artifact build. The notification is auxiliary; the installer is the deliverable. The new helper retries transient errors, surfaces Discord's actual error body next time it rejects something, and degrades to a workflow warning instead of a red X.

## Test plan
- [x] `npx prettier --check .github/workflows/release-artifacts.yml` — clean.
- [x] `yarn lint:all` — only the pre-existing eslint errors on Forge `.tsx` Tiled tilesets remain (unchanged on `main`).
- [ ] After merge, dispatch a `workflow_dispatch` run with both platforms enabled and confirm:
  - macOS job still green, embed still posts.
  - Windows job green; embed either posts cleanly or logs `::warning::` with Discord's actual error body (enabling targeted follow-up).

## Build artifacts
- [x] Build macOS .dmg on merge to main
- [x] Build Windows .exe / .msi on merge to main

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -229,11 +229,40 @@ jobs:
               }]
             }')
 
-          curl -fsS -X POST \
-            -H "Authorization: Bot $BOT_TOKEN" \
-            -H "Content-Type: application/json" \
-            -d "$PAYLOAD" \
-            "https://discord.com/api/v10/channels/$CHANNEL_ID/messages"
+          post_discord() {
+            local attempt=0
+            local max_attempts=3
+            local delay=2
+            local body http
+            body=$(mktemp)
+            while :; do
+              attempt=$((attempt + 1))
+              http=$(curl -sS -X POST \
+                -H "Authorization: Bot $BOT_TOKEN" \
+                -H "Content-Type: application/json" \
+                -d "$PAYLOAD" \
+                -o "$body" \
+                -w "%{http_code}" \
+                "https://discord.com/api/v10/channels/$CHANNEL_ID/messages" || echo "000")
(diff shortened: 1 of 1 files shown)


khaliostr (approved):
awesomesauce

PR #118, opened 2026-06-04
fix(harness): emit zone for non-battlefield card targets

fedepoi:
## Summary
- Java interactive session now communicates the zone of card targets; `publishTargetPrompt` emits a `zone` field when all card candidates share a single non-battlefield zone, and the normalizer maps a non-`Battlefield` zone to `ChooseTargetCardFromZone` so the UI's zone target picker opens.
- UI: `chooseTargetCardFromZone` no longer enters board arrow-mode — it's modal-based, so the cursor-following arrow and the hidden-OS-cursor behavior no longer engage (previously the hidden cursor made the picker unclickable).
- UI: clicking the graveyard/exile zone tile while a from-zone target prompt is active for that zone reopens the picker, so closing it is no longer a dead end.

## Why
On a Java-engine game, casting a spell that targets a card in a graveyard/exile/hand (Snapcaster Mage, Regrowth, Disentomb, Reanimate, …) left the UI stuck. Root cause: the Java host always sent a generic `choose_target_card` with no zone, so the normalizer produced `ChooseTargetCard` and the zone picker (which only opens for `ChooseTargetCardFromZone`) never appeared. The valid graveyard cards had no clickable surface. This mirrors what the Rust-native engine already does via `targeting.rs:choose_target_card_from_zone`.

Two follow-up UI bugs surfaced once the picker opened: the shared `chooseTarget*` arrow flow hid the OS cursor (so the modal couldn't be clicked), and dismissing the picker left no way to bring it back.

## Test plan
- Maven harness rebuild: `BUILD SUCCESS`.
- Verified the normalizer conversion with throwaway integration tests (run green, then removed per the no-committed-unit-tests policy): graveyard → `ChooseTargetCardFromZone` with populated `zone_cards`; battlefield / no-zone → plain `ChooseTargetCard`.
- `cargo fmt --check` clean; no clippy findings from the edited Rust files.
- Frontend: `yarn typecheck`, `eslint`, and `prettier --check` clean on the changed files.
- Parity suite does not exercise this path (it uses deterministic controllers, not the interactive session host).

## Build artifacts


khaliostr (dismissed):

khaliostr:
@fedepoi is this still relevant?

PR #122, opened 2026-06-05
feat(companion): multi-touch, LifeTap layouts, commander-damage modal

fedepoi:
## Summary

**Multi-touch gestures**
- Two-finger **pinch** on free-layout tiles to **zoom and rotate** in one gesture, center-anchored (no pan), reusing the existing `{x, y, rotation, scale}` model.
- **Multiple players adjust life simultaneously** on grid tiles — added `touch-none` to the life half-buttons so the browser stops claiming a second finger as a pan/pinch.

**Table layouts (LifeTap parity)**
- Fixed the **5-player radial** layout to match LifeTap: full-width band on the far edge (180°) over a 2×2 whose left column reads from the left (90°) and right column from the right (−90°).
- Added four orientation-correct presets — **Two across** (2p), **Three sides** (3p), **Five rows** (5p, upright 2+3), **Six sides** (6p, three per side).
- Switched the **4-player default** from `four-sides` to the **2×2 Quad**; `four-sides` stays selectable.
- Seat-rotation convention documented: `0` near, `180` far, `90` left, `−90` right.

**Commander damage**
- Replaced the cramped side-popover stepper with a **centred modal** (mirrors the dice-roll dialog) holding big touch-friendly steppers per commander slot — much easier to assign on a phone. Shows source → target accents and a **Lethal** badge at 21.
- Each commander-damage **round button now has a border in the owning player's accent colour** so it's obvious whose commander each avatar is; the lethal cue moved to the damage badge (red) so the owner colour is never overridden.

## Why

On a shared paper-play device, players want to tweak their own life at the same time, orient/size their card to face them around the table, and quickly assign commander damage. The grid life buttons lacked `touch-action: none`; the 5-player layout's upright lower tiles didn't match how these apps seat players (confirmed against a LifeTap screenshot); and the old commander-damage popover was fiddly and gave no owner-identity cue.

Pinch is free-layout only. The single-finger rotate/scale/grip handles stay for mouse and precise control.

## Test plan

- [ ] Free layout: two-finger pinch zooms + rotates together; release keeps the snapped scale/rotation; no stray ±life on finger lift; single-finger tap/drag/handles still work.
- [ ] Grid layout: two players tap their tiles at once and both totals update independently.
- [ ] Layout picker: 5-player radial matches LifeTap; new presets render with correct per-card orientation and a distinct icon; 4-player defaults to Quad.
- [ ] Commander damage: tapping a source avatar opens the modal; big steppers adjust the right source/slot; target life drops in sync; Lethal badge shows at 21; source buttons are bordered in each opponent's accent colour.
- [ ] `yarn lint:all` passes (eslint + prettier + tsc verified locally).

## Build artifacts

- [ ] Build macOS .dmg
- [ ] Build Windows .exe


khaliostr (dismissed):
well done

khaliostr (approved):

PR #130, opened 2026-06-08
fix: protocol state update being deferred

anacletos:
## Summary
* `StateUpdate`, `DisplayEvent`, and `Prompt` are now different primitives and they don't share nonsense behaviour
* Updates are now correctly honored timely by all players instead of being deferred and shown at "first available prompt"
* Remove isChoosable from card dto

## Build artifacts


forge-engine/crates/forge-agent-interface/src/game_view_dto.rs
```diff
@@ -764,16 +762,30 @@ impl GameViewDto {
...
+                .map(|cid| card_to_dto(game, cid, playable_ids, "command"))
+                .collect();
             players.push(PlayerDto {
                 id: player_id_str(pid),
                 name: ps.name.clone(),
                 is_human: pid == human_player,
                 life: ps.life,
                 poison: ps.poison_counters,
-                hand_count: game.cards_in_zone(ZoneType::Hand, pid).len(),
+                hand: zone_cards(ZoneType::Hand, "hand"),
```
khaliostr, line 784: Are we sending other players' hand on the wire? seems a privileged info leak

src/components/game/prompts/promptComponents.tsx
```diff
@@ -157,37 +158,39 @@ const PROMPT_MODALS: { [T in PromptType]?: PromptComponent<T> } = {
...
-  chooseCombatDamageAssignment: ({ prompt, respond }) => (
+  chooseCombatDamageAssignment: ({ prompt, respond, ctx }) => (
     <VAssignCombatDamageModal
       attackerId={prompt.input.attackerId}
       blockerIds={prompt.input.blockerIds}
       defenderId={prompt.input.defenderId}
       totalDamage={prompt.input.totalDamage}
       attackerHasDeathtouch={prompt.input.attackerHasDeathtouch}
-      gameView={prompt.input.gameView}
+      gameView={ctx.gameView!}
```
khaliostr, line 193: ctx needs now need to be in sync with the prompt before display. is this intended?

src/platform/web.ts
```diff
@@ -145,11 +151,11 @@ class WorkerBridge {
         Atomics.store(this.gameSignal, 0, 3); // PROMPT_ACKNOWLEDGED
         Atomics.notify(this.gameSignal, 0);
 
+        console.log("[transport←sab/local] received:", jsonStr);
```
khaliostr: noisy, could be removed

forge-engine/crates/self-hosted-node/src/main.rs
```diff
@@ -1011,16 +1011,26 @@ fn route_remote_response(engine_session: &SharedEngineSession, state: &Value) {
...
-    remote_prompt_rx: std_mpsc::Receiver<(usize, AgentPrompt)>,
+    remote_prompt_rx: std_mpsc::Receiver<(usize, AgentMessage)>,
 ) {
     thread::spawn(move || {
-        while let Ok((player_index, prompt)) = remote_prompt_rx.recv() {
-            let Ok(state) = serde_json::to_value(StateEnvelope::Prompt {
-                for_player: player_slot(player_index),
-                prompt: serde_json::to_value(prompt).unwrap_or(Value::Null),
-            }) else {
+        // State/Display have no `forPlayer` and are identical for every player,
```
khaliostr, line 1017: I am not sure it's identical for every player as isHuman can be false for an ai player

forge-engine/crates/forge-agent-interface/src/game_view_dto.rs
anacletos: We were doing that already cos everyone was getting everyone's prompts, only filtered client side.

forge-engine/crates/self-hosted-node/src/main.rs
anacletos: isHuman is not a thing and the State / Display broadcast must be identical for every agent, theres no real reason not to

src/components/game/prompts/promptComponents.tsx
anacletos: yes it's intentional. Gameview was being broadcasted via prompt, which is a bad and brittle way to send state updates around. This was the main reason state updates were actually delayed and players would see different things on screen at a given moment.
Making sure the state is up to date is quite trivial

src/platform/web.ts
anacletos: Removed

forge-engine/crates/forge-agent-interface/src/game_view_dto.rs
khaliostr: LOL okay then

khaliostr (approved):

PR #132, opened 2026-06-08
feat(ai): phase-ai-derived AI — hand-fair protocol bot + in-process forge-engine searcher

khaliostr:
## Summary

Two phase-ai-derived deliverables, both adapted from [phase-rs/phase `phase-ai`](https://github.com/phase-rs/phase/tree/main/crates/phase-ai) (Apache-2.0, used with the maintainer's permission; credited in `THIRD-PARTY-NOTICES.md`).

**1. Protocol-only bot — `forge-bot/src/agent/phase_eval/` (done).**
A hand-fair `PhaseEvalAi` that plays over the agent wire protocol (protocol-v2):
- Caches the latest `StateUpdate` view via a new `BotAgent::observe` hook (decision prompts no longer carry the view); `state.rs` routes `StateEnvelope::State` to it.
- Eval-driven **combat** (attacker/blocker selection with a 1-ply damage-resolution lookahead — deathtouch, first/double strike, indestructible, flying/reach) and hostile/friendly **targeting**.
- **Fairness:** reads only the bot's own hand + public zones and **never inspects opponents' hand contents** — only the public hand count — even though protocol-v2 now ships every hand in the DTO.
- Non-combat/targeting prompts delegate to `SimpleAi`. Selectable on the node via `SELF_HOSTED_NODE_BOT_AGENT=phaseEval` (default `Simple`).

**2. In-process searcher foundation — `forge-ai/` crate (WIP).**
A new crate (depends on `forge-engine-core`) for an in-process AI with direct `GameState` access — clones the state and drives forge-engine forward to look ahead, no ported simulator.
- **Landed:** evaluation ported onto `GameState`/`Card` (board/creature eval read by controller, keyword-bonus weights, threat assessment, strategic intent). Verified compiling.
- **Remaining:** a `SearchAgent` that caches state via `snapshot_state`, then clones + simulates each candidate forward, scoring leaves with this eval. Needs one thin engine hook (`GameLoop::advance_from_current`) because `run_turn` restarts the turn. Rust-backend only. Rollouts will determinize opponents' hidden cards to keep the no-cheating rule.

## Why

The existing Rust bot (`SimpleAi`) is a first-legal-choice placeholder; the real opponent today is Forge's Java AI (the JVM dependency the Rust port exists to shed). The protocol bot gives the relay path a competent, fair opponent now; the in-process searcher gives the Rust engine true lookahead.

## Test plan

[...]

PR #135, opened 2026-06-08
feat(ai): in-process searching AI (forge-ai) — clone-and-simulate over forge-engine

khaliostr:
## Summary

WIP. A new `forge-ai` crate (depends on `forge-engine-core`) for an **in-process** AI that, unlike the protocol-only bot in #132, has direct `GameState` access — so it can clone the state and drive forge-engine forward to look ahead (no ported simulator; our engine does the rules).

- **Landed:** evaluation ported onto `GameState`/`Card` (board/creature eval, keyword-bonus weights, threat assessment, strategic intent), adapted from phase-rs/phase `phase-ai` (Apache-2.0, credited in THIRD-PARTY-NOTICES). `board_stats` reads by controller. Compiles against the engine.
- **Next:** a `SearchAgent` (`PlayerAgent`) that caches the live state via `snapshot_state`, then in `choose_action` clones it, runs `GameLoop::run_turn` forward per candidate with a rollout policy, and scores leaves with the eval. Plus a clone-cost spike to size the time budget.

## Why

Gives the Rust-engine path a real lookahead opponent. Rust-backend only (Java's state is in the JVM). Fairness: rollouts will determinize opponents' hidden cards rather than peek.

## Test plan

- `cargo check -p forge-ai` clean (verified, dedicated target).
- Search + spike: pending — clone cost (`GameState` is a deep clone incl. libraries) gates search width/depth.

## Build artifacts

- [ ] Build macOS .dmg
- [ ] Build Windows .exe

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ members = [
     "forge-engine/crates/forge-cardset-archive",
     "forge-engine/crates/forge-limited",
     "forge-engine/crates/forge-bot",
+    "forge-engine/crates/forge-ai",
 ]
 
 [workspace.package]
--- a/THIRD-PARTY-NOTICES.md
+++ b/THIRD-PARTY-NOTICES.md
@@ -4,6 +4,22 @@ This project incorporates and derives from third-party software. The notices
 below identify each component, its origin, its license, and the form in which
 this project uses it.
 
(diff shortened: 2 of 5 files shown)


khaliostr:
Folded into #132, which is the phase-ai search+eval PR.

PR #138, opened 2026-06-08
feat(ai): Forge's native AI in the bot seat (in-engine opponent)

khaliostr:
## Summary

Puts **Forge's own mature AI** (`PlayerControllerAi` + `AiController`) in the hosted-game bot seat, replacing today's opponent — a first-legal-action/pass stub. Native to the engine hosted games already run on; the strongest opponent available.

**Design A — "bot-as-presence"** (chosen after mapping the room lifecycle): `run_bot` still joins/sets-deck/readies so the lobby's start condition holds, but the node marks its seat `ai: true` so the **engine drives it in-engine via Forge's AI**. The engine emits no prompts for that seat, so `run_bot` idles harmlessly — disconnect-grace and start logic untouched, no relay/protocol changes.

### What's here (Java + Rust, complete)
- **Harness:** `ManaBrewEngineAdapter` assigns `LobbyPlayerAi` to any seat with `ai:true` (swap is local — same `IGameEntitiesFactory` interface; `forge-ai` already a dep). `PlayerConfig` parses `ai`.
- **Node:** `config.forge_ai` (env `SELF_HOSTED_NODE_FORGE_AI`, **default off**). When on + a bot is enabled, the node finds the bot's seat by `bot_username` in `player_order` and sets `PlayerConfig.ai=true` in the `StartGameRequest`. `run_bot`/`spawn_bot` unchanged.

### Why it's safe across concurrent rooms
Each hosted game runs in its own JVM subprocess (`JavaEnginePool`), so the per-game `ai` flag isolates cleanly; the node IDs each room's bot by its per-room labeled username. Uses the rules-based AI (`LobbyPlayerAi(name, null)`, not `USE_SIMULATION`), so per-decision CPU stays modest (bounded by `MAX_GAMES`).

### The engine still prompts the human
`match.startGame()` is Forge's native loop — it calls each player's `PlayerController` in turn. The human's controller blocks on the relay (gets prompts); the AI's resolves in-engine. This is exactly how Forge's own GUI runs human-vs-AI.

## Test plan

- `cargo check -p self-hosted-node` clean (default **and** `--features java-forge`), via the dedicated target dir (`/tmp/mb-ai`) — the shared `target` symlink bleeds a sibling project's stale `forge-server`. fmt/clippy clean on the touched files.
- Java harness can't be `-pl` compiled offline (needs the Forge Maven reactor); CI/Docker builds it.
- **End-to-end not yet run.** Deploy path: merge → mini auto-deploys → set `SELF_HOSTED_NODE_FORGE_AI=true` (with `bot_enabled`) in `~/manabrew-node.env` → hosted games use Forge's AI.

## Trade-off / follow-up

Keeps an idle bot WebSocket open (cheap). A future opt-in **Design B** (relay carries an `as_bot` hint + phantom AI seat) could retire it, backward-compatibly.

## Build artifacts

- [ ] Build macOS .dmg
- [ ] Build Windows .exe

### DIFF
--- a/forge-engine/crates/self-hosted-node/src/config.rs
+++ b/forge-engine/crates/self-hosted-node/src/config.rs
@@ -21,6 +21,7 @@ pub struct Config {
     pub host_plays: bool,
     pub bot_enabled: bool,
     pub bot_username: String,
+    pub forge_ai: bool,
     pub host_deck: DeckSelection,
     pub bot_deck: DeckSelection,
 }
@@ -107,6 +108,7 @@ impl Config {
                 false,
             ),
             bot_username,
+            forge_ai: env_bool("SELF_HOSTED_NODE_FORGE_AI", "FORGE_ROOM_FORGE_AI", false),
             host_deck: load_deck_selection(&host_deck_id, host_commander),
             bot_deck: load_deck_selection(&bot_deck_id, bot_commander),
         }
--- a/forge-engine/crates/self-hosted-node/src/engine_backend/java_backend.rs
+++ b/forge-engine/crates/self-hosted-node/src/engine_backend/java_backend.rs
@@ -689,6 +689,7 @@ pub fn run_hosted_engine_game(
     decks: Vec<Deck>,
     commander_names: Vec<Option<String>>,
     local_player_index: Option<usize>,
+    ai_player_index: Option<usize>,
     starting_life: i32,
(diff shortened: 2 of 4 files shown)


khaliostr:
Folded into #139, which now targets main and contains the full Forge-AI hosted work (1v1 seat + multiplayer pods) as a single PR.

PR #141, opened 2026-06-09
fix(node): always emit game-over prompt, even if final snapshot fails

khaliostr:
## Summary

Hosted java games would reach game-over but the client never showed the result screen — it soft-locked in a dead game view (#77).

**Root cause:** in `run_hosted_engine_game_inner`, both the final state **and** the `GameOver` result-screen prompt were emitted only inside `if let Ok(snapshot) = engine.get_snapshot(...)`. If the snapshot fetch/parse failed at game-over (plausible while the session is torn down), the whole block was skipped **silently** — no state, no prompt, no log — so the client got no game-over signal.

The `GameOver` prompt carries no snapshot data, so it never needed to be coupled to the snapshot.

## Why

I traced the full path and everything else is correct: the human seat is a broadcast target; the client applies a remote game-over state unconditionally and the `gameOver` prompt when `forPlayer === myPlayerSlot` (which matches); `GameOverScreen` renders on `gameView.gameOver || promptType === "gameOver"`. The only way to get "game ended, no screen" is the emission block being skipped — i.e. the snapshot guard. Not AI-specific; this path runs for every hosted game.

## Fix

- Final state stays **best-effort** (needs the snapshot) and now **`warn!`s** when the snapshot is unavailable — which also confirms the hypothesis on the next hosted game-over.
- The `GameOver` prompt is **always** sent, so the result screen shows regardless of snapshot state.

## Test plan

- `cargo check -p self-hosted-node --features java-forge` — clean.
- Live: finish a hosted game; the result screen should appear. If the snapshot path was the culprit, the node logs the new `game over: final snapshot unavailable …` warn.

## Build artifacts
- [ ] Build macOS .dmg
- [ ] Build Windows .exe

### DIFF
--- a/forge-engine/crates/self-hosted-node/src/engine_backend/java_backend.rs
+++ b/forge-engine/crates/self-hosted-node/src/engine_backend/java_backend.rs
@@ -903,19 +903,32 @@ fn run_hosted_engine_game_inner(
 
         if engine.is_game_over(&session_id)? {
             info!("hosted java-forge session reached game over");
-            if let Ok(snapshot_json) = engine.get_snapshot(&session_id) {
-                if let Ok(raw_snapshot) = serde_json::from_str::<JavaRawSnapshot>(&snapshot_json) {
+            // Best-effort final state: it needs the snapshot, but the result
+            // screen must not depend on it. A snapshot fetch/parse failure here
+            // used to drop the whole game-over emission and soft-lock the client
+            // in a dead game view (#77).
+            match engine.get_snapshot(&session_id).and_then(|json| {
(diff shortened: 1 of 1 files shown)


anacletos (commented):
I am not convinced this was the "root cause".
Why does the snapshot fail then? That would be the root fix

khaliostr:
Gameover doesn't need any snapshot due to recent changes to protocol. It is unclear why snapshot would fail, a race is presumed (game closed before snapshot). This way the game is unblocked and the error logged.

anacletos (approved):

PR #165, opened 2026-06-12
fix(ui): smooth hand hover interactions

mermerico:
## Summary
- Decouple hand-card hit targets from enlarged hover visuals so neighboring cards stay reachable in their normal zones.
- Switch hand-card movement and scaling to time-based easing.
- Speed up the shrink animation when moving directly from one hand card to another.

## Why
Hand hover felt sticky because the enlarged visual sprite also owned pointer hit testing, so the cursor had to leave the scaled-up bounds before another card could become active. The per-frame lerp also front-loaded the movement, which made the first animation frame feel jerky.

## Test plan
- `yarn lint`
- `yarn lint:all` was attempted, but the Rust Clippy phase fails on unrelated existing findings from upstream `main` in `forge-carddb`, `forge-limited`, and `forge-server`.

## Build artifacts

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,174 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Tauri: dev (yarn)",
+      "type": "node-terminal",
+      "request": "launch",
+      "command": "yarn dev",
+      "cwd": "${workspaceFolder}"
+    },
+    {
+      "name": "Tauri: debug Rust backend",
+      "type": "lldb",
+      "request": "launch",
+      "cargo": {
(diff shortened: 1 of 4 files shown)


fedepoi (commented):
can you please add some screens, maybe a demo video of the new interaction?

anacletos:
I tested this @mermerico I think this can go in. The anim does feel better :)

anacletos (dismissed):

fedepoi (commented):
Can you please add the .vscode folder to the .gitignore? i dont think we wanna track those settings @anacletos @khaliostr WDYT?

anacletos (approved):

PR #177, opened 2026-06-12
fix(hosted): render card names in payment prompts

mermerico:
## Summary

- replace Forge `CARDNAME` and `NICKNAME` placeholders in hosted confirm-payment prompts
- keep the source card name metadata unchanged for the client

Before/After:
Confirm Payment
Confirm Payment

## Why

Forge prompt strings can contain card-name placeholders that its native client normally resolves before display. The hosted bridge was forwarding those strings directly, which made prompts such as Black Lotus sacrifice confirmation show `CARDNAME` instead of the actual card name.

## Test plan

- `yarn lint:all` fails on existing Rust clippy issues unrelated to this Java-only change (`forge-card-script`, `forge-protocol`, `forge-foundation`)

## Build artifacts

- [ ] Build macOS .dmg
- [ ] Build Windows .exe


forge-harness/src/main/java/forge/harness/host/ManaBrewInteractiveController.java
```diff
@@ -2256,6 +2257,14 @@ private static String sourceName(final SpellAbility sa) {
         return sa == null || sa.getHostCard() == null ? null : sa.getHostCard().getName();
     }
 
+    private static String sourceNamePrompt(final String prompt, final SpellAbility sa) {
+        final String name = sourceName(sa);
+        if (name == null) {
+            return prompt;
+        }
+        return prompt.replace("CARDNAME", name).replace("NICKNAME", name);
```
fedepoi: ive never seen Nickname in a prompt do we really need that replace? have you seen it ? 

code looks good

anacletos (dismissed):

khaliostr:
Hey @mermerico , nice catch on this one, the `CARDNAME` leak has been bugging me for a while. Direction is right, but I think the `NICKNAME` half isn't quite correct: right now both placeholders resolve to the full card name, but Forge treats them differently. `CARDNAME` is the full name and `NICKNAME` is just the first segment (it splits on comma, then colon, then space), so for "Rasputin Dreamweaver" the nickname is "Rasputin", not the whole thing. The canonical way Forge does it (in `SpellAbility.toUnsuppressedString`, `Trigger`, `GameAction.notifyOfValue`) is always:

```java
text = TextUtil.fastReplace(text, "CARDNAME", name);
text = TextUtil.fastReplace(text, "NICKNAME", Lang.getInstance().getNickName(name));
```

so I would suggest we resolve the nickname with `Lang.getInstance().getNickName(name)` instead of reusing the full name, otherwise a prompt with `NICKNAME` shows the long name where Forge would show the short one.

@fedepoi to your question, `NICKNAME` is actually used a lot, it shows up around 1100 times across the cardsfolder (e.g. Rasputin Dreamweaver, "if NICKNAME started the turn untapped"), so I'd keep the replace, we just need it pointing at the short name. Other than that it looks good to me, happy to approve once that's sorted!

mermerico:
Thanks @khaliostr! Done and tested.

khaliostr (approved):
Great! Thanks @mermerico , approved :)

PR #178, opened 2026-06-12
fix(hosted): preserve java game over outcomes

mermerico:
## Summary

- forward hosted Java final state and game-over prompt together before ending the room
- preserve the Java snapshot winner by reading the Forge game outcome and registered players
- suppress disconnect/abort overlays once a game-over prompt is already pending

## Why

Hosted Java games could end by broadcasting `EndGame` before the client had received the final Java state and game-over prompt. That made wins and losses look like disconnects or draws, especially when the final snapshot carried the decisive winner state.

### Game-over ordering race

The hosted Java game-over path had two independent outbound streams:

Before:
- Final Java state and the marker-only `gameOver` prompt were sent through `remote_prompt_tx`, then forwarded by `spawn_remote_prompt_forwarder`.
- The room end signal was sent through `game_over_tx`, then forwarded by `spawn_game_over_forwarder`.
- Both forwarders wrote to the same `outbound_tx`, but because they ran on separate threads there was no FIFO guarantee between the final prompt/state and `EndGame`.

That meant `EndGame` could overtake the final state or `gameOver` prompt. When that happened, the relay reset the room first and clients saw the normal game finish as an abort/disconnect instead of a result screen.

After:
- The Java game-over path bundles the final state, marker prompt, room `gameOver` relay message, and `EndGame` into one `HostedGameOver` delivery path.
- One forwarder sends those messages in the required order.
- The client-side guards suppress abort/disconnect UI once a game-over state or prompt is already pending.

This preserves the Java winner snapshot while keeping `EndGame` as the room lifecycle teardown. The native Rust hosted path still uses the older two-channel pattern; hosted AI currently runs through Java-Forge, but we should either unify both paths into one ordered hosted-engine event stream or track that as a follow-up.

## Test plan

- `cargo check -p self-hosted-node --features java-forge`
- `./node_modules/.bin/eslint src/hooks/useGameEventListeners.ts src/hooks/useMultiplayerInterruption.ts && ./node_modules/.bin/tsc -p tsconfig.app.json --noEmit`
- `yarn lint:all` fails on existing Rust clippy issues unrelated to this change (`forge-card-script`, `forge-protocol`, `forge-foundation`)

## Build artifacts

- [ ] Build macOS .dmg
- [ ] Build Windows .exe

### DIFF
--- a/forge-engine/crates/self-hosted-node/src/main.rs
+++ b/forge-engine/crates/self-hosted-node/src/main.rs
@@ -1092,6 +1092,10 @@ fn spawn_game_over_forwarder(
     });
 }
 
+// Java game over is an ordered sequence: final state updates and the
+// gameOver prompt must reach clients before EndGame resets the room.
+// This forwarder receives the sequence as one bundle so EndGame cannot
+// overtake the final engine messages on outbound_tx.
 fn spawn_java_game_over_forwarder(
     outbound_tx: tokio_mpsc::UnboundedSender<ClientMessage>,
     game_over_rx: std_mpsc::Receiver<java_backend::HostedGameOver>,
--- a/forge-engine/crates/self-hosted-node/src/main.rs
+++ b/forge-engine/crates/self-hosted-node/src/main.rs
@@ -838,8 +838,8 @@ fn maybe_start_hosted_engine(
             drop(guard);
 
             spawn_remote_prompt_forwarder(outbound_tx.clone(), remote_prompt_rx);
-            let (game_over_tx, game_over_rx) = std_mpsc::channel::<String>();
-            spawn_game_over_forwarder(outbound_tx.clone(), game_over_rx);
+            let (game_over_tx, game_over_rx) = std_mpsc::channel::<java_backend::HostedGameOver>();
+            spawn_java_game_over_forwarder(outbound_tx.clone(), game_over_rx);
             spawn_engine_thread(move || {
                 info!(
                     game_id,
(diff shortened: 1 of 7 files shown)


forge-engine/crates/self-hosted-node/src/main.rs
```diff
@@ -1092,6 +1092,55 @@ fn spawn_game_over_forwarder(
     });
 }
 
+fn spawn_java_game_over_forwarder(
```
anacletos: So this one I'm not sure about.
In theory it shouldn't be necessary to spawn a new thread to fix the game over.
Why did you go for an extra thread?

mermerico: Sorry, I let Codex run wild with this. I probably shouldn't have bitten off this PR because I don't have the full architecture in my head and all the tradeoffs involved. My understanding is that the underlying issue is a race between the client receiving a GameAborted signal and the GameOver prompt. I think the potential solutions are nuanced and I don't have enough understanding to make the tradeoffs. For instance the relay could be responsible for sending different kinds of game end messages depending on who won and that would drive the UI. Maybe I'll just file the issue and close this PR?

khaliostr:
Hey @mermerico , don't worry, you're closer than you think! I have looked into this before and I concluded that the Java game ends right around when the final state and the GameOver prompt go out, and since the GameOver prompt is just a marker (no data) after the recent protocol changes, I had opted to simply emit game end as a fallback (see #141).

Looking at your code though, what I understand now is that it's actually an ordering thing: the final state and the GameOver prompt were going out on `remote_prompt_tx` (handled by `spawn_remote_prompt_forwarder`), while the end signal went on a separate `game_over_tx` (handled by `spawn_game_over_forwarder`), so we have two channels and two threads both writing to `outbound_tx` with nothing guaranteeing the order between them, and `EndGame` can overtake the prompt, which is why the client shows it as a disconnect/abort. Bundling them into one `HostedGameOver` on a single channel so they go out in order is the right fix, and the client side guards (hiding the abort overlay once a gameOver prompt is pending) are a good extra safety net.

@anacletos to your point, it's not really an extra thread, it replaces `spawn_game_over_forwarder` on the Java path, the win is that the prompts now ride the *same* channel as `EndGame` instead of racing it on a second one. Two small things while we're here: `spawn_game_over_forwarder` is still used by the native Rust backend, so we now have two almost identical forwarders and it might be worth merging them into one. And the Rust path still has the old two channel pattern, so it carries the same latent race, it doesn't bite us today since hosted AI runs on Java-Forge but I would suggest we either fix it here or file a follow up so we don't forget.

So I would suggest to keep this open and merge once we're happy, it's a proper root cause fix and not a patch. @mermerico if you can write up the race in the description, even just a quick before/after on the ordering, that'd help us a lot!

mermerico:
Okay, I added a bit more documentation

khaliostr (commented):
Thanks @mermerico , looking good! Let's wait for @anacletos ' approval to merge

forge-engine/crates/self-hosted-node/src/main.rs
```diff
@@ -838,8 +838,8 @@ fn maybe_start_hosted_engine(
             drop(guard);
 
             spawn_remote_prompt_forwarder(outbound_tx.clone(), remote_prompt_rx);
-            let (game_over_tx, game_over_rx) = std_mpsc::channel::<String>();
-            spawn_game_over_forwarder(outbound_tx.clone(), game_over_rx);
+            let (game_over_tx, game_over_rx) = std_mpsc::channel::<java_backend::HostedGameOver>();
+            spawn_java_game_over_forwarder(outbound_tx.clone(), game_over_rx);
```
anacletos: What I would do @mermerico, is to merge these two 
`spawn_game_over_forwarder` and `spawn_java_game_over_forwarder` into one that works for both, so we don't add an edge case. Then this becomes a nice fix!

anacletos (dismissed):

forge-engine/crates/self-hosted-node/src/main.rs
mermerico: That makes sense, will do!

mermerico: Done! I'm guessing the Rust engine might have some final messages that it would want to send at game over, but I'm treating that as out of scope and leaving it to people who are working in that engine.

khaliostr (approved):
Top work! Ready to ship

PR #179, opened 2026-06-12
fix(hosted): preserve nonland mana source choices

mermerico:
## Summary

- forward hosted Java mana ability produced-mana metadata through the prompt bridge
- preserve chosen mana colors when activating nonland mana sources
- let the client present nonland mana sources through the same tap/action flow as lands

### Black Lotus Before / After
Pasted Graphic
Pasted Graphic 3

### Talisman of Creativity Before / After:
Pasted Graphic 2
Pasted Graphic 4

## Why

Hosted Java mana prompts were treating cards like Black Lotus as generic artifact actions instead of preserving the exact mana ability choice selected by the player. That meant the client could lose which color was chosen, or fail to surface nonland mana sources consistently during payment.

## Test plan

- `cargo check -p forge-agent-interface && cargo check -p self-hosted-node --features java-forge`
- `./node_modules/.bin/eslint src/views/Game.tsx src/components/game/manaUtils.ts && ./node_modules/.bin/tsc -p tsconfig.app.json --noEmit`
- `yarn lint:all` fails on existing Rust clippy issues unrelated to this change (`forge-card-script`, `forge-protocol`, `forge-foundation`)

## Build artifacts

- [ ] Build macOS .dmg
- [ ] Build Windows .exe

### DIFF
--- a/src/components/game/manaUtils.ts
+++ b/src/components/game/manaUtils.ts
@@ -9,9 +9,44 @@ export function extractManaLetters(desc: string | undefined): string[] {
 
 export const ANY_COLOR_LETTERS = ["W", "U", "B", "R", "G"];
 
-/**
- * Expand mana abilities by detecting color options in descriptions.
- */
+const MANA_TOKEN_TO_LETTER: Record<string, string> = {
+  WHITE: "W",
+  W: "W",
+  BLUE: "U",
+  U: "U",
+  BLACK: "B",
+  B: "B",
+  RED: "R",
+  R: "R",
+  GREEN: "G",
+  G: "G",
+  COLORLESS: "C",
+  C: "C",
+};
+
+function extractProducedManaTokens(cost: string | undefined): string[] {
+  if (!cost) return [];
+  return cost
+    .replace(/[{}]/g, " ")
+    .split(/[\s,/]+/)
+    .map((token) => token.trim().toUpperCase())
+    .filter(Boolean);
+}
+
+function extractProducedManaLetters(cost: string | undefined): string[] {
+  return extractProducedManaTokens(cost)
+    .map((token) => MANA_TOKEN_TO_LETTER[token])
+    .filter((letter) => letter != null);
+}
+
+function hasAnyColorText(text: string): boolean {
+  return (
+    text.includes("any color") ||
+    text.includes("any one color") ||
+    text.includes("mana of any color")
+  );
+}
+
 export const getExpandedManaAbilities = (
   cardId: string,
   options: ActivatableAbilityInfo[],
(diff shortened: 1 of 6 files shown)


src/components/game/manaUtils.ts
```diff
@@ -9,9 +9,44 @@ export function extractManaLetters(desc: string | undefined): string[] {
 
 export const ANY_COLOR_LETTERS = ["W", "U", "B", "R", "G"];
 
-/**
- * Expand mana abilities by detecting color options in descriptions.
- */
+const MANA_TOKEN_TO_LETTER: Record<string, string> = {
```
fedepoi, line 12: we wanna actually map those to the mana symbol svg , not just their string letter. the rust engine already does this, so prabably the fix here should follow what rust already does

mermerico: Is this working with the rust engine? I'm seeing the same thing there. 
image

I'll try to fix both.

khaliostr:
Hey @mermerico , this is a solid fix. Dropping the `types.includes("Land")` gate is exactly the right move, that's what was keeping artifacts like Black Lotus and the Talismans from ever surfacing their mana abilities, so nonland sources finally get the same tap flow as lands. The `setExpressChoice` plumbing to pin the chosen color is the correct Forge API too, so the core of this is good.

On @fedepoi's point about mapping to the mana symbol svg rather than the string letter, I think he's right and it's worth a follow up. The pips render fine for lands today because the description already enumerates the colors (e.g. "Add {R} or {G}" on a dual), so `getExpandedManaAbilities` splits it and `TextWithMana`/`ManaSymbols` draws them. The artifact case is where it gets shaky, the description is something like "three mana of one color" with no concrete `{X}`, so we end up re-deriving the colors from the `cost` string with the new `MANA_TOKEN_TO_LETTER` table. The rust engine already models this structurally (the `ProducedMana` enum, `Fixed`/`Any`/`Chosen`), it just doesn't expose it to the client yet, which is also why @mermerico you're seeing the same thing on the rust path, neither backend sends the structured produced mana down so the client is stuck parsing strings on both.

So what I would suggest is we land this for the working behavior since it does fix the artifacts, and as a follow up add a structured produced-colors field to `ActivatableAbilityInfo` (populated from `produced_ir` on the rust side and the same metadata you're already plumbing on the java side), then have the client render from that instead of the regex + letter table, our own `game/AGENTS.md` even says don't roll your own symbol parsing. That way both backends get fixed in one go and we can drop the string parsing entirely. Happy to help with that follow up!

mermerico:
Okay, I think what was happening with the icons was actually a local-only problem to do with loading the SVG icons from scryfall and CORS. I implemented a fallback, but we may decide not to carry the code bloat. I also took on the larger project of standardizing how mana abilities are encoded across Rust and Java. Now when a mana ability produces multiple colors of mana from a single choice they get grouped together:
Pasted Graphic 6
vs
Pasted Graphic 7

The code for that is here: https://github.com/witchesofthehill/manabrew/commit/a34b5ec81d6042d77636be761ac6b6b394d63287

Should I pull that into this PR or make a follow-up?

src/components/game/manaUtils.ts
khaliostr: Hey @fedepoi , I think this is actually fine as is. The letter here isn't rendered directly, the pips get resolved later down the line: the description ends up as "Add {W}" and `TextWithMana`/`ManaSymbols` turns that `{W}` into the Scryfall svg, so the string letter is just an intermediate step. mermerico also tracked the icons not showing down to a local-only Scryfall CORS issue rather than the encoding, so the mapping here is correct.

khaliostr (commented):
Looking good to me, the gate removal and `setExpressChoice` plumbing are solid and the behavior works. 

@mermerico for the bigger Rust/Java mana-ability standardization (commit a34b5ec), I'd open an issue describing it and then do it as a follow up PR rather than pulling it into this one, that keeps this PR focused and easy to review/revert and gives the standardization its own parity run. The Scryfall CORS fallback you can probably drop unless you've hit it outside local dev.

I'll defer to @fedepoi for the proper approval since it's his area.

fedepoi (approved):
ok approved lets open a follow up pr for commit commit https://github.com/witchesofthehill/manabrew/commit/a34b5ec81d6042d77636be761ac6b6b394d63287 as @khaliostr suggested

Issue #193, opened 2026-06-15
Sources that have abilities proving two mana of different colors show up as two abilities providing one mana each


mermerico:
Here Composite Golem has a mana ability that provides WUBRG, but it shows up as five separate abilities.

Composite Golem:

<img width="200" height="275" alt="Image" src="https://github.com/user-attachments/assets/66cc285b-ced0-4470-adf6-8282367ac666" />

Currently renders in the UI like this:

<img width="85" height="116" alt="Image" src="https://github.com/user-attachments/assets/d3d6b499-0159-495a-97e8-cd250b53c2f6" />

We'd like it to show up as one ability:

<img width="103" height="144" alt="Image" src="https://github.com/user-attachments/assets/0a0ab625-152f-4c70-b88e-5eb2f783cc81" />

khaliostr:
[[Composite Golem]]

khaliostr:
Fixed in #216

PR #194, opened 2026-06-15
fix(ui): fix mana source symbols for cards that produce multiple mana types

mermerico:
## Summary
- Addresses https://github.com/witchesofthehill/manabrew/issues/193
- Unifies the representation of mana abilities specified by the Forge engine and the Rust engine
- New mana ability parser drives the rendering of mana abilities on cards
- Includes a fallback for loading the mana icons which is useful in my local testing but we can remove it if others don't want it

## Why
The Forge and Rust backends represented mana abilities in different ways. The Forge interface put the produced mana in the "cost" field, while the Rust interface put it in the description. There was also a difference in whether it was specified in human readable "Add {G}" style or just "G". With this PR things are split up. The description is spelled out and human readable. There is a new field called ProducedMana that encodes the produced mana in forgescript style, making it unambiguous whether a card produces two mana of different types with one activation or a choice of one mana type or another. For some cards the type of mana they can produce is chosen at cast time. In these cases, Forge will resolve what type of mana the card can produce and pass it through to the clients (`ability.getManaPart().mana(ability)`).

## Test plan
- yarn lint:all (passes eslint, Prettier, and TypeScript; fails in cargo clippy -D warnings on unchanged current-main warnings outside this patch: forge-card-script unnecessary_lazy_evaluations, forge-foundation unused_mut/derivable_impls, forge-protocol large_enum_variant)

## Build artifacts


khaliostr (commented):
Hey @mermerico , clean PR and exactly the standardization we wanted. Verified the contract lines up both ways: Rust `as_script_text` gives `Combo W U` for choose-one vs plain `W U`/`W` for `Fixed` (`is_choice_like` agrees), and Java passes `getOrigProduced()` through with the same `Combo` prefix — so duals → two pips, signets → `Add {W}{U}`, choice rocks → split. Nice.

A few things:
- `Combo Any` ([[Cascading Cataracts]]) renders as one all-five pip while plain `Any` expands to five choices — fine as an illustration for "any combination" cards, but does the tap flow still let the player pick the combo, or lock them to WUBRG? (`Combo ColorIdentity` falls through to the description path — rare, fine.)
- `const MANA_SYMBOLS = ["W","U","B","R","G","C"]` in `produced_mana_text` is a third copy of the color set (`Color::short_name`/`ALL` in forge-foundation, plus the `mana_to_name` table right above it) and reinvents `chosen_colors_to_atoms` + `atom_short` that your `computer_util_mana.rs` copy already uses. Let's converge both impls on those (promote them to `pub`) and drop the array — output's right, just DRY.
- Minor: `ability_descriptions` is now a 4-tuple; a tiny named struct would read safer.
- Scryfall fallback: let's hand that piece to @fedepoi, he's familiar with it — maybe split it out. It's dev-web only and pips render fine live, so no rush.

Happy to approve once `Combo Any` is confirmed and the `MANA_SYMBOLS` duplication is cleaned up; we can sort the fallback with Federico separately.

mermerico:
You were right, `Combo Any` wasn't working correctly. Fixed that, addressed the other comments, and then had to do some rewiring after the merge, but I think it's all working now.

khaliostr (commented):
Re-reviewed @mermerico — earlier points all addressed (`MANA_SYMBOLS` reuses `chosen_colors_to_atoms`/`atom_short`, `NormalizedAction` is a named struct, `Combo Any` renders one all-five pip). Nice. Two things to confirm before approving:

**1. (please confirm) The `express_choice` path dropped the deliberate parity RNG-consume.** In `determine_mana_production_ir`, the forced-color branch used to call `choose_color` on a single-option list to mirror Java (*"Java calls chooseColor even when expressChoice is set… Consume the RNG pick for parity"*). The rewrite returns `Some(forced)` directly and drops that call. `express_choice` is only set from interactive taps (AI runs pass `None`), so diff traces likely won't catch it — but if Forge core still calls `chooseColor` under an express choice, this desyncs the RNG stream vs Java. Intentional? If so, worth a one-line note.

**2. `Combo Any` tap flow.** For Cascading Cataracts (`Combo Any`, `Amount$ 5`) the single pip is right, but can you confirm the tap surfaces the `SpecifyManaCombo` prompt so the player picks 5 freely (vs locking to one-of-each)? The plumbing is here; the client component rendering `specify_mana_combo` isn't in this diff.

**Nits**
- `amount <= 0` now early-returns before any agent call — same RNG-ordering class as #1, rare/probably benign, but worth a glance vs Java.
- `manaColorChoiceFromAction` returns `letters[0]` even for plain fixed lands, so basics route through `tapLand` + a redundant `express_choice` (harmless; returning `null` keeps them on the simpler path).
- `callback_fmt.rs` only emits the 3-arg form when `express_choice.is_some()` — keeps parity baselines stable. 👍

Sound and well-scoped — happy to approve once .1 is confirmed parity-safe and `Combo Any` is verified end-to-end.

Logistics: we're in a **code freeze today** for a sweeping change, so let's hold the merge and pick this up **tomorrow** once the tree settles — you may want to re-pull `main` first.

fedepoi (commented):
this pr wil surely conflict with my pr !, ive completely deleted and refactored the pixi canvas to make it better more performant.

It's ok I i think we can handle the conflicts #195

khaliostr:
@mermerico got conflicts! let me know if you can use some help to merge. Should be mostly naming

forge-engine/crates/forge-protocol/src/prompts/common.rs
```diff
@@ -21,6 +21,9 @@ pub struct ActivatableAbilityInfo {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[ts(optional)]
     pub cost: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub produced_mana: Option<String>,
```
khaliostr, line 26: Why do we need to add `produced_mana` to the protocol?

forge-engine/crates/forge-protocol/src/prompts/choose_action.rs
```diff
@@ -28,6 +28,9 @@ pub enum AvailableActionKind {
         #[serde(default, skip_serializing_if = "Option::is_none")]
         #[ts(optional)]
         produced_colors: Option<Vec<String>>,
```
khaliostr, line 30: Could we use produced_colors instead?

forge-engine/crates/forge-agent-interface/src/java_prompt_normalizer.rs
```diff
@@ -654,21 +669,29 @@ fn build_choose_action(
...
+                    .clone()
+                    .unwrap_or_else(|| action.label.clone());
+                AvailableActionKind::ActivateAbility {
+                    card_id: card_id.clone(),
+                    ability_index: action.index,
+                    description,
+                    cost: action.cost.clone(),
+                    is_mana_ability: true,
+                    produced_colors: action.produced_mana.as_deref().and_then(parse_mana_colors),
+                    produced_mana: action.produced_mana.clone(),
```
khaliostr, line 684: I think these should be collapsed into one field only

anacletos:
@mermerico I think the easiest way to fix the conflicts is to just rename the folders to the new names like in main.
After that there should be no conflicts at all

mermerico:
This PR used to be quite clean, but things have devolved a bit with the merges. I may end up re-implementing it.

fedepoi:
> This PR used to be quite clean, but things have devolved a bit with the merges. I may end up re-implementing it.

Yeah i think its the easiest approach here, sorry for the conflicts hell 🙈

mermerico:
Closing in favor of https://github.com/witchesofthehill/manabrew/pull/216

PR #212, opened 2026-06-16
chore(protocol): condense targeting into one prompt

anacletos:
## Summary
* Single `choose_board_targets` interface to interactively target on the board

## Why
Targeting board was split into 4 (or 5?) different prompts. Mostly duplicated, quite sloppy, with no clear interface

## Build artifacts


fedepoi (approved):
looking solid

PR #216, opened 2026-06-17
fix(ui) unify mana ability selection (v2)

mermerico:
## Summary
- Unifies mana-source selection so engines can identify a specific mana ability and optional expressed color through priority and combat-cost prompts.
- Adds produced-mana metadata to ability prompts and renders mana ability buttons from that metadata, including multi-symbol rows for abilities such as Cascading Cataracts.
- Reduces noisy mana-choice prompts in the Java interactive harness and adds a local Vite proxy for Scryfall mana symbols.

## Why
The previous mana ability PR had drifted after large main-branch refactors. This rebuild keeps the behavior focused on the current code shape: mana sources are specified consistently by ability index/color, users see the actual produced mana where available, and Forge-hosted prompts avoid unnecessary choose-mode and floating-mana interruptions.

## Test plan
- `yarn gen:protocol`
- `cargo check -p manabrew-agent-interface`
- `yarn tsc --noEmit`
- `yarn build:harness`
- `cargo fmt --all`
- `yarn test`
- `git diff --check`
- `yarn prettier --check vite.config.ts`
- `yarn lint:all` currently fails in Rust clippy on pre-existing workspace warnings outside this PR scope: `forge-card-script/src/lib.rs`, `forge-foundation/src/{phase.rs,card_split.rs}`, and large enum variant warnings in `manabrew-protocol/src/{prompts/mod.rs,protocol.rs}`.

## Build artifacts


manabrew-rs/crates/manabrew-protocol/src/prompts/choose_action.rs
```diff
@@ -24,7 +24,7 @@ pub enum AvailableActionKind {
         is_mana_ability: bool,
         #[serde(default, skip_serializing_if = "Option::is_none")]
         #[ts(optional)]
-        produced_colors: Option<Vec<String>>,
+        produced_mana: Option<String>,
```
anacletos, line 27: is this meant to be the same as `common.rs::ActivatableAbilityInfo`
That one has both:

```
pub produced_mana: Option<String>,
    #[ts(optional)]
    pub color: Option<String>,
```

forge-harness/src/main/java/forge/harness/host/ManaBrewInteractiveController.java
```diff
@@ -2003,13 +2013,9 @@ public Mana chooseManaFromPool(final List<Mana> manaChoices) {
...
-        final List<String> labels = new ArrayList<>();
-        for (final Mana mana : manaChoices) {
-            labels.add(mana == null ? "Mana" : mana.toString());
-        }
-        final List<Integer> chosen = session.awaitModeChoice(me(), labels, 1, 1, null);
-        final int idx = chosen.isEmpty() ? 0 : chosen.get(0);
-        return idx >= 0 && idx < manaChoices.size() ? manaChoices.get(idx) : manaChoices.get(0);
+        // Probes and auto-pay must not prompt, and exact floating-mana selection is too noisy until it has an opt-in UI.
+        // Greedy selection of the first option is a reasonable approximation in the meantime.
+        return manaChoices.get(0);
```
anacletos, line 2018: Wouldn't this break autopay?

mermerico: The behavior without this was that if you had floating mana and you clicked auto-pay it would always repeatedly prompt you to ask if you want to use the floating mana towards the card you are casting.

manabrew-rs/crates/manabrew-protocol/src/prompts/choose_action.rs
mermerico: `color` in `common.rs::ActivatableAbilityInfo` is used to disambiguate between different ability options (for instance choosing "green" on black lotus). It's not necessary here because each ability option has a separate ability id in choose_action. In the future we may want to replace `color` with a more generic `choiceId`

forge-harness/src/main/java/forge/harness/host/ManaBrewInteractiveController.java
anacletos: interesting, I've never seen that happen. Good to know.

fedepoi (approved):

anacletos (approved):

PR #245, opened 2026-06-20
refactor(harness): move protocol boundary to remove duplication

anacletos:
## Summary
The `manabrew-agent-interface` was acting as a boundary translator between a *made up* forge IPC contract and `manabrew-protocol`. I don't see why this is useful. The proposal is to make `manabrew-protocol` the de-facto standard between harness and agent interface as well. This removes the need of an intermediate representation of state.

## Why
`JavaRaw` and `java_prompt_normalizer` were confusing abstractions that are not needed.

## Extra Bugfixes
* `simple_ai` would get stuck on many loops
* Many prompts were relying on snapshots being inserted into the prompt against protocol specification

## Build artifacts


khaliostr (dismissed):
nice work

khaliostr (approved):

PR #246, opened 2026-06-20
fix(ui): pixi lifecycle/leak cleanup + dependency security bumps

fedepoi:
### Pixi lifecycle / leak / correctness pass
- **Teardown leaks:** `safeDestroy` now cascades into children (`{children:true}`) so Graphics geometry is freed on mid-game teardown instead of leaking until the Pixi app is disposed (safe because `installPixiPatches` neutralises the `returnTexture` crash app-wide). `PlayerSquarePanel`/`PhaseStripLayer`/`BoardRegion` destroy their full container tree; `CardSprite` frees its `ColorMatrixFilter` + cloned frame styles; `PointerLayer` frees its owned textures. `BoardRegion.destroy` was the worst offender — `configure()` destroys regions when a player leaves mid-game, so every departed opponent's sprites leaked for the session.
- **Shared-TextStyle bug:** every `PlayerSquarePanel` stat cell shared one module `statStyle`, then per-cell code wrote `fontSize`/`fill` onto it — last write won, collapsing life's font-16 to 13 and bleeding highlight fills. Each cell now clones its own style. Dropped `PhaseStripLayer`'s duplicate `combatActiveStyle`.
- **Async raster guards:** skip pending sprites already destroyed when a raster/texture load resolves (panelIcons + mana load); `CardSprite.loadImage` gets a generation token so a slow earlier load can't overwrite the current preview face.
- **Gradient leak:** moving combat/cast arrows allocated two `FillGradient`s per frame and leaked the old ones; now freed before replacement and on teardown.
- **Hygiene:** `clearManaSymbolCache` destroys textures instead of dropping refs; removed a dead grid guard and a no-op arithmetic expression.

### Security advisories (dependabot)
- **npm:** react-router-dom 7.13 → 7.18.0 (7 advisories incl. high), vitest 3.2.4 → 3.2.6 (critical), vite 7.3.3 → 7.3.5 (high); pinned js-yaml ^4.2.0 (medium) and @babel/core ^7.29.6 (low) via `resolutions` (transitive dev-only).
- **Rust:** openssl 0.10.79 → 0.10.80 (medium); deleted the vestigial `src-tauri/Cargo.lock` (src-tauri is a member of the root cargo workspace, so its lock is unused and only produced false alerts for a stale tauri 2.10.2 / rand 0.8.5 — the root lock already carries the patched tauri 2.11.1).

## Why

A read-through of `src/pixi` surfaced GPU-resource leaks on mid-game teardown (regions, panels, phase strip, card filters, arrow gradients) and one visible correctness bug (stat fonts/colours bleeding through a shared `TextStyle`). Pixi scope was deliberately bounded — findings that turned out to be bounded caches, transient drag-only redraws, or controlled-input boundaries were left alone rather than fixed speculatively (see per-commit messages). The dependency bumps clear the dependabot alerts surfaced on push.

## Not fixed (out of scope / would be breaking)
- **esbuild** low-severity dev-server advisory: vite 7.3.5 pins esbuild `^0.27.0`, so forcing 0.28.1 would break vite's bundling.
- **glib 0.18** and transitive **rand 0.7.3**: pinned by the webkit2gtk/tauri stack; require a breaking ecosystem upgrade.
- **website** (`astro`, separate `package-lock.json`) and the `tree-sitter-forge-card-script` subproject: separate projects/deploys — better handled in their own PR.

## Test plan
- `yarn lint` (eslint + prettier + tsc), `yarn test` (9 passed), and `yarn vite:build` all pass with the bumped toolchain.
- **Manual smoke test still recommended** for the pixi teardown change (the one with runtime risk): play an engine game, churn cards, run combat, have a player leave mid-game, switch themes, fast-toggle a card's preview face. Confirm no crash, watch the console for `returnTexture guard swallowed` / destroy throws (the patch should absorb them), and verify stats render at correct sizes.

## Build artifacts


khaliostr (dismissed):

anacletos (commented):
Some too many comments but looks good. Why is the lock gone?

fedepoi:
trimmed the comments this PR added, and the deleted `src-tauri/Cargo.lock` was a stale, unused duplicate (builds resolve against the root workspace lock), so removing it just clears false dependabot alerts.

anacletos (approved):

PR #248, opened 2026-06-21
docs: Manabrew protocol lands in the docs!

anacletos:
## Summary
With the real first `v1.0.0` of the`manbrew-protocol` approaching, the docs are starting to form.

### As part of this PR
* `choose_type` was collapsed into `choose_from_selection`
* `choose_from_selection` UI was enhanced with searchbar if many options
* `choose_divider` was a struggler, only mentioned in the `forge-harness` and never honored anywhere

### Enforcing the protocol
As part of the PR, the `forge-harness` now gets auto-generated protocol Java type files, as well as constructors to migrate away from`JsonObject`. Finally, `forge-harness` will fail compilation on protocol changes, if not addressed.

## Build artifacts


khaliostr (dismissed):
let's see it live!

PR #251, opened 2026-06-22
feat(ui): show stack target indicators

mermerico:
## Summary

- Serialize stack-target relationships from Forge snapshots, including battlefield targets, player targets, nested sub-ability target nodes, and stack spell targets.
- Render persistent target indicators for stack objects that target battlefield cards or players, including de-overlapped glyphs for repeated non-attach effects and arrow handling for attach-style effects.
- Render stack-to-stack target arrows in the board stack, target-selection stack view, and stack modal with shared curve geometry.
- Note: This code is not as well reviewed by me as I would like. I'm still getting up to speed on the UI code.

## Why

Stack objects could carry target data, but the UI did not consistently preserve or display that relationship after target selection. Battlefield/player targets needed persistent effect glyphs that would not overlap when multiple stack objects targeted the same card, and counterspell-style interactions needed stack-to-stack target relationships serialized and rendered after the prompt completed.

## Test plan

- [x] Merged latest `origin/main` into `feat/stack-target-indicators`.
- [x] `git diff --check --cached`.
- [x] `yarn lint:all` reached and passed ESLint, Prettier check, protocol generation, and TypeScript before failing in Rust clippy on pre-existing latest-`main` warnings in `forge-carddb` and `forge-limited`.
- [x] Commit hook ran `eslint --fix` and `prettier --write` on staged TS/TSX files successfully.
- [x] Captured Target Lab screenshots from `/dev/target-lab` with the target-lab stash applied locally.
- [ ] Full `yarn lint:all` is blocked by the existing Rust clippy baseline.

## Demo

Battlefield/player target indicators plus completed board stack arrows:
(Note: this is in a demo environment, the card doesn't actually do all these things)

Counter target selection modal:

View-only stack modal after targets are chosen:

I didn't commit the demo environment, but I think these are generally useful. We might want to add support for hidden demos that can be used to fine tune UI stuff.

## Build artifacts


fedepoi (commented):
this is sick !, can we add this as a controlable preference in the settings? there are 2 reason i can think of 
- some people might find too many arrows annoying
- all those arrow layers make the game laggy after a while , this is something im trying to takle in #246 but still all these pixi render are slowing down the app a bit

anacletos (commented):
I was thinking maybe this is a good chance to define everything needed to render targeting in `manabrew-protocol::common::TargetRef`

currently it only shows:
```rs
pub enum TargetRef {
    Player { id: String },
    Card { id: String },
    Spell { id: String },
}
```

But I do think it's worth standardizing all the info required to define a target.
For example the hostile, target type (buff, sac, tap, etc..) and everything.

Then we could further remove surface for bugs as we would be able to show targets in multiple occasions. Something that has been a problem historically is a prompt that references an effect, with no target information.

mermerico:
> this is sick !, can we add this as a controlable preference in the settings? there are 2 reason i can think of
> 
>     * some people might find too many arrows annoying
> 
>     * all those arrow layers make the game laggy after a while , this is something im trying to takle in #246 but still all these pixi render are slowing down the app a bit

Thanks! I've included a setting for that:
image

mermerico:
> I was thinking maybe this is a good chance to define everything needed to render targeting in `manabrew-protocol::common::TargetRef`
> 
> currently it only shows:
> 
> ```rust
> pub enum TargetRef {
>     Player { id: String },
>     Card { id: String },
>     Spell { id: String },
> }
> ```
> 
> But I do think it's worth standardizing all the info required to define a target. For example the hostile, target type (buff, sac, tap, etc..) and everything.
> 
> Then we could further remove surface for bugs as we would be able to show targets in multiple occasions. Something that has been a problem historically is a prompt that references an effect, with no target information.

Gave it a shot

PR #274, opened 2026-06-24
ci: PR comment commands to trigger builds and reruns

fedepoi:
## Summary

- Add `.github/workflows/pr-commands.yml`: an `issue_comment` handler that lets a trusted commenter trigger installer builds and workflow re-runs from a PR comment.
- Document the commands in the README under Contributing → "On-demand CI commands".

### Trigger sentences

Post one of these as a comment on a PR:

| Comment | Effect |
| --- | --- |
| `/build mac` _(or `/build macos`)_ | Build the macOS `.dmg` from the PR branch |
| `/build windows` _(or `/build win`)_ | Build the Windows `.exe` / `.msi` from the PR branch |
| `/build all` _(or `/build both`)_ | Build both installers |
| `/rerun <workflow>` | Re-run the latest run of a workflow on the PR branch — e.g. `/rerun regression`, `/rerun build-checks`, `/rerun deploy` |

`/build *` dispatches `release-artifacts.yml` on the PR branch with the matching `build_macos` / `build_windows` inputs (reusing its existing `workflow_dispatch` gate). `/rerun <name>` re-runs the most recent run of `<name>.yml` for the PR branch via `gh run rerun`, so it works for push/PR-triggered workflows too — not just dispatch-enabled ones.

## Why

Today installer builds only happen on merge to `main` (via the PR checkboxes), on `v*` tags, or by hand in the Actions UI. There is no way to produce a `.dmg` / `.exe` for an **open** PR, or to re-kick a flaky check, without leaving the review. These comment commands close that gap.

## How it stays safe on a public repo

- **Write-access gate:** the job only runs when `comment.author_association` is `OWNER` / `MEMBER` / `COLLABORATOR`. That field is set server-side by GitHub and cannot be spoofed, so a drive-by commenter gets nothing (default-deny).
- **Runs from `main`:** `issue_comment` workflows always execute the default-branch copy of the YAML, so a PR cannot weaken the gate and have the edit take effect before it's merged.
- **Forks rejected:** the handler exits on `isCrossRepository` — privileged commands never run against an untrusted head.
- **No script injection:** the comment body is read only through an env var, never interpolated into a shell command.
- **Token scoping:** dispatch uses `WORKFLOW_DISPATCH_PAT` if present (a fine-grained `actions: write` PAT makes `workflow_dispatch` reliably spawn a run), falling back to `GITHUB_TOKEN`.

## Test plan

- `python`/prettier: YAML + README pass `prettier --check`.
- Manual verification after merge (the handler must be on `main` to take effect, since `issue_comment` runs the default-branch copy): comment `/build mac` on a same-repo PR → 👀 reaction, dispatched-run reply; comment from a no-access account → ignored; comment on a fork PR → rejected reply.

> [!NOTE]
> Optional follow-up: add a fine-grained `WORKFLOW_DISPATCH_PAT` repo secret (`actions: write` + `contents: read`, this repo only) so `/build *` dispatches are 100% reliable. Without it the workflow falls back to `GITHUB_TOKEN`.

## Build artifacts

### DIFF
--- a/.github/workflows/pr-commands.yml
+++ b/.github/workflows/pr-commands.yml
@@ -24,28 +24,25 @@ permissions:
   issues: write
   contents: read
 
+concurrency:
+  group: pr-commands-${{ github.event.issue.number }}
+  cancel-in-progress: false
+
 jobs:
   command:
     if: >-
       github.event.issue.pull_request &&
-      startsWith(github.event.comment.body, '/') &&
+      (startsWith(github.event.comment.body, '/build') || startsWith(github.event.comment.body, '/rerun')) &&
       contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
     runs-on: ubuntu-latest
     steps:
-      - name: Acknowledge the comment
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          gh api -X POST \
-            "repos/${{ github.repository }}/issues/comments/${{ github.event.comment.id }}/reactions" \
-            -f content=eyes >/dev/null
-
       - name: Parse command
         id: parse
         env:
           COMMENT_BODY: ${{ github.event.comment.body }}
         run: |
-          cmd=$(printf '%s' "$COMMENT_BODY" | head -n1 | tr '[:upper:]' '[:lower:]' | xargs)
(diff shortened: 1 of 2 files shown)


fedepoi:
build mac

anacletos (dismissed):

JacopoMadaluni (approved):

PR #293, opened 2026-06-29
feat(ui): default tauri play-vs-ai to the forge engine

khaliostr:
## Summary

- On the Tauri (graalvm) build, **Play vs AI now defaults to the bundled Forge engine** instead of hardcoding Manabrew and skipping the engine choice.
- Adds a Tauri Forge-vs-AI flow: host a local Forge room (`start_forge_host`) → spawn the bot → join as a client. Extracted the shared join/deck/spawn/start logic out of the web `startHostedAiGame` path into `joinHostedRoomAndPlay`.
- **The relay is now embedded too** (option B from the discussion below): `start_local_relay` runs `manabrew-server` in-process on `127.0.0.1:<ephemeral>` with a random per-run key, and both the webview relay client and the self-hosted-node host connect to it. Forge Play-vs-AI needs no external relay and works fully offline. `manabrew-server` gains a `serve()` entry point that takes a pre-bound listener and a caller-owned shutdown handle; the standalone binary path is unchanged.
- Fallback chain: loopback relay unavailable → baked-in external relay (`VITE_RELAY_*`) → **in-process Manabrew engine**, so a game always launches.
- Web is unchanged: still shows `EngineChoiceModal` and uses a pooled hosted room.

## Why

Forge has no in-process runtime (`runtimeRegistry` has `forge: null`); it only runs as a relay-hosted room. The Tauri path hardcoded `"Manabrew"` (`Play.tsx`) and the engine modal was web-only, so the bundled GraalVM Forge engine was never reachable from Play vs AI on desktop. And since the engine host already runs inside the Tauri process, routing its messages through an external relay added a WAN round-trip and an availability dependency for a single-machine game — the loopback relay removes both while reusing the whole relay protocol untouched.

## Test plan

- `yarn tsc` — clean.
- `eslint` + `prettier --check` on the changed files — clean.
- `cargo check -p manabrew` with and without `forge-room`, `cargo check -p manabrew-server`, `cargo fmt --check` — clean.
- Booted the standalone relay binary (which now runs through the new `serve()` path) and verified the health endpoint responds.
- ⚠️ Not yet exercised end-to-end on a real GraalVM Tauri build — the host-and-join handshake against the loopback relay and the fallback chain still need a manual desktop run (now possible fully offline).

## Build artifacts


khaliostr:
## Follow-up: do we need the relay round-trip for Forge Play-vs-AI?

Worth recording the architecture here, since this PR routes desktop Forge Play-vs-AI through the relay even though everything runs on one machine.

**The engine already runs in-process.** `start_forge_host` spawns `self_hosted_node::host_room` as a task inside the Tauri process (either `graal-forge` — a native-image embedded via C FFI — or `java-forge`, a child JVM). So "use Forge in-process" wouldn't move the engine anywhere. What the relay actually costs us is the **message detour**: `UI → WebSocket → external relay → back → self-hosted-node (same process) → engine`. The relay is currently mandatory because `host_one_room` must `RelayClient::connect(...)` and the UI only receives Forge events via relay events (`game:remote_state` / `game:remote_prompt`).

### Options to drop the external-relay dependency

**A. Keep the relay (this PR).** Zero new transport code — reuses all of the room / ready / `spawnBot` / `GameStarted` / `game:remote_*` machinery. Downside: Forge Play-vs-AI depends on an external relay being reachable; when it isn't, it silently falls back to Manabrew. Fine while a relay is always provisioned and players are online.

**B. Embedded loopback relay — preferred.** `manabrew-server` (the relay) is an embeddable lib (`pub async fn run_server(state, addr, health_addr)`), so Tauri can spawn it bound to `127.0.0.1:<ephemeral>` and point both the self-hosted-node host and the webview relay client at `ws://127.0.0.1:<port>`. **No changes to the protocol, the UI relay client, or self-hosted-node** — they already speak relay; we just hand them a local one. Removes the external dependency, removes WAN latency, makes Forge Play-vs-AI work fully offline, and turns the Manabrew fallback into a genuine fallback instead of a silent default. Estimated ~100–200 LOC (a Tauri command to start/stop the in-process server + pick a free port, feeding host/port into the existing `connect` / `start_forge_host` calls).

**C. Direct in-process transport (new `forge` GameRuntime + loopback IPC channel).** Replace the WebSocket with a Rust↔webview channel and a dedicated forge runtime. ~500–800 LOC, touches UI transport + self-hosted-node + re-implements the room/handshake locally. Lowest latency, but **multiplayer still needs the relay**, so we'd maintain two transports.

### Recommendation

Ship this PR as-is for now, and if dropping the external-relay dependency matters, do **B**, not C. The engine is already in-process and multiplayer keeps the relay regardless, so C is extra complexity for little gain — it hand-rolls a transport and handshake the relay already provides. **B** gets ~95% of the benefit (offline, no external SPOF, low latency) by *reusing* the relay protocol locally, and it strictly improves this PR (its Manabrew fallback stays as the safety net) rather than competing with it.

khaliostr:
Folded **option B** from the comment above into this PR (608f408): `start_local_relay` embeds `manabrew-server` on `127.0.0.1:<ephemeral>` with a per-run key, and the Play-vs-AI flow connects both the UI client and the Forge host to it. The external relay drops from a hard dependency to the first fallback, and the Manabrew fallback stays as the safety net. ~60 LOC plus a `serve(listener, shutdown)` split in `manabrew-server` so an embedder can own the port and the shutdown signal.

anacletos (approved):

PR #341, opened 2026-07-05
fix: poor resume logic on bots and lost/conceded players

anacletos:
## Summary
* Fix bots not reconnecting on relay deploy
* Fix conceded / lost player breaking multiplayer phase handling


fedepoi (approved):

anacletos:
Included in #343

PR #346, opened 2026-07-05

Chiplis:
## Summary
- Adds Ironsmith as a Manabrew trusted engine/runtime option and loads the local sibling `ironsmith-wasm` package.
- Keeps Manabrew's public wire protocol almost unchanged: the only protocol enum addition is `EngineKind::Ironsmith`; Ironsmith-specific private traffic rides over the existing `roomRelay` envelope.
- Delegates the Manabrew compatibility layer to Ironsmith WASM instead of carrying a TypeScript adapter in this repo. Manabrew now calls `validateManabrewMatchConfig`, `startManabrewMatch`, `manabrewView`, `manabrewPublicState`, and `manabrewCommandFromPromptOutput`.
- Removes the Manabrew-side Ironsmith DTO/prompt/action/deck-source mapper, its tests, and the temporary Ironsmith smoke preset deck.
- Publishes one room-wide public `state` from Ironsmith's public Manabrew view, then sends encrypted per-seat Ironsmith state/prompt overlays and remote responses through an Ironsmith room-relay subprotocol.
- Wires the Ironsmith user-facing flow: the lobby can create and identify Trusted Ironsmith match rooms, Play can choose the `ironsmith` runtime for web starts, multiplayer startup carries the selected decks/seats into that runtime, and Ironsmith match validation failures are surfaced before launch.
- Keeps local bot auto-responses for Trusted-mode smoke tests and skips hidden-card placeholder prefetches.
- Makes protocol generation idempotent by de-duping duplicate top-level TypeScript exports emitted by overlapping `ts-rs` export roots.

## Local Ironsmith testing
1. Keep the repositories checked out as siblings so Manabrew can find the local WASM package:
   ```text
   ../ironsmith
   ../manabrew
   ```
2. Build the Ironsmith WASM package from the Ironsmith repo:
   ```bash
   cd ../ironsmith
   ./rebuild-wasm.sh --skip-scryfall-sync
   ```
   If the Ironsmith registry DB or card assets need refreshing, omit `--skip-scryfall-sync`. The Manabrew side expects `../ironsmith/pkg` to contain `ironsmith.js`, `ironsmith_bg.wasm`, and the generated `.d.ts` files.
3. Install/sync Manabrew dependencies from this repo:
   ```bash
   yarn install
   yarn ensure:wasm
   yarn sync:ironsmith
   ```
   `yarn sync:ironsmith` copies `../ironsmith/pkg` into `node_modules/ironsmith-wasm` and clears Vite's cached optimized dependency.
4. Start the local relay and web app in separate terminals:
   ```bash
   yarn dev:relay
   yarn dev:web
   ```
5. Open `http://localhost:1420`, connect to the local relay (`localhost:9443`, default server key `forge`), create a Match room with the `Ironsmith trusted` engine, choose any deck that Ironsmith can validate, add a bot with the same deck, and start the game.

Notes:
- This is Trusted mode only. There is no audit/proof/lockstep consensus path in this PR.
- The host tab owns the authoritative Ironsmith `WasmGame`; host tab loss still ends the v1 match.
- If Ironsmith WASM is rebuilt while Vite is running, rerun `yarn sync:ironsmith` and restart `yarn dev:web`.
- A debug `wasm-pack --dev` build can compile the Rust crate but fail during wasm-bindgen on the unoptimized module with `too many locals`; use the normal rebuild flow above for a usable local `pkg`.

## Validation
- Manabrew: `yarn tsc -p tsconfig.app.json --noEmit`
- Manabrew: `yarn test`
- Ironsmith: `cargo fmt --all -- --check`
- Ironsmith: `CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 wasm-pack build crates/ironsmith-wasm --target web --dev --out-dir ../../pkg --out-name ironsmith --no-default-features --features wasm-lean` compiled the Rust crate, then wasm-bindgen failed on the debug module with `too many locals`.

The protocol-generation step still emits the existing `ts-rs failed to parse serde attribute` warnings.


khaliostr:
hey @Chiplis, did a first pass

first off, really good stuff. first engine plugged end to end, hidden info handled properly, and the concede smoke tests are a genuinely useful contribution beyond ironsmith itself

now, i'd splt the PR into four different concerns.

**1) the protocol changes**
ChooseActionOutput::Concede removed, conceded_player_ids -> per player PlayerStatus, the new DirectiveInput / ClientToServerMessage channel, the Responder trait change. these break the engine-facing wire.

we have #351  to sort out v1 of the protocol,i'd say let's discuss and track this there

btw, i think you're right about directives. at launch we had players stuck unable to concede because concede lives in-band and the prompt loop wedged. it'll very likely get adopted. 

**2) the concede / disconnect rework**
per seat grace instead of the global token, seat concedes instead of the game dying, concede without priority, the 2p and 3p concede smokes. this is gold and i'd merge right away once split out.

just, what happens when conceding during a mulligan?

**3) deck validation blocking -> warning**
this changes behaviour for every engine, not just ironsmith rooms. the panel text says decks can be saved playable "for engines that can load them", but i don't see what guards a manabrew engine game start against an unsupported deck. what stops that today?

**4) the runtime itself**
- the e2e relay is nice. but three things:
  a) there's no identity binding, so it's encryption without authentication. fine for trusted mode, write it down in a comment and the doc, someone will read this code and assume more than it gives
  b) rememberPeerKey (:496-511) seems to silently *replac* a seat's key when a new hello arrives with a different fingerprint. looks like anyone who can inject a hello mid game takes over a seat's private channel undetected. i'd change it so once a key is pinned for a seat in a game, refuse the swap or warn loudly
  c) crypto.subtle is undefined off https and the guard just returns null (:485), so the failure mode is eternal "relay not ready". one explicit error would save a self-hoster on plain http a bad evening
- **botPlayerSlots keys off the username containing "-bot-"** (ironsmithRuntime.ts:591). i see the lobby names its bots that way (Lobby.tsx:377) so the check matches your own convention, fair. but a human who registers with -bot- in their name joins an ironsmith room and the host silently auto-answers their prompts. porbably could use an an explicit bot flag in the room config
- pendingBinding is one slot (ironsmithRuntime.ts:106) and publish() overwrites it per prompt in the loop (:322). two players with open prompts and applyResponse silently drops one (:260-264, wrong player -> republish -> return). if ironsmith is one prompt at a time by design, assert it. if not, map keyed by player
- any seat can claim isHost in its hello (:96) and hostPlayerSlot just takes it (~:411). am i reading it wrong?
- the ts-rs export dedupe in gen_protocol is a good fix to our own tooling, thanks for that, it stays regardless of the split

please note that i've had claude help me with the review as it's a fairly big pr so if anything is off do tell!

khaliostr:
Hey @Chiplis , top work! 

Small nit is if we could have ironsmisth as a git submodule rather than a peer checkout. But that can come in later after we test his PR

LGTM. Approved and let's wait for @fedepoi and @anacletos

khaliostr (approved):

khaliostr:
got green light from the team!

PR #350, opened 2026-07-06
feat: production observability (metrics, logs, analytics, game capture)

khaliostr:
## Summary

- Production observability in one PR, four staged commits (one conventional commit per part):
  1. **Relay**: Prometheus `/metrics` on the health listener, env-gated analytics JSONL events, env-gated per-game zstd stream capture, winner/game-over peek in the replay cache, rejection classification incl. `outdated_wire`.
  2. **Compose stack**: prometheus + pushgateway + grafana + loki + alloy behind an `observability` profile, Caddy routes (`grafana.` / `push.` / `loki.manabrew.app`), deploy.sh classifier.
  3. **Node metrics**: env-gated pushgateway exporter + engine-health instruments in `self-hosted-node`.
  4. **Ingester + dashboards + alerts**: JSONL→SQLite sidecar, live-ops + product dashboards, provisioned alert rules.

## Why

Every launch-weekend incident (wire skew, engine crashes, room collisions, the silently-dead node) was diagnosed by grepping container logs after players reported symptoms. This makes the same signals live (metrics + logs + alerts), the analytics trail durable (relay-emitted events → SQLite, covering **all** games incl. human-vs-human rooms the old node-side spec missed), and records complete per-game streams for future replay tooling.

Revised from the closed #333 spec: the relay is the sole analytics source (it sees decks/players/lifecycle typed, and peeks `gameOver`/`winnerId` in the state blobs it already routes), and the OTel collector is replaced by the `metrics` crate + pushgateway (outbound HTTPS push solves the remote-node NAT problem with ~5 fewer crates and ~130 MB less RSS).

Everything is env/profile-gated: merged but unconfigured = zero behavior change.

## Test plan

- `cargo check` / clippy clean on `manabrew-server` (workspace clippy has pre-existing toolchain-drift failures in `forge-foundation`/`manabrew-protocol`, untouched here).
- Local relay + scripted websocket clients (2 players, full lobby→deck→start→prompts→game-over→end cycle): events JSONL shows `seat_joined`/`deck_selected` (aggregated card counts)/`game_started`/`game_ended{reason:game_over, winner, conceded}`; per-game `.jsonl.zst` capture decompresses to header + all envelopes + footer; `/metrics` counters increment (`games_started`, `games_ended`, `resyncs`, `rejections{outdated_wire}` via a legacy-shaped deck fixture).
- Kill-client-mid-game → `game_ended{reason:reconnect_timeout}` after the room's timeout.
- Relay without any new env: `/health` unchanged, no files written, behavior identical.
- Compose: `docker compose config` validates with and without `COMPOSE_PROFILES=observability`; `caddy validate` passes both with the auth secret set and unset (locked-placeholder fallback), so a missing secret can never crash-loop Caddy; `bash -n deploy.sh`.
- Node: builds with and without `java-forge`; run against the local relay + a scripted pushgateway sink → 15s PUTs arrive with `manabrew_node_rooms_hosted{pool="solo"} 1` + `node_build_info`; without the push env the exporter is never installed (found and fixed a rustls CryptoProvider panic in the push client along the way).
- Ingester: ran twice over the smoke run's real JSONL → identical row counts (idempotent); games/players/decks/cards tables verified incl. winner attribution and the reconnect-timeout game; all dashboard SQL queries executed against the resulting `events.db`.
- All provisioning YAML parsed; dashboard JSON validated.

## Activation checklist (post-merge, on the box)

0. **Before merge**: `GRAFANA_ADMIN_PASSWORD` in `ops/production.secrets` — compose now hard-requires it (`:?`), so a deploy without it fails loudly instead of ever exposing a default-password Grafana.
1. DNS A records: `grafana.manabrew.app`, `push.manabrew.app`, `loki.manabrew.app`.
2. `ops/production.secrets`: `PUSHGATEWAY_PASSWORD_HASH` (single-quoted — bcrypt contains `$`; generate with `caddy hash-password`), `DISCORD_WEBHOOK_URL`.
3. `COMPOSE_PROFILES=observability` in the server env, then `docker compose --profile observability up -d`.
4. Remote node operator: set `SELF_HOSTED_NODE_METRICS_PUSH_URL=https://push.manabrew.app/metrics/job/self-hosted-node/instance/<name>` + `_USERNAME`/`_PASSWORD`.

Until these are done the merge is inert: relay features are env-gated, the stack is profile-gated, and the auth endpoints fall back to a locked hash.

## Privacy

This PR makes the server retain gameplay data it previously only routed, so being explicit about scope:

- **Metrics (Prometheus)**: aggregates only — counts, rates, durations. Labels are deliberately low-cardinality: no usernames, no room ids, no IPs.
- **Analytics events (JSONL → SQLite)**: in-app usernames, deck lists (card names/sets), commanders, formats, game outcomes/durations. Same scope the launch-weekend archive already collected. No IPs are captured anywhere in the pipeline, and there are no accounts/emails to leak. Stored on our box, admin-only, survives indefinitely (it's small).
- **Logs (Loki)**: the same container logs we already produce (which do contain usernames in relay log lines) — Loki makes them queryable and puts a 30-day retention on them instead of unbounded docker json-files.
- **Game-stream captures**: the sensitive one. Per-game compressed streams include **player-targeted prompts, i.e. hidden information (hands, searches) that no single player sees during the game**. This is strictly more than a spectator view. Mitigations: env-gated (off unless `MANABREW_GAME_CAPTURE_DIR` is set), admin-only access on the box, size-capped retention with oldest-first deletion (`MANABREW_GAME_CAPTURE_MAX_GB`, default 20). Open question (tracked outside the repo): redacting targeted prompts would make captures safe for player-visible replays.
- **No client-side telemetry**: nothing runs on or phones home from player machines. All collection is server-side over traffic the relay already relays.

Worth considering before enabling in production: a line in the site's terms/privacy text disclosing that online games (including decklists and results) are recorded for operations and product analytics.


khaliostr:
Review findings addressed in 2c404ab:

- **Grafana default password**: `GF_SECURITY_ADMIN_PASSWORD` is now `${GRAFANA_ADMIN_PASSWORD:?}` like the server key — a deploy without the secret fails loudly at compose interpolation; there is no window where admin/admin can be live. Activation checklist updated (secret is now a pre-merge step).
- **aws-lc-rs**: gone from `Cargo.lock` entirely. The exporter's `push-gateway-no-tls-provider` feature skips hyper-rustls' default provider; the node already installs `rustls::crypto::ring` at metrics init. Re-verified the push path end-to-end after the switch. No cmake in any build.
- **Idle captures**: the capture thread now uses `recv_timeout` with a 60s housekeeping tick — `close_idle` runs every minute regardless of traffic (previously it only ran when a message happened to arrive after a date change), retention still runs on date rollover.
- **Empty date dirs**: retention now attempts `remove_dir` on each date subdir after file deletion (no-op unless empty).
- **GDPR scrub note**: docs/OBSERVABILITY.md privacy section now documents exactly where a username lives and how to scrub it on request (SQLite DELETEs + JSONL grep + capture deletion/age-out).

Also swept comments down to the ≤2-line rule across compose/Caddyfile/deploy.sh/prometheus.yml and the ingester docstring.

fedepoi (dismissed):

fedepoi (approved):

PR #366, opened 2026-07-06
fix(ui): retry duplicate usernames under a hidden @NNNN tag

### LINKED ISSUE #362: web: auth retry storm on duplicate_username (terminal error retried without backoff)
One client ('g1uw44') generated 6,164 `duplicate_username` auth rejections in ~4h — a retry every ~2s — after an unclean disconnect left their original session alive on the relay (or a second tab competed for the name). Several smaller cases (90/78/50 retries) show the same shape: the client treats `duplicate_username` as transient and retries forever with no backoff.

- `duplicate_username` is terminal: retrying the same name can only succeed after the zombie session ages out (up to 180s). The client should stop and surface "name in use — pick another / close other tab", or retry with exponential backoff capped at a few attempts.
- Relay-side (optional hardening): after N consecutive duplicate rejections from the same address, close with a hint or throttle.
- Observability note: this storm is visible in `manabrew_relay_client_rejections_total{reason="duplicate_username"}` (3.2k in 4h) — a candidate for a per-reason alert threshold once client behavior is fixed.

khaliostr:
## Summary

- When the relay rejects a login with `Username '…' is already taken`, the web client now retries under a tagged name (`name@NNNN`, 4 random digits, max 3 attempts) instead of surfacing the error and re-sending the same doomed name on every reconnect tick.
- The auth-result event now carries the username that actually authenticated, so the server store, draft peers, and relay-restart rejoin all track the tagged name. Nothing changes on the wire — the relay and protocol are untouched.
- New `lib/username.ts` (`suffixUsername` / `stripUsernameTag`): the tag is part of the identity (relay routing, seat matching, `roomByName` lookups), so it is stripped only at render time — lobby user list, room player rows, chat senders, and the in-game HUD/zone titles/a11y text.
- Documented the convention in `src/AGENTS.md`.

Fixes #362.

## Why

One stuck client produced ~6k `duplicate_username` rejections in 4 hours: auth failure closes the socket, the reconnect loop retries the identical name forever. Rejecting duplicates is correct relay behavior; the bug is the client retrying an unwinnable name. Tagging client-side ends the storm without touching the protocol, and hiding the tag keeps the second "Ramen" reading as "Ramen".

Known limits (accepted): engine-generated action-log prose still shows the tag; detection matches the relay's error string; two same-named players sharing a room both display untagged.

## Test plan

- `yarn typecheck`, eslint, and prettier clean on all touched files.
- Manual: open two browser sessions with the same username against a local relay — the second authenticates as `name@NNNN` after one retry, shows the bare name in lobby/chat/game HUD, and no repeated `duplicate_username` rejections appear in relay logs.


anacletos (approved):

anacletos (changes requested):
Hold on

anacletos (changes requested):
Yea no.
It can't be done like this.

The username has to be appended with the @NNNN at the very beginning, when it's created, and it has to be stored as such in localstorage.

The current implementation breaks page refreshes, and you won't be able to return to the game because it will choose a different id.

The @NNNN should be set in stone at the moment of username generation, the username becomes something@nnnn, and we just hide the @nnnn

khaliostr:
Reworked as suggested in d00a4de8b — the tag is now set in stone at username creation:

- `usePreferencesStore` appends the `@NNNN` tag when the username is first generated or changed (`ensureUsernameTag`), and the tagged form is what's persisted in localStorage. Existing pre-tag names are migrated once on rehydrate (routed through the setter so the tagged name is written back and stays stable).
- The tag survives renames (changing the base name in Settings keeps your tag), and the Settings input shows/edits the bare name only.
- The collision-time rename in `platform/web.ts` is gone entirely. Page refresh now re-authenticates with the exact same tagged name, so session reclaim and seat retake work unchanged.
- Display stripping is unchanged (lobby, chat, HUD), plus the "(You)" row and bot names now use the bare name too.
- One small leftover in `web.ts`: an auth rejection now bumps the reconnect backoff (socket open resets it before the AuthResult arrives), so the same-name-in-two-tabs case backs off to 30s instead of hammering the relay.

anacletos (approved):

PR #369, opened 2026-07-07

fedepoi:
## Summary
- Adds Ironsmith as a Manabrew trusted engine/runtime option and loads the local sibling `ironsmith-wasm` package.
- Keeps Manabrew's public wire protocol almost unchanged: the only protocol enum addition is `EngineKind::Ironsmith`; Ironsmith-specific private traffic rides over the existing `roomRelay` envelope.
- Delegates the Manabrew compatibility layer to Ironsmith WASM instead of carrying a TypeScript adapter in this repo. Manabrew now calls `validateManabrewMatchConfig`, `startManabrewMatch`, `manabrewView`, `manabrewPublicState`, and `manabrewCommandFromPromptOutput`.
- Removes the Manabrew-side Ironsmith DTO/prompt/action/deck-source mapper, its tests, and the temporary Ironsmith smoke preset deck.
- Publishes one room-wide public `state` from Ironsmith's public Manabrew view, then sends encrypted per-seat Ironsmith state/prompt overlays and remote responses through an Ironsmith room-relay subprotocol.
- Wires the Ironsmith user-facing flow: the lobby can create and identify Trusted Ironsmith match rooms, Play can choose the `ironsmith` runtime for web starts, multiplayer startup carries the selected decks/seats into that runtime, and Ironsmith match validation failures are surfaced before launch.
- Keeps local bot auto-responses for Trusted-mode smoke tests and skips hidden-card placeholder prefetches.
- Makes protocol generation idempotent by de-duping duplicate top-level TypeScript exports emitted by overlapping `ts-rs` export roots.

## Local Ironsmith testing
1. Keep the repositories checked out as siblings so Manabrew can find the local WASM package:
   ```text
   ../ironsmith
   ../manabrew
   ```
2. Build the Ironsmith WASM package from the Ironsmith repo:
   ```bash
   cd ../ironsmith
   ./rebuild-wasm.sh --skip-scryfall-sync
   ```
   If the Ironsmith registry DB or card assets need refreshing, omit `--skip-scryfall-sync`. The Manabrew side expects `../ironsmith/pkg` to contain `ironsmith.js`, `ironsmith_bg.wasm`, and the generated `.d.ts` files.
3. Install/sync Manabrew dependencies from this repo:
   ```bash
   yarn install
   yarn ensure:wasm
   yarn sync:ironsmith
   ```
   `yarn sync:ironsmith` copies `../ironsmith/pkg` into `node_modules/ironsmith-wasm` and clears Vite's cached optimized dependency.
4. Start the local relay and web app in separate terminals:
   ```bash
   yarn dev:relay
   yarn dev:web
   ```
5. Open `http://localhost:1420`, connect to the local relay (`localhost:9443`, default server key `forge`), create a Match room with the `Ironsmith trusted` engine, choose any deck that Ironsmith can validate, add a bot with the same deck, and start the game.

Notes:
- This is Trusted mode only. There is no audit/proof/lockstep consensus path in this PR.
- The host tab owns the authoritative Ironsmith `WasmGame`; host tab loss still ends the v1 match.
- If Ironsmith WASM is rebuilt while Vite is running, rerun `yarn sync:ironsmith` and restart `yarn dev:web`.
- A debug `wasm-pack --dev` build can compile the Rust crate but fail during wasm-bindgen on the unoptimized module with `too many locals`; use the normal rebuild flow above for a usable local `pkg`.

## Validation
- Manabrew: `yarn tsc -p tsconfig.app.json --noEmit`
- Manabrew: `yarn test`
- Ironsmith: `cargo fmt --all -- --check`
- Ironsmith: `CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 wasm-pack build crates/ironsmith-wasm --target web --dev --out-dir ../../pkg --out-name ironsmith --no-default-features --features wasm-lean` compiled the Rust crate, then wasm-bindgen failed on the debug module with `too many locals`.

The protocol-generation step still emits the existing `ts-rs failed to parse serde attribute` warnings.


khaliostr:
Okay, sorry for the mess. We should decide to either:

- submodule ironsmith like we do for forge
- consume ironsmith through crates.io or git cargo dependency

khaliostr:
Closing in favour of #371, which restores the original #346 stack on `fix/concedes-and-leaves`. This rebase onto `main` lost the review-feedback commit (93d9c1cd — relay sender authentication, hello seat verification, peer-key pinning, and the concede prompt-unwind fix with its regression test), so as it stands any room member can re-pin another seat's key and read/inject that seat's private traffic. Rather than patch it here, let's land the reviewed branch on its intended base and bring it to main after #343. Sorry for the back-and-forth, @fedepoi — the branch rewrite under #346 made this messier than it needed to be.

PR #371, opened 2026-07-07

khaliostr:
## Summary

Reopens @Chiplis's Ironsmith trusted runtime (#346) on a clean base. This branch is the exact #346 head — all 5 commits, including `93d9c1cd` *fix: address ironsmith runtime review feedback* — stacked directly on the current `fix/concedes-and-leaves` tip. No code changes.

- Adds Ironsmith as a trusted engine/runtime option loading the sibling `ironsmith-wasm` package; the only wire-protocol change is `EngineKind::Ironsmith`, with Ironsmith-private traffic riding the existing `roomRelay` envelope.
- Delegates the Manabrew compatibility layer to Ironsmith WASM (`validateManabrewMatchConfig`, `startManabrewMatch`, `manabrewView`, `manabrewPublicState`, `manabrewCommandFromPromptOutput`) and removes the TS-side mapper.
- Publishes one room-wide public state, plus encrypted per-seat state/prompt overlays and remote responses over an `ironsmith-trusted` room-relay subprotocol, with relay-authenticated sender checks and pinned peer keys.
- Wires lobby/Play/multiplayer startup for Trusted Ironsmith rooms, keeps bot auto-responses for smoke tests, and makes protocol generation idempotent.

See #346 for the full description and local-testing walkthrough.

## Why

The branch history around #346/#369 got tangled — apologies to @Chiplis for the messy process. #346 shows as merged but `fix/concedes-and-leaves` was rewritten afterwards, so its commits fell off the branch; #369 recreated the work rebased onto `main` but lost the review-feedback commit (`93d9c1cd`), which carried the relay sender-authentication/key-pinning hardening and the concede prompt-unwind fix + test. This PR restores the reviewed state on the base it was written for.

Follow-up (on me): I'll set up the git submodule structure for Ironsmith so we can drop the `../ironsmith` peer-checkout requirement.

## Test plan

Validated locally at this exact head:

- `yarn tsc -p tsconfig.app.json --noEmit` — clean
- `yarn test` — 9/9
- `cargo test -p manabrew-agent-interface` (includes the concede-during-mulligan unwind test) and `-p manabrew-protocol` — pass
- Full `yarn vite:build` with a fresh `../ironsmith` release build (`./rebuild-wasm.sh`) — succeeds; protocol generation is idempotent (clean tree after regen)


khaliostr:
Pushed the submodule + feature-flag follow-up (a358e774d) — this is the structure change I promised, so we can drop the peer-checkout requirement.

**What changed**
- **`ironsmith` is now a git submodule** (`Chiplis/ironsmith`, pinned at `673ce30a4`), configured `update = none` so CI's `submodules: recursive` skips it — the wasm is never compiled in CI.
- **Removed the `file:../ironsmith/pkg` dependency.** `yarn install` no longer needs a sibling checkout.
- **`yarn sync:ironsmith` reads the in-repo `ironsmith/pkg`** and, when it's absent (default on CI / any clone that hasn't built it), installs a tiny **stub** `ironsmith-wasm` package so the app still typechecks, builds, and ships. Wired into `vite:build` / `pretypecheck` / `pretest` / `build:web`.
- **New `ironsmithRuntime` feature flag (default off)** in `src/featureFlags.ts`: the runtime registry only registers Ironsmith when enabled, and the lobby hides the engine option — a shipped build carries zero Ironsmith surface until the submodule is built and the flag flipped.

**Verified locally, both paths**
- Flag off + stub → `tsc`, `yarn test` (9/9), `yarn vite:build` all pass; the 45 MB `ironsmith_bg.wasm` is **not** bundled.
- Flag on + built pkg → same checks pass; the real wasm bundles as expected.

**To enable Ironsmith locally:** `git submodule update --init ironsmith` → `./ironsmith/rebuild-wasm.sh` → `yarn sync:ironsmith` → flip `ironsmithRuntime` to `true`.

@Chiplis — sorry again for the messy branch shuffle getting here. Your runtime code is untouched; this only changes how the wasm is vendored and gates it so it can ship dark. Submodule URL points at `Chiplis/ironsmith` for now; we can repoint it to the org whenever the repo moves.

khaliostr:
## Tested end-to-end (web + desktop) — two runtime bugs found & fixed, e2e tests added

Ran this in an isolated worktree with the `ironsmith` submodule built, the flag on, a local relay, and Playwright driving the real web client.

### Bugs found and fixed (pushed here)
1. **Match start never produced a game view.** `manabrewView` / `manabrewPublicState` build their return from a `serde_json::Value`, which serde-wasm-bindgen encodes as a JS **`Map`**, not a plain object (unlike the typed struct `validateManabrewMatchConfig` returns — which is why validation worked but the view read didn't). `view.state?.gameView` is `undefined` on a Map, so **every** start threw *"Ironsmith WASM did not return a Manabrew game view"*, even for a deck Ironsmith accepts. Fixed by deep-normalizing Maps→objects at the boundary (`plainify`). Upstream, Ironsmith could instead serialize these with `serialize_maps_as_objects(true)`; the Manabrew-side normalize makes us robust either way.
2. **Dev server couldn't load the wasm.** Under `yarn dev:web`, `/node_modules/ironsmith-wasm/ironsmith_bg.wasm` was served as the SPA HTML fallback → `WebAssembly.instantiate` magic-word crash. Fixed via `optimizeDeps.exclude: ["ironsmith-wasm"]` + `assetsInclude: ["**/*.wasm"]`. (Production `vite build` was already fine — it bundles the asset.)
3. (Earlier in this PR) `metrics.rs` non-exhaustive `EngineKind` match — CI green now.

### What works now
- Web **production build** and **dev server** both start an Ironsmith match; the board renders and the game reaches **mulligan → priority**.
- **Single client + bot** and **two humans** both reach a live board — the two-human run exercises the encrypted per-seat relay (ECDH hello + AES-GCM overlays), and the bot-seat naming (`*-bot-*`) matches the runtime's detection heuristic.
- Lobby integration is correct: Ironsmith engine tile + "Trusted mode" warning appear only with the flag on; relay auth, room create/tag, and game-start signaling all work.

### e2e tests added (`tests/e2e-ironsmith/`)
`yarn test:e2e:ironsmith` (client+bot) and `:pair` (two humans) — raw Playwright against a real relay, system Chrome, no Chromium download. README covers prereqs + env knobs. These double as a reusable multiplayer driving harness.

### Notes / not-blockers-for-this-PR
- **Deck support is experimental.** Many presets are rejected at validation (e.g. `Izzet Lessons` → `Gran-Gran` unsupported); the simple `real_*` / `starter_deck_*` pools (Mono Red Prison, Workshop…) validate. The rejection is surfaced to the UI and the game doesn't start — that path works as intended.
- **Desktop:** the Tauri shell loads the same web bundle + `IronsmithTrustedGameApi`, so this coverage applies to desktop. The Rust crate compiles; a full native bundle additionally needs the `forge` submodule + harness build (standard prereqs, unrelated to Ironsmith).
- Flag still ships **off**; I verified `tsc` + production build stay green with it off.

khaliostr:
Upstream fix filed: **Chiplis/ironsmith#739** — makes the `manabrewView`/`manabrewPublicState`/`manabrewPrompt` bridge return plain JS objects (via `Serializer::json_compatible()`) instead of `Map`s. Verified in isolation: rebuilt the Ironsmith wasm with that fix, synced it here, and ran the e2e **with this PR's `plainify` workaround bypassed** — the game still starts and reaches a live board. Once #739 lands and we bump the submodule, the `plainify` normalize in `ironsmithRuntime.ts` can be dropped (harmless to keep either way). Fork lives at `witchesofthehill/ironsmith`.

anacletos (approved):

PR #379, opened 2026-07-07
fix(engine): M1 parity quick-fixes + workspace clippy cleanup

fedepoi:
## Summary

M1 quick-fix batch from the engine-parity roadmap — six independent engine-vs-Java correctness fixes, each mirroring a specific Forge rule, plus a workspace-wide lint cleanup.

- **FlipCoin dispatch** — `ApiType::FlipACoin` → `FlipCoin` so all 84 `AB$/DB$ FlipCoin` cards actually resolve instead of parsing with `api=None` and silently no-opping.
- **transform() state preservation (CR 712.8)** — stop zeroing marked damage / P-T modifiers / granted keywords on a face change; those live on the object, not the face.
- **`Condition$` shorthand** — expand `Threshold/Metalcraft/Delirium/Hellbent/Revolt/Desert/Blessing/Kicked` into the IR condition flags at parse time (mirroring `SpellAbilityCondition.setConditions`); previously every value except `Kicked` fell through and the gated effect always resolved.
- **Free first mulligan** — bottom `timesMulliganed − 1` when the game started with more than two players (Commander pods were over-bottoming one card after the first mulligan).
- **Dead-player turn rotation** — drive turn advance through `Game::next_player`, which skips players who left the game (mirrors `PhaseHandler.getNextActivePlayer`); eliminated players kept getting turns in 3+ player games.
- **ExchangeZone dispatch** — route the `ExchangeZone` script name to the real `ZoneExchangeEffect` (as Forge's `ApiType` does) instead of plain `ChangeZoneEffect`, and drop the unreachable `ZoneExchange` twin variant.
- **Workspace clippy/fmt cleanup** — clear ~90 pre-existing clippy + rustc warnings and two `E0004` non-exhaustive matches that had left `parity-debugger` unable to compile, so `cargo clippy -D warnings` and `cargo fmt --check` pass again.

## Why

The engine-port audit found the parity suite red after ~3 weeks of dark CI. These are the highest-blast-radius, well-scoped correctness gaps from that audit's quick-fix batch — each is a general rule ported from `forge/forge-game/`, not a card-specific patch. Separately, the workspace `cargo clippy -D warnings` gate had been red on `main` (and `parity-debugger` didn't compile at all), so `yarn lint:rust` couldn't pass; this restores it.

## Test plan

- `cargo check -p manabrew-engine` — clean after each fix.
- `cargo fmt --all -- --check` — passes.
- `cargo clippy --workspace --exclude manabrew --lib --bins -- -D warnings` — passes (previously ~90 warnings + 2 compile errors). `manabrew` (src-tauri) is excluded because it needs `forge-harness/native/build/libforgeharness.dylib`, which is built in CI, not locally.
- `yarn lint` (eslint + prettier + tsc) — passes.
- Fixes verified by mirroring the corresponding Java in `forge/forge-game/` (`ApiType.java`, `Card.setState`, `SpellAbilityCondition.setConditions`, `MulliganService`/`LondonMulligan`, `PhaseHandler.getNextActivePlayer`, `ZoneExchangeEffect`). Full parity-suite verification is deferred: the current red regression entries trace to a deterministic-agent RNG-alignment issue in the Java harness (the "parity engine"), which is being fixed separately — so parity CI stays disabled until that lands rather than being re-enabled red.


anacletos (approved):
mmmmh not sure why this was made? Approved though

fedepoi:
closing for now , we might reopen this in future

PR #426, opened 2026-07-08
feat(parity): trace-replay parity harness

khaliostr:
## Summary

Adds a new `parity-replay` bin crate that replays a captured Forge game trace through the Rust manabrew engine and diffs the engine's `GameViewDto` against the Forge states embedded in the trace, surfacing parity divergences.

Modules:
- `trace.rs` — parses a trace JSONL into `{ header, decisions: Vec<(state_before, prompt, response)> }`, reusing the real protocol DTOs (`GameViewDto`, `AgentPrompt`, `PromptOutput`). Also extracts opening hands (from the recorded mulligan prompts) and a best-effort partial decklist (distinct visible cards per seat).
- `replay_agent.rs` — a `Responder` (wrapped in `PromptAgent`) that captures the engine's `GameViewDto` via `present`, pops recorded responses in engine order (type-matched to tolerate skipped setup prompts), and **re-resolves recorded card ids by name/position against the live prompt** — it never echoes Forge's ids.
- `diff.rs` — perspective-aware `GameViewDto` diff comparing by card name (ids differ between engines), classifying hidden-library-dependent diffs (hand contents, library count) as the expected boundary.
- `main.rs` — CLI: load trace (+ optional deck json) → build game with forced opening hand → drive the engine via `PromptAgent<ReplayAgent>` + `GameRuntime` (`run_opening_hand_actions` then `run_turn` loop) → report the first divergence (or "clean through N decisions").

## Why

We have hundreds of captured hosted-game traces from launch but no way to check whether the Rust engine reproduces Forge's decisions on real games. This harness turns any captured trace into a replay-and-diff parity check, complementing the synthetic `parity` matrix with real game data.

## Test plan

- `cargo check -p parity-replay` — clean (no warnings/errors from the crate).
- End-to-end run against a sample trace drives the real engine and reports a field-level diff:
  ```
  cargo run -p parity-replay -- sample-traces/sample.jsonl \
    --cardset <path/to/cardset.rkyv> \
    --cards-dir <forge>/forge-gui/res/cardsfolder
  ```
  Current divergences on the sample are all attributable to the two documented data gaps below (partial reconstructed deck → library count / hand contents) plus first-player determination, not to engine drift.

## Known data gaps (where the real joins plug in)

1. **Decklist completeness** — the trace only contains opening hands + revealed cards, so the reconstructed deck is partial. The full list comes from the analytics `DeckSelected` event (a separate stream), joined on the deck name. `build_identities` in `main.rs` is the plug-in point; `--deck0/--deck1 <deck.json>` already accept a full list.
2. **Library order capture** — the trace never exposes library order, so the forced opening reproduces hand *contents* but not draw order; any post-opening draw is where a real (non-bug) divergence first appears. `force_opening_hand` in `main.rs` is where a captured library order would be applied.

Additionally, first-player determination is not yet replayed from the trace's opening dice roll (engine defaults to seat 0), which shows as an `activePlayerId` diff on turn 1.

## Notes

- No production crates modified beyond registering the new member in the workspace `Cargo.toml`.
- Sample traces under `sample-traces/` and `docs/CAPTURE_FORMAT_V2.md` are left untracked in the branch for now — say the word and I'll add fixtures.


khaliostr:
## Scope: manual engine override ("maintenance mode") + parity reconcile

Prompted by the Reddit feedback (Dimmins2) about a tabletop/maintenance mode. Key realization: **maintenance mode and the parity harness want the same primitive** — a typed, validated base-state edit API on `&mut GameState`. Build once, expose to both.

### The shared primitive (production-crate work — NOT in this PR)
- **Protocol:** new `ClientToServerMessage::Directive(Directive::Maintenance(MaintenanceEdit))`, riding the existing non-prompt directive channel (precedent: `Directive::Concede`).
- **Engine edit API** on `&mut GameState`, base state only (source of truth, edits stick):
  `set_life`, `set_poison`, add/remove counter (player + card), `set_tapped`, `move_card(zone)`, `set_mana`, add/remove perpetual effect.
- **Derived state via causes, not effects.** P/T, keywords, colors, granted abilities are recomputed every state check by the CR 613 layer system (`staticability/layer.rs`), so a direct "set power = 5" is clobbered on the next recompute. The editor surfaces *causes* (add counter / add perpetual / move zone) and lets the layer system repaint. This is the one thing to get right up front.
- **SBAs after edit:** run state-based actions after each edit (so life→0 actually kills, legend rule fires); make it explicit/toggleable in the UI.

### Freeze / resume
The synchronous loop already parks at every priority prompt — "freeze" is just "we're between prompts." Maintenance edits arrive on the directive channel, mutate `&mut Game`, broadcast the new `GameViewDto`, then resume. No suspension machinery needed.

### Capture + integrity
- Log every edit into the game-capture stream (`MANABREW_GAME_CAPTURE_DIR`) as a new frame kind, so (a) replays stay faithful and (b) the parity harness can tell "engine did X" from "human patched X."
- Host-authoritative in hosted games; gate behind an explicit both-players toggle; log everything. Casual/paper-companion territory (untap.in style), so social contract carries the rest.

### Parity consumer: patch-based reconcile (the follow-up payoff)
Add a `reconcile(&mut GameState)` hook (new `PlayerAgent` callback or `GameRuntime` step) invoked at a divergence point. On divergence the harness patches Rust's **base state** to the trace's captured state and continues — isolating each divergence to the single action that caused it, and letting the replay push past the hidden-library boundary that currently caps every game. Turns the 837-game corpus into a ranked bug backlog. (Base-state resync only; stack / continuous-effect divergences still abort. Same tiering as maintenance mode, so the two efforts fund each other.)

### Shipped in THIS PR (in-crate precursor)
`--continue-on-divergence` (+ `--max-divergences`): a single replay now records **every** divergence instead of aborting at the first. It's *uncorrected* — post-first diffs cascade from the initial unreconciled difference (demonstrated: after a turn-2 wrong draw, the replay stayed aligned ~13 more turns, then everything drifted at turn 15). That cascade is exactly why the patch-based reconcile above needs the edit API. Default behavior is unchanged (abort-at-first single verdict).

### Phasing
1. (this PR) continue-on-divergence collector.
2. base-state edit API + `Directive::Maintenance` + capture frame + host/consent gate → maintenance-mode MVP (life/counters/zones/tapped).
3. reconcile hook wired to the edit API → patch-based parity.
4. derived-state-via-causes editor; trigger/stack injection (the expensive 20%) last.

khaliostr:
Design and scope for the maintenance-edit work prototyped on this branch now live in #669. This PR keeps its own scope: the trace-replay harness and the divergence collector. The epic covers the shared primitive, the Forge side of it, and the open questions about authority and capture.

khaliostr:
Design moved to #669 in August. What is left here reads pre-patch captures (everything since 2026-08-20 is stateDelta) and needs the reconcile hook to get past the hidden-library boundary, so parking it. Branch `worktree-parity-trace-replay` kept.

PR #443, opened 2026-07-12
fix: rooms stuck, bad starts, zombie rooms

anacletos:
## Summary
- The relay mints a `game_id` per game and it is now required on the wire, relay only source of truth on active / available rooms
- The engine session is tagged with the relay's `game_id`. If the relay contradicts it (room reset to Lobby, or a different game started) the node aborts
- `.taxe()` can't wedge. Turn take into a timeout poll, so that every 1s, it checks if the game is supposed to be continuing. That unlocks the `.take()` deadlock in case the thread is waiting for an answer of a dead game.
- Games with no human seats end immediately
- Two Grafana alerts on existing metrics (`node-session-wedged`, `abandoned-game-spike`).
- Fix ghost chilling users


.github/workflows/build-checks.yml
```diff
@@ -147,6 +147,37 @@ jobs:
       - name: Build manabrew-server (release)
         run: cargo build --release -p manabrew-server
 
+  # TODO: Group these into a nice networking regression suite
```
fedepoi: what about we do this todo in this pr ? doesn't look too complicated it's actually a good idea

fedepoi (dismissed):

.github/workflows/build-checks.yml
anacletos: Done

PR #445, opened 2026-07-12
feat: manabrew-protocol v1

anacletos:
## Summary

Protocol V1 :)


manabrew-rs/crates/manabrew-protocol/src/game/mod.rs
```diff
@@ -1,53 +1,131 @@
...
+use std::collections::{BTreeMap, HashMap};
 
 use serde::{Deserialize, Serialize};
 use ts_rs::TS;
 
 use crate::prompts::common::TargetRef;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, TS)]
+#[ts(export, export_to = "game/index.ts")]
+pub enum ManaColor {
```
khaliostr, line 10: we could use a strum enum here to match https://github.com/witchesofthehill/manabrew/pull/445/changes#diff-6bc8801bb10a7935b6e21caefee840717561e21cd85aa5fdab442a33373c9b9e
 for tostring

manabrew-rs/crates/manabrew-protocol/src/prompts/mod.rs
```diff
@@ -97,6 +100,57 @@ pub enum PromptOutput {
...
+pub enum ResponseViolation {
+    WrongPromptType,
+    UnknownActionId(String),
+}
+
+impl PromptInput {
+    /// The formal prompt/response contract: a response is valid only if its
+    /// output family matches this prompt and every echoed action id was
+    /// advertised by it. Engines reject anything else with a `ProtocolError`.
+    pub fn validate_response(&self, output: &PromptOutput) -> Result<(), ResponseViolation> {
```
khaliostr, line 119: top

manabrew-rs/crates/manabrew-protocol/src/protocol.rs
```diff
@@ -2,10 +2,20 @@ pub use crate::deck_dto::Deck;
 use serde::{Deserialize, Serialize};
 use ts_rs::TS;
 
-pub const PROTOCOL_VERSION: u32 = 2;
-
-#[cfg(test)]
-const PROTOCOL_SCHEMA_FINGERPRINT: &str = "b8f0eb5cf679220a";
+// The wire-compat number IS the crate major: a breaking wire change must ship
+// as a breaking (major) release of this crate, never as a separate constant.
```
khaliostr, line 6: how do we enforce that a bump on this crate must enforce a corresponding brekaing change? so that downstream fails if out of sync

manabrew-rs/crates/manabrew-protocol/src/protocol.rs
```diff
@@ -352,83 +365,3 @@ pub enum EngineKind {
...
-        let _ = std::fs::remove_dir_all(&dir);
-        schema
-    }
-
-    #[test]
-    fn wire_schema_is_pinned() {
-        let actual = format!("{:016x}", fnv1a(rendered_schema().as_bytes()));
-        assert_eq!(
-            actual, PROTOCOL_SCHEMA_FINGERPRINT,
-            "protocol wire schema changed — bump PROTOCOL_VERSION and update PROTOCOL_SCHEMA_FINGERPRINT"
```
khaliostr, line 433: Idea is a mechanism similar to this that force a major bump if a change is detected. quesiton is how big of a change would trigger it

manabrew-rs/crates/self-hosted-node/src/engine_backend/java_backend.rs
```diff
@@ -1163,6 +1245,23 @@ fn wait_for_prompt<B: JavaBridge>(
     Ok(None)
 }
 
+#[cfg(forge_backend)]
+fn send_observer_state(
```
khaliostr, line 1249: top

manabrew-rs/crates/self-hosted-node/src/host.rs
```diff
@@ -121,12 +121,16 @@ struct HostSnapshot {
...
+    last_state_by_slot: HashMap<String, Value>,
     pending_prompts: HashMap<String, Value>,
 }
 
 type SharedHostSnapshot = Arc<Mutex<HostSnapshot>>;
 
 const RECONNECT_BACKOFF_SECS: [u64; 6] = [1, 2, 4, 8, 15, 30];
+
+/// Pseudo-seat used by per-recipient backends for the public (spectator) view.
+pub(crate) const OBSERVER_SEAT: usize = usize::MAX;
```
khaliostr, line 133: 💯

src/components/game/game.constants.ts
```diff
@@ -1,19 +1,27 @@
-/** Phase bar definitions — must match phase_to_step() in src-tauri/src/game_view_dto.rs */
-export const PHASES = [
+import type { StepKind } from "@/protocol";
+
+/** The single UI-side list of turn steps, ordered to match the engine's turn
+ *  structure. Ids are protocol `StepKind` values — never restate them elsewhere. */
```
khaliostr, line 4: very nice

src/AGENTS.md
```diff
@@ -72,11 +72,11 @@ Commander deck analysis (combos, Game Changers, bracket estimate) lives in `stor
...
...nnel carries **four separate message families**, never conflated: `state` (a `StateUpdate` = `{ gameView }`, the sole carrier of game state), `display` (a `DisplayEvent` animation hint), `prompt` (a `Prompt` call-to-action that carries **no** gameView), and `error` (a `ProtocolError` rejecting the last response — the engine validates every response against the open prompt via `PromptInput::validate_response` and re-sends the prompt right after the error, so client recovery is just unblocking the UI). The store has one apply fn each — `applyState` / `applyDisplay` / `applyPrompt` / `applyProtocolError` (`stores/gameStore.constants.ts`) — and `useGameEventListeners` routes `game:state` / `game:display` / `game:prompt` / `game:error` (plus `game:remote_*` for relay seats) to them. State envelopes carry `forPlayer` when the engine emits per-recipient views (Java backend; hidden hands/library/face-down exile per Forge's `canBeShownTo`) and clients apply only their own; an unaddressed state is a public view every seat applies. A prompt or error only becomes actionable when `forPlayer === myPlayerSlot`. Prompt modals read gameView from the store (via `ctx.gameView`), never off the prompt.
```
khaliostr, line 79: very nice. ironsmith should slot in nice with hidden views and privileged information

website/src/content/docs/protocol/index.mdx
```diff
@@ -12,8 +12,10 @@ engine that speaks this protocol can drive the Manabrew client, and any client
 that implements it can host the Manabrew engine. This section documents the
 protocol so you can implement it for your own engine.
 
-:::caution[Work in progress]
-The Manabrew protocol is still evolving and we're working hard towards a **v1.0.0**, after which, we'll start honoring [semantic versioning](https://semver.org/)
+:::note[Versioning]
+This is **v1** of the protocol, versioned as the
+[`manabrew-protocol`](https://crates.io/crates/manabrew-protocol) crate under
+[semantic versioning](https://semver.org/).
```
khaliostr, line 18: magic !

website/src/content/docs/protocol/index.mdx
```diff
@@ -51,6 +53,23 @@ are never conflated:
     An `AgentPrompt` — a **call to action**. The engine has paused and needs a decision from a
     specific player. Carries **no** game state.
   </Card>
+  <Card title="error" icon="warning">
+    A `ProtocolError` — the engine **rejected** the last response. See
+    [Errors](#errors).
+  </Card>
+</CardGrid>
```
khaliostr, line 60: much needed

manabrew-engine-neutral-protocol-design.md
```diff
@@ -0,0 +1,585 @@
...
+   decision outputs correlated to prompts.
+3. **Targets and choices are distinct.** A rules target is not the same thing as
+   a card chosen for discard, sacrifice, reordering, payment, or display.
+4. **Prompts are stable contracts.** Every response is validated by `promptId`,
+   prompt family, deciding player, and response shape.
+5. **Transport wrappers do not change payload semantics.** Direct responses and
+   relayed responses wrap the same `PromptResponse` object.
+6. **Core shapes are typed; extensions are namespaced.** Engines can innovate
+   through declared vendor extensions, but core clients can ignore unknown
+   extension fields safely.
```
khaliostr: i think this is great

manabrew-engine-neutral-protocol-design.md
```diff
@@ -0,0 +1,585 @@
...
+
+type TargetRef = {
+  object: ObjectRef;
+  targetRole?: "source" | "target" | "assignee" | "defender";
+  intent?: string;
+  oracle?: string;
+};
+```
+
+node: what is the difference between ObjectRef and target ref? when is one supposed to use one and not the other?
```
khaliostr: I think the document says it a bit earlier. 

**Targets and choices are distinct.** A rules target is not the same thing as
   a card chosen for discard, sacrifice, reordering, payment, or display.

i suppose TargetRef would be the more specialised type. Object ref the sacrifice object

manabrew-engine-neutral-protocol-design.md
```diff
@@ -0,0 +1,585 @@
...
+    | "unsupportedResponse"
+    | "unsupportedExtension";
+  message: string;
+  promptId?: PromptId;
+  stateSeq?: StateSeq;
+};
+```
+
+Clients should treat these as recoverable unless the engine marks the session
+terminated.
```
khaliostr: very nice

manabrew-rs/crates/manabrew-protocol/src/protocol.rs
khaliostr: so had a quick go at the enforcement side locally. we could hash the generated TS bindings into a committed protocol.lock.json (version + fingerprint) so CI fails if the shape hash moves without a version bump. adding an EngineKind variant without bumping makes ci go red and reverting is green again.

it would only force a bump though, doesn't tell us how big. the hash just says something changed so we'd still pick major vs minor by hand. to automate it we could store the shape itself in the lock and diff it. adding an optional field or a variant to an open enum should be safe. removing or renaming things, changing a type, making an optional field required, or a new variant on a closed enum would all break old peers, so those would be major. cargo-semver-checks covers the rust api but the wire shape is what the client actually sees, so a diff over the ts-rs output is more accurate.

only holds if our types are forward compatible though. if the enums don't have a catch-all an old client just fails to deserialize any new variant, and then everything is breaking and the versioning barely helps. that'd be a change to the types not tooling

runtime side we'd also want to send version+fingerprint on connect and reject mismatches, which outdated_wire partly does already.

manabrew-rs/crates/manabrew-protocol/src/protocol.rs
anacletos: So my rationale is it's up to the developer to be mindful of semver actual rules.
Once we publish the protocol on crates.io and npm, then protocol changes would go in standalone, hence would be able to bump major on breaking changes E.g. `feat!(protocol): something non backward compatible`

Since all dependents would consume the crates/npm version, they wouldnt even have to update immediately. So pushing breaking changes to a new major version wouldn't break anything up until one decides to update on the consumer side. Then it's up to the consumer to make sure the whole system is updating in a healthy way.

One thing genuinely broken about this is that currently the manifest would update on a protocol change in the repo, triggering a wrong safety net.
Happy to take suggestions on how to handle it.

Overall I'd rather not have to update a fingerprint and a second version in the code, that's usually not how libs handle their packages. I think it helps avoid a lot of headache if the semver rules are respected, and the crate version on the repository becomes the one source of truth.

manabrew-rs/crates/manabrew-protocol/src/protocol.rs
anacletos: I think all these problems go away once we move the responsibility on the consumer side.
If the rule is all consumers take the package from crates/npm, then it's on them to pull a version and realise it's breaking.

My idea is pushing versions of the protocol won't affect anything, it's just a new version of the lib

manabrew-engine-neutral-protocol-design.md
anacletos: Ah sorry this is actually not implemented, and likely rejected for V1 😅

anacletos: Reason being: extensions are really really hard lol

manabrew-engine-neutral-protocol-design.md
anacletos: Sorry I need to remove this doc it wasn't meant to be pushed

forge-harness/src/main/java/forge/harness/host/InteractiveSnapshotExtractor.java
```diff
@@ -54,15 +56,56 @@ private static Map<String, Object> buildGameView(
...
+                }
+            }
+            zones.add(visibleZone("command", ownerId, game, commandZone, index, true));
+            // Library bulk stays hidden (count only); the top card is included
+            // when the viewer may see it (Augur/Courser-style statics), so
+            // `count >= cards.length`.
+            final List<JsonObject> libraryCards = new ArrayList<>();
+            final CardCollectionView library = player.getCardsIn(ZoneType.Library);
+            if (!library.isEmpty() && shownTo(library.get(0), viewerPlayer)) {
+                libraryCards.add(visibleCard(toCard(game, library.get(0), index, true)));
```
fedepoi, line 84: very much needed !!!! awesome

website/public/protocol/declare-attackers.webp
fedepoi, line 1: some of this screen are also valuable for the readme and the actual game shall we point it there too ? im thinking manabrew about

anacletos: They can be used anywhere. Im mostly wiring them to the docs but no reason why you wouldn't use them elsewhere

fedepoi (approved):

Tyrathalis:
I've run the AI-decision-coverage pass that we discussed and checked this prompt surface against our PlayerController callback census, which has 109 methods, 64 of which fire in our current (highly limited) card pool. This structure definitely matches a lot of what agents need. Engine-legal advertised actions, composite one-shot combat declarations, and UseResource for convoke/improvise all match our perception of the best practices here. Here are a few additional items that I think would be worth considering:

1. **Response provenance.** Nothing marks an answer as agent-made vs bot-takeover vs engine auto-fill (Pay{auto} excepted). That creates a potential trap when training on data collected through this protocol, since mid-game bot substitutions could create fake or meaningless wins and losses. I think the cheapest high-value addition here would be to add an answered_by field in Response, or similar.

2. **Deterministic seeds.** Currently only draft and sealed formats have lobby seeds. GameStarted doesn't have one, and the replay cache just sends a current-state snapshot for reconnection. Seeding all games and verifying reproducibility would be helpful for any training project.

3. **No mid-cast abort.** Several RL bot projects have run into issues where the model chooses a spell but gets forced into a bad target and can't back out. In theory models should be able to train around that, but since it's come up multiple times I think it's worth considering adding optional cancellation support. Let me know if you'd like to see how we're handling this.

4. **GameOver is empty.** We nearly lost a corpus of value labels to a winner-attribution bug, so supplying an explicit termination state would provide a handy authoritative certification.

5. **Noncombat amount allocation.** Some splits, like fireball or counter distribution, don't have obvious prompts. Again, could be worked around in practice, but would be a nice feature.

6. **prompt_id join semantics.** It's worth documenting these more thoroughly. (game_id, prompt_id) would be a sound trajectory key for training joins if these are guaranteed unique. Fresh ids per re-issue would be a safe convention.

PR #448, opened 2026-07-14
ci(deploy): pull pre-built ghcr images instead of building on the host

khaliostr:
## Summary

Prod builds every image locally in `deploy.sh`, but the web image is a ~1h WASM+Vite build that no longer fits the host's 38 GB disk — the Ironsmith trusted runtime tipped it into **ENOSPC → failed deploys**. CI already builds and pushes these images to GHCR, so this switches prod to **pull the pre-built images** instead of rebuilding.

- **`docker-images.yml`**: add **`manabrew-hub`** to the build matrix (was only server/web/node), so the deck-hub API is a pulled image too.
- **`compose.production.yml`**: `manabrew` (web), `manabrew-server`, `manabrew-hub`, `self-hosted-node` → `image: ghcr.io/witchesofthehill/:${MANABREW_IMAGE_TAG:-latest}`, no local `build:`. The web image is relay-agnostic, so its relay endpoint moves from `VITE_RELAY_*` build args to **`RELAY_*` runtime env** (the entrypoint writes `config.js`). `parity-dashboard` still builds (Java, not published to GHCR).
- **`deploy.sh`**: pull the GHCR images (retry until CI has pushed them), recreate **only** services whose image actually changed; the relay keeps its **binary-diff guard** so an unrelated image change never restarts it and drops live games; final `up -d` gains **`--no-deps`** so recreating the web never drags the relay down as a side effect (this is what took the site down during the manual migration).

### Why this is also more robust
- **Fail-safe:** if a CI image build fails, no new image is pushed → deploy pulls the **last-good** image → prod keeps running. Today a failed build aborts the whole deploy.
- **No host toolchain / disk pressure:** the host only pulls; the heavy builds run on CI runners with adequate disk.
- **Deck hub** finally ships as a real image (it was never built/pushed before — the container simply didn't exist in prod).

## Known tradeoff
The generic GHCR web image is built **without** `VITE_SCRYFALL_SYMBOL_BASE`, so mana symbols load from `svgs.scryfall.io` directly rather than the `/scryfall-symbols/` Caddy proxy. Cosmetic/perf only — if the proxy matters, make it runtime-configurable (like `RELAY_*`).

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -51,6 +51,8 @@ jobs:
             dockerfile: Dockerfile.web
           - image: manabrew-node
             dockerfile: manabrew-rs/crates/self-hosted-node/Dockerfile
+          - image: manabrew-hub
+            dockerfile: manabrew-rs/crates/manabrew-hub/Dockerfile
     steps:
       - uses: actions/checkout@v4
         with:
--- a/compose.production.yml
+++ b/compose.production.yml
@@ -1,20 +1,14 @@
 services:
   manabrew:
-    build:
-      context: .
-      dockerfile: Dockerfile.web
-      args:
-        VITE_RELAY_HOST: relay.manabrew.app
-        VITE_RELAY_PORT: "443"
-        VITE_RELAY_PASSWORD: ${MANABREW_SERVER_KEY:?MANABREW_SERVER_KEY is required}
-        VITE_MANAGED_RELAY: "1"
-        VITE_SCRYFALL_SYMBOL_BASE: /scryfall-symbols/
-        # Off by default: "Play vs AI" runs the engine client-side, so the web
-        # does NOT depend on the self-hosted-node. Build with
-        # VITE_HOSTED_AI_ENABLED=true (and run the node via --profile hosted-ai)
-        # to expose the Settings engine toggle.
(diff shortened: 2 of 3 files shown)


khaliostr:
Added the two robustness follow-ups (commit 912367fbe):

- **Self-update re-exec** — after `git pull`, if `deploy.sh` itself changed, re-exec the updated script once (`DEPLOY_ORIG_PREV` preserves the deploy range). This closes the trap where deploy.sh edits only took effect the *next* deploy — the exact reason Ironsmith shipped dark first time.
- **Health-gated rollout + rollback** — snapshot each service's running image, roll out with `up -d --wait` (blocks on healthchecks), and on failure re-tag the ghcr services back to their previous image and recreate them. A bad image now self-heals instead of leaving prod broken.

Both still need the staging run before merge (same caveat as the rest of the PR).

### The two remaining items I mentioned need your call, not just a follow-up:
- **Full installer/deploy parallelization** — We need to have backward compaitbility in order to build and deploy web in parallel
- **Scryfall symbol proxy** — almost certainly a non-issue: the generic image loads symbols from `svgs.scryfall.io` directly (a CORS-friendly static CDN), so the `/scryfall-symbols/` proxy isn't required for correctness. I'd leave it unless symbols actually misbehave, then make it runtime-configurable like `RELAY_*`.

anacletos (approved):

PR #450, opened 2026-07-14
feat(web): in-game board settings, unified card sizing, scry hover preview

khaliostr:
Game-UI quality-of-life batch: the big card preview in scry/surveil, an in-game settings surface, zone-pile locking, and a rework of card sizing into a single honest multiplier — plus a Playwright UI e2e suite wired as `cargo xtask e2e-ui`.

## Scry / surveil big preview

`ScryModal` (both scry and surveil route through it) now shows the large hover `CardPreview`, wired like the zone viewers via delegated mouse handlers, and dismisses when a drag starts. Touch intentionally keeps hold-to-drag: the app-wide 450ms long-press preview can't coexist with the modal's 200ms `TouchSensor` drag activation.

scry hover preview

## In-game Board settings

New **Board settings** entry in the board gear menu opens a modal usable mid-game — card size, card style, battlefield layout, zone-pile lock, animations, preview trigger + delay. Everything writes `usePreferencesStore`, so changes apply to the live board instantly and persist.

board settings modal

## Zone-pile lock

New `lockZoneTiles` preference freezes the deck/graveyard/exile/command tiles against accidental drags (tap-to-open keeps working), threaded `BoardCanvas → BoardScene → BoardRegion`. Exposed in the in-game modal and the Settings page.

## Unified card sizing — one knob, 75–300%

`cardSizeMultiplier` replaces `battlefieldCardScale` (and an interim hand-scale pref): **100% = the classic 3-row board, up to 300%**, applied to *every* battlefield and clamped per field so **at least 2 rows always fit** (per review — a 1-row board is unplayable).

- Scale and the combat-row band are now solved **jointly** in closed form (`battlefieldScaleForMultiplier`); the old two-pass estimate starved the grid at 1–2 rows, which is why "max size" used to land at ~0.85× scale.
- The hand fan keeps its classic size and the board grows **up to match it** (the fan only enlarges on displays tall enough for board cards to outgrow it) — so big sizes end near a 1:1 hand-to-board ratio instead of the fan swallowing the field.
- `placeZoneTiles` gained a fallback for giant single-row grids: when the hand + action-cluster keep-outs leave no free cell, a tile overlaps a blocker instead of stranding at stale geometry.
- Old persisted size prefs are ignored; everyone starts at 100% (identical to the classic board).

board at 300% (2-row cap)

## UI e2e suite

`tests/e2e-ui/board-settings.mjs` drives a real bot game (reusing `tests/e2e-ironsmith/lib.mjs`) and checks the gear menu → settings modal, size + lock persistence, and the scry hover preview (injected via a dev-only `window.__gameStore` seam, never present in production builds). Run with `cargo xtask e2e-ui` against the usual dev stack (relay :9443 + `yarn dev:web`).

## Test plan

- [x] `yarn tsc --noEmit`, `eslint`, `prettier` clean; `cargo check -p xtask`
- [x] `yarn vitest run` (pixi board tests)
- [x] `cargo xtask e2e-ui` passes end-to-end against a live relay + dev server (screenshots above are its `SHOT` output)
- [x] Manually verified 100% is pixel-identical to the classic board; 300% saturates at a 2-row fill on both fields

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/game/game.constants.ts
+++ b/src/components/game/game.constants.ts
@@ -87,3 +87,9 @@ export const RING_ABILITIES: readonly string[] = [
   "Whenever your Ring-bearer becomes blocked by a creature, that creature's controller sacrifices it at the end of combat.",
   "Whenever your Ring-bearer deals combat damage to a player, each opponent loses 3 life.",
 ] as const;
+
+// Card preview hover-delay slider bounds — shared by the Settings page and the
+// in-game board settings modal.
+export const HOVER_DELAY_MIN = 100;
+export const HOVER_DELAY_MAX = 1500;
+export const HOVER_DELAY_STEP = 50;
--- a/src/components/game/modals/GameSettingsModal.tsx
+++ b/src/components/game/modals/GameSettingsModal.tsx
@@ -1,0 +1,198 @@
+import { Modal } from "./Modal";
+import { Button } from "@/components/ui/button";
+import { Label } from "@/components/ui/label";
+import {
+  CARD_SIZE_MULTIPLIER_MAX,
+  CARD_SIZE_MULTIPLIER_MIN,
+  usePreferencesStore,
+} from "@/stores/usePreferencesStore";
+import type { BattlefieldCardStyle, CardPreviewMode } from "@/stores/usePreferencesStore";
+import {
+  HOVER_DELAY_MAX,
+  HOVER_DELAY_MIN,
+  HOVER_DELAY_STEP,
+} from "@/components/game/game.constants";
+
+const CARD_STYLES: { value: BattlefieldCardStyle; label: string }[] = [
+  { value: "realistic", label: "Realistic" },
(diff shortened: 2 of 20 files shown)


CyberJaco (dismissed):
There was a gated check on card sizing that would calculate sizes in order to ensure 3 rows of cards at all times.
For bigger settings, we could reduce this to 2 rows. However, 1 row  makes the game truly unplayable and I don't think it should be allowed to get cards to be that big.

khaliostr:
Sure, np — done in 1113191fa: `BATTLEFIELD_MIN_ROWS_LARGEST` is now 2, so the size slider saturates at a 2-row fill per field and a 1-row board is unreachable. Hints/comments updated and the 300% screenshot in the description regenerated with the cap in place; `cargo xtask e2e-ui` still passes.

anacletos (approved):

PR #453, opened 2026-07-14
fix(engine): use protocol v1 ironsmith bridge

Chiplis:
## Summary

- consume Ironsmith's protocol-v1 `manabrewView(viewer)` result directly
- relay numeric prompt IDs and submit responses through `manabrewRespond`
- apply directives through `manabrewApplyDirective` and forward protocol errors to local and remote seats
- support protocol-v1's generic reorder prompt in the trusted-runtime bot fallback

## Why

This is a companion to #445 and should merge after it. Ironsmith's protocol-v1 adapter owns prompt IDs and response bindings, while the current trusted runtime still allocates string prompt IDs and translates prompt outputs on Manabrew's side.

The PR targets `main` so the review remains a focused one-file diff; #445's head predates `ironsmithRuntime.ts` and is currently conflicting with newer `main`.

## Test plan

- [x] `yarn eslint src/game/ironsmithRuntime.ts`
- [x] `yarn prettier --check src/game/ironsmithRuntime.ts`
- [x] `yarn tsc -p tsconfig.app.json --noEmit`
- [x] frontend portion of `yarn lint:all`
- [ ] full `yarn lint:all` (workspace Clippy stops on the pre-existing `unused_mut` in `forge-foundation/src/phase.rs:147`; #445 removes that code)

### DIFF
--- a/src/game/ironsmithRuntime.ts
+++ b/src/game/ironsmithRuntime.ts
@@ -18,27 +18,24 @@ import type { RoomMessagePayload } from "@/types/server";
 let ironsmithInit: Promise<unknown> | null = null;
 const IRONSMITH_RELAY_PROTOCOL = "ironsmith-trusted";
 
-interface IronsmithPromptBinding {
-  promptId: string;
-  playerSlot: string;
-  decisionKind: string;
-  actionRefs: Record<string, unknown>;
-  targetKinds: Record<string, "player" | "object" | "planeswalker" | "battle">;
-  optionIndices: Record<string, number>;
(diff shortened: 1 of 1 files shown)


khaliostr:
Hey @Chiplis ironsmith is now merged on main and available in prod!

it can be toggled via the settings

image

and then allows to host a ironsmith room

Screenshot 2026-07-22 at 12 09 10

not all decks work quite yet but it does return the unsupported cards

image



if you have a few known good decklist to share we can add them as presets!

closing this as done

Issue #465, opened 2026-07-15
Double-faced cards: fix at the root (DeckCard back-face representation + engine MDFC/DFC-land handling)


khaliostr:
## Problem

Manabrew's `DeckCard` representation carries **no back-face data** — only front-face fields plus `layout` and `isDoubleFaced` (`src/protocol/deck`). The front-face name is stored deliberately, because the engine indexes DFCs by front face (`frontFaceName`, `src/lib/scryfall.utils.ts`). As a result, **everything double-faced depends on live Scryfall metadata**: the board resolves the back image/text from the Scryfall cache at render time (`CardSprite` → `getCardTexture(..., faceIndex: 1)`), and there is no back-face data on the wire (`CardDto` carries only `isDoubleFaced`/`isTransformed` booleans — no `backName`, no `faces`).

This is a workaround, not a representation. It breaks down on a cold cache / offline, and it pushes face knowledge to whoever happens to have queried Scryfall.

## Context

PR #454 added a **transport** fix for the Ironsmith runtime specifically (commit db31cfb0f): before starting a match it enriches double-faced deck cards from the Scryfall cache and attaches both `card_faces`, because Ironsmith only builds a linked two-face definition when the deck entry carries both faces. That fixes the symptom for Ironsmith, but the underlying gap — our own repr lacking the data — remains.

## Root fix (two layers)

**1. Frontend / protocol — `DeckCard` carries the back face.**
- Add a back-face block (or a `faces` array) to `DeckCard`: back name, oracle text, mana cost, type line, image uris.
- Populate it **at deck import** from Scryfall (`scryfallToDeckCard`), so it's part of the deck, not a live lookup.
- Switch back-face rendering (`CardSprite`, `CardDetailModal`, previews) and the Ironsmith boundary off live-cache lookups onto the stored faces.
- Keep the front-name index intact (`frontFaceName`) — this is additive.

**2. Engine — double-faced / MDFC land handling at the root.**
- `CardDto` carries no face data; the Rust carddb indexes by front face. Double-faced **lands** (MDFC lands, back-face lands) and transform cards with different costs create real gameplay issues.
- Fix the mechanic at the root in the engine (carddb / face handling), per the parity "fix the rule, not the card" directive — not a per-card patch.

## Related

- Transport fix: #454 (commit db31cfb0f).
- #328 — transform commander with different casting costs: engine "can't decide" when mana is available for either face.
- #219 — tokens and double-faced not rendering on the stack.
[...]

PR #468, opened 2026-07-15
feat(deck): carry double-faced back faces in the deck representation (#465 pt.1)

khaliostr:
deckcards only kept the front face, so anything double-faced pulled the back off a live scryfall lookup at render time. this stores the back on the deck instead: optional `back_face` on `CardRulesSummary`, filled at import from `card_faces[1]`, read when rendering the back (live fallback for old decks). presets re-enriched too — 59 dfc cards across 23 decks. also fixes the enrich script's scryfall 400 (missing user-agent).

all additive/optional so old decks still load. `deck_dto.rs` isn't touched by #445, so no conflict — but it should rebase onto v1 so the field rides that version bump.

part 1 of #465. no `CardDto` change, the ui joins back via `asDeckCard`.

checked the deck persist path: `backFace` is additive+optional so old saved/hub decks still load, `normalizeDeck` copies cards by reference so it's preserved, and zustand-persist round-trips it. no `migrateDeck` entry needed — that's for sync structural reshapes, and the back face can only come from a scryfall fetch. old saved decks render via the live fallback until re-imported.


khaliostr:
note on deck persistence: `backFace` is additive+optional so old saved/hub decks still load, `normalizeDeck` copies cards by reference so it's preserved, and zustand-persist round-trips it. no `migrateDeck` entry needed (that's for sync structural reshapes, and the back face can only come from a scryfall fetch). old saved decks render via the live fallback until re-imported.

PR #474, opened 2026-07-16
fix(staging): pin origin to HTTPS + hard-reset in deploy-staging.sh

fedepoi:
## Summary

The staging VM's repo checkout was cloned over **SSH** (`git@github.com:…`), so `deploy-staging.sh`'s `git pull` failed with `git@github.com: Permission denied (publickey)` — the box has no GitHub SSH key. The repo is public, so no auth is needed at all over HTTPS.

`deploy-staging.sh` now, before fetching:
- pins `origin` to `https://github.com/<slug>.git` (anonymous), or the token URL if `GITHUB_TOKEN` is set — so an SSH-cloned box works with no key on it;
- `git fetch origin <branch>` + `git checkout -f -B <branch> FETCH_HEAD` to hard-reset the local branch to the remote, which also survives a diverged/detached/dirty checkout (the old `pull --ff-only` would bail).

## Why

This is the "auto-update" robustness the box needs: match the remote exactly on every deploy regardless of clone method or local drift, without requiring a GitHub key or token on a public repo.

## Test plan

- `bash -n deploy-staging.sh` — passes.
- Manual bootstrap on the box (`git remote set-url origin https://github.com/witchesofthehill/manabrew.git`) unblocks the current checkout; after this change lands, the script keeps origin on HTTPS itself.

### DIFF
--- a/deploy-staging.sh
+++ b/deploy-staging.sh
@@ -50,10 +50,23 @@ if [ -n "${GITHUB_TOKEN:-}" ]; then
 else
     git remote set-url origin "https://github.com/${REPO_SLUG}.git"
 fi
-PREV=$(git rev-parse --short HEAD)
+# DEPLOY_STAGING_ORIG_PREV preserves the pre-fetch commit across the self-update
+# re-exec below, so the re-run still reports the right range instead of an empty
+# "no new commits" changelog.
+PREV="${DEPLOY_STAGING_ORIG_PREV:-$(git rev-parse --short HEAD)}"
 git fetch origin "$BRANCH" >> "$RAW_LOG" 2>&1
 git checkout -f -B "$BRANCH" FETCH_HEAD >> "$RAW_LOG" 2>&1
 CURR=$(git rev-parse --short HEAD)
+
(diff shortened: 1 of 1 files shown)


khaliostr:
Nice! we might want to add a provision like this for robustness


```
# Self-update: the pull may have changed this very script. deploy.sh is already
# loaded into memory, so without this the *old* logic runs and script edits only
# take effect the NEXT deploy — exactly how the Ironsmith submodule checkout was
# missed and the runtime shipped dark on its first deploy. Re-exec the updated
# script once (guarded so it can't loop).
if [ -z "${DEPLOY_ORIG_PREV:-}" ] && [ "$PREV" != "$CURR" ] \
   && ! git diff --quiet "${PREV}..${CURR}" -- deploy.sh; then
    echo "deploy.sh changed in this pull — re-exec'ing the updated script" >> "$RAW_LOG"
    export DEPLOY_ORIG_PREV="$PREV"
    exec bash "$0" "$@"
fi
```

khaliostr (approved):

Issue #360, opened 2026-07-06
relay: winner not attributed for Forge-hosted games (game_ended reason=host_ended, winner=null)


khaliostr:
Every Forge-hosted game today records `game_ended` with `reason: host_ended` and `winner: null`, while durations/players/decks record fine. The relay's winner peek (`replay.rs::observe`) watches for `gameOver`/`winnerId` inside `kind:"state"` envelopes — the Manabrew engine path emits that (verified in the #350 smoke test), but the Forge path signals game-over via the node's `RoomRelay` gameOver payload + `EndGame`, which the peek never inspects.

Fix options:
- teach the peek the `RoomRelay` gameOver payload shape, or
- have the node broadcast a final `kind:"state"` envelope carrying `gameOver`/`winnerId` before `EndGame` (probably cleaner — keeps the relay's parsing surface minimal).

Until fixed, the product dashboard's completion/winner stats undercount all node-hosted games.

PR #513, opened 2026-07-20
ci(staging): restore the staging slot topology on the prod box

khaliostr:
## Summary

- Restores the staging slot topology: staging runs as a second compose project on the prod box (`/opt/manabrew-staging`), joined to the prod docker network and fronted by the prod Caddy.
- Rewires `staging-deploy.yml` from Twingate-to-a-dedicated-VM to direct SSH plus a `.env` secret sync, and updates `deploy-staging.sh` to match the `-staging` service names.
- Routes the staging hub API same-origin under `/api/*`, passes hub auth env through compose, and caps slot resources so prod always wins.

## Why

`main` is internally inconsistent today. `ops/Caddyfile` (added by #508) proxies `staging.manabrew.app` to `manabrew-staging:80` and `relay-staging.manabrew.app` to `manabrew-server-staging:9443`, but `main`'s own `compose.staging.yml` names those services `manabrew` and `manabrew-server`. Only the slot topology creates the names the Caddyfile points at, so main's ops config as a set cannot work.

The slot is what is actually deployed. The deploy workflow triggers on pushes to `staging`, so the file that ever executes is the one on that branch, and #508 was a production hotfix acknowledging the slot is live. This PR makes main match reality.

Landing this does not touch the production rollout. `compose.production.yml` gains only additive auth env, and `compose.staging.yml`, `staging.Caddyfile`, `deploy-staging.sh` and `staging-deploy.yml` are not read by the production deploy.

Two deliberate exclusions:

- `ops/manifest.json` stays at main's `version: 42` / `manabrew 1.12.4`. That is release-bot state and must not move backwards.
- The node stays at `SELF_HOSTED_NODE_MAX_GAMES=1` and `mem_limit: 1g`. Raising it to 4 games per JVM depends on the endstep concurrency patches, which ship separately. That flip is a follow-up once both have landed.

Known gap, filed as a follow-up rather than fixed here: the slot drops the events/captures observability config that main's standalone `compose.staging.yml` carries, so telemetry regressions of the #360 kind stop being catchable on staging. Restoring it means a second capture directory on the prod box disk, which is a resourcing decision worth taking on its own.

## Test plan

- `docker compose -f compose.staging.yml config` parses.
- Verified the service names now match the `reverse_proxy` targets in `ops/Caddyfile` (`manabrew-staging:80`, `manabrew-server-staging:9443`).
- Confirmed `ops/manifest.json` is unchanged from `main`.
- Diffed the result against `origin/staging`: `staging-deploy.yml`, `ops/staging.Caddyfile` and `deploy-staging.sh` are byte-identical to what is deployed. `compose.staging.yml` differs only by the intentionally-omitted endstep env block.
- No application code touched, so no typecheck or test run applies.

### DIFF
--- a/.github/workflows/staging-deploy.yml
+++ b/.github/workflows/staging-deploy.yml
@@ -1,8 +1,16 @@
 name: Staging deploy
 
-# Deploys the `staging` branch to staging.manabrew.app — a staging slot on the
-# PRODUCTION box (/opt/manabrew-staging, compose.staging.yml on the prod docker
-# network, fronted by prod Caddy's staging.manabrew.app / relay-staging vhosts):
+# Label-gated: a PR labelled `deploy-preview` takes the staging slot, the way
+# per-PR previews worked before #470 turned this into a branch auto-deploy.
+# Pushes to `staging` no longer deploy on their own — run this workflow from the
+# Actions tab (workflow_dispatch) to put the `staging` branch back on the slot.
+# There is ONE slot, so labelling a second PR replaces whatever is on it; the
+# slot is never torn down, it just holds the last thing deployed. Only the
+# `staging` branch gets a hosted-AI node (see deploy-staging.sh).
+#
+# Deploys to staging.manabrew.app — a staging slot on the PRODUCTION box
+# (/opt/manabrew-staging, compose.staging.yml on the prod docker network,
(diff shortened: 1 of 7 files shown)


khaliostr:
superseded by #524

PR #514, opened 2026-07-20
feat(engine): double-faced deck faces and endstep concurrency patches

khaliostr:
## Summary

- Carries double-faced back faces in the deck representation so both faces reach the engine at game start (#465 pt.1, #468).
- Gates playing a double-faced back-land on the card being modal (#469).
- Adds the endstep concurrency patches that let one Forge JVM host several games (#493).

## Why

Double-faced cards lost their back face in the deck representation, so the engine started games without it. The back-land fix is the matching rules gate: a back face that is a land is only playable when the card is modal.

The endstep patches are the engine-side prerequisite for running more than one game per Forge JVM. Evaluation on `scratch/endstep-patch-test` came back green at 40 games in 3.5GB.

The compose change that raises the staging node to 4 games per JVM is deliberately not in this PR. It belongs with the slot topology and only makes sense once both have landed.

## Test plan

- `yarn typecheck` clean.
- `cargo test -p manabrew-engine --lib`: 333 passed, 0 failed. This is the same gate `build-checks.yml` runs.
- `cargo check -p self-hosted-node` clean, verified against an initialized `forge` submodule since that crate `include_str!`s from it.
- `cargo check -p manabrew-engine -p manabrew_game_runtime` clean.
- The `forge` submodule bump (`6ab8389` to `1b900c6`) is carried unchanged from the original commit.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -10,3 +10,7 @@ forge/
 
 # release-please owns this file; reformatting it churns every release PR.
 CHANGELOG.md
+
+# gen-protocol-doc-types.mjs owns this file; CI diffs it against the raw
+# generator output, so a prettier pass makes the check permanently stale.
+website/src/generated/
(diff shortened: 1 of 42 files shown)


khaliostr:
Split into #520 (double-faced deck faces) and #521 (endstep concurrency). The two share no files, so they are better reviewed apart.

PR #515, opened 2026-07-20
build(ironsmith): consume ironsmith-wasm from npm, drop the submodule

khaliostr:
## Summary

- Consumes `ironsmith-wasm` from npm and drops the git submodule, along with the `sync-ironsmith-wasm.mjs` bridge and the WASM build stage in `Dockerfile.web`.
- Enriches double-faced deck cards with both faces before a game starts.
- Bumps the submodule to the protocol-v1 bridge in the step before the removal, so the history stays bisectable.

## Why

The submodule made every clone and every image build depend on a second repo and a local WASM build. `ironsmith-wasm` is published on npm (0.1.0), so the dependency can be an ordinary package. This removes the submodule from `.gitmodules`, the sync script, the Docker build stage, and the `deploy.sh` plumbing.

## Test plan

- `yarn typecheck` clean.
- Confirmed `ironsmith-wasm@0.1.0` resolves from the public registry (`npm view ironsmith-wasm version`).
- Confirmed no `ironsmith` entry remains in `.gitmodules`.
- Net effect is 258 deletions against 64 insertions across 15 files.


anacletos (approved):
lgtm

khaliostr:
@Chiplis, merging this!!

PR #517, opened 2026-07-20
feat(hub): accounts, oauth and magic-link auth, deck ownership

khaliostr:
## Summary

- Adds user accounts to the hub: GitHub and Discord OAuth plus email magic-link codes, in `manabrew-hub/src/auth.rs`.
- Adds `accounts` / `identities` / `sessions` tables to the existing `hub.db`, with tokens stored as sha256 hashes and all sign-in flows converging on a one-time code exchanged at `POST /api/auth/exchange`.
- Gates deck publishing on a bearer session. Deck rows carry `account_id` and the author is forced to the account handle. The per-deck sha256 management token stays as the delete fallback for pre-account decks.
- Adds the sign-in dialog, handle dialog, account section in Settings, and the `/auth/callback` view.
- The whole auth surface ships dark behind the new `accounts` feature flag.
- Rewrites the docs privacy page to describe what actually ships now: public-relay game records, optional accounts (OAuth providers, email codes through Resend), account-owned Deck Hub publishing, and the browser-side Scryfall image proxy.

## Why

Deck publishing had no notion of an owner. Anyone holding the per-deck management token could delete a deck, and authorship was self-declared. Accounts make ownership real and are the prerequisite for the top-decks board.

## Resolved blockers

Earlier drafts listed two blockers. Both are fixed.

**1. Magic-link codes were logged in plaintext.** When `RESEND_API_KEY` was unset, the request handler logged the code and link at `info` level, a complete sign-in credential flowing to stdout and Loki. The log line is gone. `POST /api/auth/email/request` now returns `503` before minting a code when sending is disabled, `providers_handler` reports `email` from `resend_api_key.is_some()` instead of hardcoding `true`, and the sign-in dialog hides the email option when the server reports it off.

**2. The auth surface was not feature-flagged.** New `accounts` flag in `featureFlags.ts`, default `false`. It gates the sidebar sign-in entry, the Settings account tab, the save nudge in the deck builder, the `/auth/callback` route, and the dialog mount in `AppShell`. Merging ships nothing user-visible. Flipping the flag is the rollout step, after the prod OAuth apps and the Resend sending domain are registered.

## Smaller things, now addressed

- The auth rate limiter is split in two. `/email/request` gets its own per-IP limiter (`HUB_AUTH_EMAILS_PER_HOUR`, default 30), and `/exchange` plus `/email/verify` share a code-attempt limiter (`HUB_AUTH_ATTEMPTS_PER_HOUR`, default 60). A shared office IP no longer starves on 5 requests an hour, and the per-email cap of 3 codes an hour still bounds spam per victim.
- `reqwest` now uses `rustls-tls`. The TLS stack no longer depends on libcurl dragging libssl3 into the runtime image, and the webpki roots are bundled.
- The localStorage question in `useAuthStore` is answered in place: the token persists deliberately, staying signed in is the product behavior, an XSS that could read it could call the API directly anyway, and sign-out revokes server-side where only the sha256 is stored.
- CORS no longer allows any origin. `allow_origin` is now the `WEB_APP_URL` origin plus the two Tauri webview origins (`tauri://localhost`, `http://tauri.localhost`).

## Test plan

- `yarn typecheck` clean.
- `cargo check -p manabrew-hub` clean.
- `prettier --check` clean.
- Website build (`npm run build`) passes with the rewritten privacy page.
- The `hub_decks.account_id` migration is guarded by `ensure_column`.
- Login codes have a 5-attempt lockout plus IP limiting, and OAuth `return_to` is resolved server-side so there is no open redirect.
- With `RESEND_API_KEY` unset: `/api/auth/providers` reports `email: false`, the dialog offers OAuth only, and a direct `/email/request` returns 503 without writing a login token.
- With the `accounts` flag off (the default): no sign-in entry in the sidebar, no account tab, no nudge, `/auth/callback` does not route.
- Exercised end to end in staging.


khaliostr:
Addressed the four smaller items too: split rate limiters (30/hr email sends, 60/hr code attempts, both env-tunable), reqwest on rustls-tls, the localStorage decision documented in the store, and CORS restricted to the WEB_APP_URL origin plus the Tauri webview origins. Hub tests all pass.

khaliostr:
Note on rollout order: the privacy page in this PR is not gated by the `accounts` flag. Docs deploy with the release, so between merging this and flipping the flag the page will describe sign-in and account-owned publishing that users can't see yet.

I think that's fine. The window should be short, and a privacy page that discloses early is better than one that discloses late. If we want them in sync instead, the accounts and Deck Hub wording can move to the flag-flip PR.

khaliostr:
@anacletos this is ready for review and merge. @fedepoi #549 will merge cleanly once this is in.

khaliostr:
Split auth.rs into a module in 22b7418: shared session and account plumbing in auth/mod.rs, providers in oauth.rs and email.rs, account endpoints in account.rs. Purely mechanical, no behavior or visibility changes, routes untouched.

I kept sessions with auth on purpose. Every sign-in path ends in the same account resolution and session mint, so separating them would only move the coupling behind an interface. Happy to revisit if the module keeps growing.

khaliostr:
Re signing authority: there is none in this PR, by design. Sessions are opaque random tokens stored as sha256 hashes; only the hub can verify them, by DB lookup. Single service, instant revocation, no keys to manage.

Upgrade path when another service needs to verify sessions (relay, most likely): the hub becomes the signing authority. Add a token exchange endpoint that turns a valid bearer session into a short-lived signed assertion (ed25519, claims: account_id, handle, exp). The relay verifies with the hub's public key. No shared DB, no per-connection hub call. Purely additive: the session model, storage, and every endpoint in this PR stay as they are.

khaliostr:
draft of JWT and signing authority work is at https://github.com/witchesofthehill/manabrew/pull/554

khaliostr:
Ready for review. Context and the full plan: https://github.com/witchesofthehill/manabrew/pull/573#pullrequestreview-4842235905

What just landed on the branch:

- merged main (clean), so the branch is current
- `16db818`: three small things pulled forward from #573 so this merges self-contained
  - `hub.ts` and `hydrate()` only touch the stored token when the accounts flag is on. Before, a token left in localStorage (dev builds and staging enable the flags) kept calling api.manabrew.app from flags-off builds. This was the real leak in the 517 surface, everything else was already gated
  - email.rs returns 502 and deletes the login code when Resend fails, instead of a silent OK the user waits on
  - auth dialogs say Community instead of naming a section that doesn't exist yet

Also audited every auth and hub mount point on the merged branch with both flags off: nothing renders, nothing fetches. The migrations are already in SQL here (1_schema.sql plus the build.rs runner), so #573's chain stacks on it directly.

Merging this dark also closes the ungated publish item in DeckBuilder that people are using on prod today. That's the deliberate call from the linked comment: publishing pauses until the flags flip.

khaliostr:
btw, I know this one got big too. 7.6k lines, and it picked up the docker dev env, the migration runner and the staging wiring along the way. Not going to split it though: it's one feature, it ships dark, #573 sits on top of it, and the three of us wrote pieces of it already.

Reading order that makes it manageable:

1. `migrations/1_schema.sql`, the whole data model
2. hub `src/auth/`: read `token.rs` and `oauth.rs` slowly, `email.rs` for the code flow and rate limits, `account.rs` is short
3. `routes.rs`, check what's public vs session-gated
4. `src/featureFlags.ts`, the runtime enable mechanism
5. UI: AppShell, TopBar, Settings, SignInDialog, useAuthStore, the bearer in `api/hub.ts`
6. docker, compose, yarn.lock: skim

I'd suggest to spend the time on token hashing, session lifetime and the oauth callback. And try to find UI I missed with the flags off. By my reckoning there's none.

let's get this in so #573 can proceed easily

anacletos (approved):

PR #520, opened 2026-07-20
feat(deck): carry double-faced back faces into the engine

khaliostr:
## Summary

- Gates playing a double-faced back-land on the card being modal (#469).
- Carries double-faced back faces in the deck representation so both faces reach the engine at game start (#465 pt.1, #468), and re-enriches the preset decks accordingly.

## Why

Double-faced cards lost their back face in the deck representation, so the engine started games without it. The back-land gate is the matching rules constraint: a back face that is a land is only playable when the card is modal.

Split out of #514 so the deck-representation work and the endstep concurrency patches can be reviewed separately. The two share no files.

## Test plan

- `yarn typecheck` clean.
- `cargo test -p manabrew-engine --lib`: 333 passed, 0 failed.
- Generated protocol types match the committed `protocol-types.json`, no drift.


manabrew-rs/crates/manabrew-protocol/src/deck_dto.rs
```diff
@@ -55,6 +55,31 @@ pub struct CardRulesSummary {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[ts(optional)]
     pub is_double_faced: Option<bool>,
+    /// The back face of a double-faced card (transform / modal_dfc), captured at
+    /// deck import from Scryfall so rendering never needs a live lookup. The
+    /// front-face fields above remain the card's identity; this is purely
+    /// additive. `None` for single-faced cards and older decks.
```
anacletos: These comments auto-go in the docs they should either be removed or de-slopified

manabrew-rs/crates/manabrew-protocol/src/deck_dto.rs
```diff
@@ -55,6 +55,31 @@ pub struct CardRulesSummary {
...
+    pub name: String,
+    #[serde(default)]
+    pub mana_cost: String,
+    #[serde(default)]
+    pub type_line: String,
+    #[serde(default)]
+    pub oracle_text: String,
+    #[serde(default)]
+    pub uris: CardImageUris,
 }
```
anacletos: Comment says One face of a double faced.
But the DeckCard only uses this on backface?
Do single face cards use `CardFaceSummary`
I think the interface is a bit confused.

manabrew-rs/crates/manabrew-protocol/src/deck_dto.rs
```diff
@@ -55,6 +55,31 @@ pub struct CardRulesSummary {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[ts(optional)]
     pub is_double_faced: Option<bool>,
```
anacletos, line 57: This one becomes redundant with the existence of:
```
back_face: Option<CardFaceSummary>,
```

anacletos (dismissed):
Couple of comments.

I also don't see a new migration for the saved decks.

When changing `DeckDto` the frontend needs a new versioned transform.
Otherwise, all the decks saved in localstorage on the old Dto will start crashing the app.

khaliostr:
thx will look into this

manabrew-rs/crates/manabrew-protocol/src/deck_dto.rs
khaliostr: Trimmed to one factual line on the field. The struct comment is gone.

manabrew-rs/crates/manabrew-protocol/src/deck_dto.rs
khaliostr: Renamed to CardBackFaceSummary. It only ever carries the back face. Single-faced cards carry nothing, and the front face stays as the flattened fields on the card, same as before this PR.

manabrew-rs/crates/manabrew-protocol/src/deck_dto.rs
khaliostr: It is redundant for new imports (both come from the same two-image gate), but the field predates this PR and decks saved without backFace still rely on it in the editor and limited views. backFace cannot be backfilled offline since it needs the Scryfall image URLs, so dropping isDoubleFaced would regress those decks. I would rather deprecate it in a follow-up once old decks have aged out.

khaliostr:
On the migration: the change is additive, so there is no transform for a new migration to apply. backFace is optional on both sides (serde default on the Rust struct, optional in the generated TS), an old deck deserializes unchanged with backFace absent, and getCardTexture falls back to the live Scryfall lookup exactly as before when it is missing. Nothing in the stored shape moved or was renamed. The 0.2.0 migration existed because the identity fields physically moved. migrateDeck still restamps old decks to the current version on load, so they will not re-enter the migration path. If you see a concrete input that breaks I will add the transform.

khaliostr:
Follow-up on the localstorage decks: pushed the backfill. The editor auto-enrich (the same gate that backfilled allParts) now treats a double-faced card without backFace as needing enrichment, so legacy decks pick up the back face from Scryfall the first time they are opened. Until then getCardTexture falls back to the live lookup, so nothing renders broken in the meantime.

manabrew-rs/crates/manabrew-protocol/src/deck_dto.rs
anacletos: But theres nothing in the code that does that.
The stale decks are gonna be stale for ever. They are never gonna age out.
The PR introduces a "hack" at the lowest level of the stack (`lib` / `editor`) instead of applying a migration so that the decks everywhere actually have a `DeckDto` shape. As of know, old decks will stay old. As soon as another migration comes in, this becomes officially a mess.

khaliostr: Fair. The editor-only backfill meant a deck that was only ever played would stay stale forever. Reworked it as a registered 1.1.0 deck migration (d35e719): the sync half can't exist, backFace needs Scryfall data that isn't on the client, so the migration's transform is an async backfill that runs from the store's hydration hook, once per app start, over every saved deck. It selects work by shape instead of version stamp so an offline load retries on the next start instead of stranding the deck. Keyed 1.1.0 because migrateDeck restamps loaded decks to the protocol VERSION (already 1.0.0), so a lower key would never fire.

Once this ships every client converges on first launch, and isDoubleFaced can be dropped in a follow-up. The getCardTexture fallback stays only as the render safety net for the window before the sweep lands.

khaliostr: Dropped the backstop (e8a74c7). Back faces now read strictly from backFace.uris; a card that reaches the board without it renders an empty texture instead of triggering a live lookup. The migration is the only compat path. The face-0 lookup stays, it predates this PR and covers cards whose stored uris lack the requested crop.

JacopoMadaluni (approved):

PR #521, opened 2026-07-20
feat(engine): endstep concurrency patches for multi-game forge

khaliostr:
## Summary

- Adds the endstep concurrency patches that let one Forge JVM host several games (#493).
- Bumps the `forge` submodule from `6ab8389` to `1b900c6`, which is the endstep-patched branch.
- Reworks `java_backend.rs` around the reset path and updates the hosted smoke test.

## Why

Running one JVM per game is the current ceiling on the hosted-AI node. These are the engine-side patches that make a shared JVM viable. Evaluation on `scratch/endstep-patch-test` came back green at 40 games in 3.5GB.

Split out of #514 so it can be reviewed on its own. It shares no files with the double-faced work in #520.

The compose change that raises the staging node to 4 games per JVM is deliberately not here. It belongs with the slot topology in #513, which stays at `MAX_GAMES=1` so the slot cannot enable multi-game before this lands. Flipping it is a follow-up once both are in.

## Note for reviewers

This PR does not build against `main`'s forge submodule, by design. The Java harness changes require the bumped revision, so a stale submodule fails with API mismatches (`CostPayShards` in `ICostVisitor`, `Multiset.Entry` vs `Map.Entry` for counters, `refreshSkin()` on `IGuiBase`). If you see those, run `git submodule update --init --recursive` rather than reading them as defects in this diff.

## Test plan

- `yarn build:harness` BUILD SUCCESS against forge `1b900c6`, which is what this PR pins. Verified the same build fails against `6ab8389`, confirming the submodule bump is load-bearing rather than incidental.
- `cargo check -p self-hosted-node` clean.
- `cargo test -p manabrew-engine --lib`: 333 passed, 0 failed.
- `yarn typecheck` clean.
- Not exercised as a live multi-game session on the node. That happens once #513 and this are both in and the slot is flipped to 4 games per JVM.

### DIFF
--- a/forge-harness/src/main/java/forge/harness/Main.java
+++ b/forge-harness/src/main/java/forge/harness/Main.java
@@ -278,8 +278,10 @@ public final class Main {
 
     /**
      * Interactive server mode: per-game JSON-RPC over stdin/stdout dispatching
-     * to a long-lived {@link ManaBrewEngineAdapter}. One process owns one game
-     * at a time; pool callers send `reset` between games and `quit` to exit.
+     * to a long-lived {@link ManaBrewEngineAdapter}. Sessions are routed by
+     * sessionId and each runs on its own game thread, so one process can host
+     * several concurrent games; pool callers send `reset` only while the
+     * process is idle and `quit` to exit.
      *
      * Request shape:  {"command":"startGame","payload":"<json>"}  (sessionId
      * goes in `sessionId`; payload is the inner JSON string).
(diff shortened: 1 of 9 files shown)


anacletos (commented):
It looks good but Im a bit worried of the resets especially cos we use reflection to reset them.

Was this well tested? Perhaps warrants a small regression test that ensure the room survives it?
Also does graavm work? It has known issues with reflection

khaliostr:
Yes it was super tested. 1200 games in JVM in batches of 40 sequential. no hiccups and big gains. graalvm was tested too and it's working very well but there is a bit of trouble on serial GC slowing down on long concurrent session on macos. So not advised to run concurrent production on graalvm yet but stable

anacletos (approved):

PR #522, opened 2026-07-20
feat(protocol)!: add source card metadata to prompts

### LINKED ISSUE #237: Prompting - sourceCardId doesn't render preview on the UI prompt if card is in a hidden zone
If the UI gets prompted by a card in the library, which is not visible by the client, the UI will fail to render the preview.
This is a limitation of the protocol as it only sends a engine card id on the wire, so if the card isn't in a visible zone, the UI is unable to resolve the id onto a card.

The solution is to change the `sourceCardId: String` everywhere in the protocol to send the following:
```
type SourceCard { 
      engineId: string;
      lookup: "$cardName:$setCode:$cardNumber"
}
```

other solutions might exist to this problem, whatever it is, it has to be a generic protocol enhancement to allow the engine to send info about any card regardless of whether it's currently visible in the state

fedepoi:
## Summary

- replace prompt-level `sourceCardId` with a top-level `SourceCard` descriptor containing the engine id and optional printing identity
- populate source metadata in both Rust and Forge, then resolve hidden-zone source previews through visible state, deck printings, Scryfall, and the token archive
- route per-seat state, prompts, errors, and reconnect replays privately through `BroadcastState.target_player`

## Why

Hidden-zone source cards are intentionally absent from `GameViewDto`, so an engine-local id alone cannot resolve prompt artwork. Adding printing identity fixes the preview generically, while private relay targeting prevents that identity from leaking to other room members.

Fixes #237.

## Test plan

- `yarn lint`
- `yarn test` (9 tests passed)
- `yarn build:harness`
- `cargo check -p manabrew-agent-interface -p self-hosted-node -p manabrew-server`
- `yarn test:networking` (11 relay/node scenarios passed)
- `yarn gen:protocol-docs`
- `npm run build` in `website/`
- `git diff --check`
- `yarn lint:all` reaches the repository's documented cold-cache Clippy debt in unchanged code (`forge-limited`, `CardView`, and the local native Forge runtime prerequisite); frontend lint, formatting, typecheck, and targeted Rust checks pass

## Demo

No capture included: the changed UI appears only while answering a prompt sourced by a card outside visible game state. The source descriptor is covered by the generated prompt fixture, both engine producers compile against the new contract, and modal plus mana-payment render paths consume the resolved card.


manabrew-rs/crates/manabrew-protocol/src/prompts/common.rs
```diff
@@ -14,9 +24,6 @@ pub struct PromptPresentation {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[ts(optional)]
     pub text: Option<String>,
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    #[ts(optional)]
-    pub source_card_id: Option<String>,
     #[serde(default)]
     pub targets: Vec<TargetRef>,
```
anacletos: So the reason this is shaky is the following:

Whilst you *can now send a lookup to perform a preview, you still do not have the full `CardDto` at hand, meaning you're still gonna lack a lot of info for a lot of cases.

To resolve this issue, a lot of prompts give you the ability to inject `CardDto` inside the prompt for some specific purposes (exactly because the card might be hidden, hence not retrievable by id).

I think if this is needed (and I assume it is), I would put 
`sourceCard: CardDto` straight up, removing all edge cases

anacletos (changes requested):

manabrew-rs/crates/manabrew-protocol/src/prompts/common.rs
fedepoi: Addressed in 61b806d44. `AgentPrompt.sourceCard` is now `CardDto` directly; the standalone `SourceCard { engineId, lookup }` type has been removed. Rust emits the canonical `card_to_dto(...)` result, Forge emits `InteractiveSnapshotExtractor.cardDto(...)`, and the UI now consumes `sourceCard.id` / `sourceCard.identity` while retaining client-side artwork enrichment. I also updated the fixture, Java/TypeScript generators, and protocol docs. Targeted private prompt transport and reconnect replay remain unchanged.

anacletos (dismissed):

khaliostr (approved):

JacopoMadaluni (approved):

Issue #512, opened 2026-07-20
CI - Move deploy decisions into xtask, and consider running orchestration on the CI runner


khaliostr:
The deploy path is spread across `publish.yml`, `staging-deploy.yml`, `deploy.sh` (311 lines of code), `deploy-staging.sh` (95) and `deploy-local.sh` (28), with decisions in bash and state on the box. #510 came out of that: `publish.yml` never passed `MANABREW_IMAGE_TAG`, `deploy.sh` fell back to `:latest`, and releases briefly shipped the previous release's images without anything going red.

Two changes are worth considering, in order.

## 1. `xtask deploy-plan`

`deploy-gate` in `publish.yml` is about 25 lines of bash and `jq` that re-derive semver compatibility from `ops/manifest.json`. That file is written by `xtask` (`manifest.rs`). The gate reimplements "0.x breaks on minor" by hand with `IFS=. read` while `xtask` already depends on `semver = "1"`.

Move the decision into xtask and have it emit the whole plan:

```
xtask deploy-plan --from v1.12.3 --to v1.12.4
  -> {"image_tag":"v1.12.4","early":true,"services":["manabrew"],"hold_manifest":true}
```

`deploy.sh` stays bash and consumes the JSON. It becomes a docker compose driver, which is what it is good at, and it keeps working on a box with no Rust toolchain.

The specific payoff is the #510 bug class. A `DeployPlan` with a required `image_tag` field makes a missing tag a compile error. In bash there is no way to say "this is required" that survives someone forgetting to pass the env var, and the fallback to `:latest` was silent and green.

Scope: replaces `deploy-gate`, leaves the rollout scripts alone. Low risk.

## 2. Orchestration on the CI runner

Bigger, separate call. Today the box holds a git checkout and the deploy scripts run from it. That is the source of a whole class of drift:

- `deploy.sh` self-updates mid-run with `exec bash "$0"` because the pull can change the running script.
- The early deploy advances the box HEAD, so the final deploy hits "no new commits" and exits 0. Broken deploys reported green. #511 works around it with an unconditional `FORCE_DEPLOY`.
- `deploy-staging.sh` executes from whatever branch got checked out, so a staging-branch deploy runs a different copy of the script than a PR preview does. `compose.staging.yml` has already diverged between `main` and `staging` (different service names, node profile only on `main`).

Running the orchestration on the runner against the box through a Docker context removes the checkout, the self-update, and the branch-dependent behaviour. No code on the box.

[...]

Issue #510, opened 2026-07-19
CI - Broken release pipeline


anacletos:
Release pipeline is broken.
* It doesn't correclty release the right version, `deploy.sh` always pulls `:latest` and nothing in CI waits for the builds to be ready, hence potentially pulling the old one with a new manifest.
* The early release breaks production if the relay gets updated with a protocol change, because it pushes without bumping manifest.

<img width="865" height="475" alt="Image" src="https://github.com/user-attachments/assets/29a63c76-0f53-44a3-9e60-08cb94792383" />

PR #524, opened 2026-07-21
ci: run the production deploy from the runner via cargo xtask deploy

khaliostr:
Includes #513/#519: this branch merges `khaliostr/staging-slot-ops`, since the staging conversion builds on the slot topology. #513 closed in favour of this

## Summary

- Deletes `deploy.sh`, `deploy-staging.sh` and `deploy-local.sh`. `cargo xtask deploy` replaces them: `--tag vX.Y.Z` (production, publish.yml), `--staging [--branch B] [--ref SHA]` (the slot, staging-deploy.yml), `--local` (selfhost VM), plus `--gate`, `--web-only` and `--only <svc...>`.
- Deploys run from the runner. The box keeps docker, `.env` and the data dirs. No scripts, no git.
- Config is fetched at deploy time as a codeload tarball of the deployed ref, so config and images always come from the same commit. Tarballs contain only tracked files, so the sync cannot touch the data dirs. staging-deploy.yml pins the fetch to the sha build-images built; `--config-from <dir>` uses a local checkout instead.
- `/manifest.json` redirects to the latest Release's copy, attached as an asset like `tauri.json`. The manifest hold and `--release-manifest` are gone.
- Every release attaches `manabrew-xtask-linux-x86_64` (static musl, ~1.4MB). One download deploys or rolls back to any ref. Needs ssh, rsync, curl and tar.

## Why

#510: publish.yml never passed the image tag, so deploys briefly shipped the previous release's images. #512 traced it to decisions in bash and state on the box, and v1.12.5 confirmed two more of its predictions: the deploy.sh self-update re-exec fired, and the final deploy survived only on `FORCE_DEPLOY`. The remaining gap was deploy.sh pulling `main` while images came from the tag; #515 merged 34 seconds after the v1.12.5 tag and its config reached prod outside any release. Fetching config at the deployed ref closes all of it.

Two decisions to flag:

- `ops/deploy-tool-version`: the fetched config names the tool version it needs, and an older binary refuses with a download hint. Refs predating the marker warn and continue, so rollbacks to old releases stay open.
- Tarball bytes are not checksummed, because GitHub does not keep them stable. Pinning is done by fetching a sha.

Unchanged: the early-deploy flow (gate, early web deploy, relay and hub in the final deploy; the manifest cannot advance early because the redirect points at the latest published Release), the relay binary-sha guard, the staging slot semantics (hosted-AI node only on `staging`, re-tag rollback since `:staging` is mutable), and the Discord embeds.

Git remains only where CI compiles the tool, and in `--local`, which builds images from source.

## Test plan

- clippy clean (`-p xtask --no-deps`; without `--no-deps` it hits the known lint drift on main).
- Gate matches the bash gate on real history: `early=true` for v1.12.3/4/5, `false` for v1.12.0.
- On bare Alpine with only the static binary, no git, cargo or checkout: codeload fetch and busybox-tar extraction of v1.12.5, gate defaulting to the embedded tag, fetch by exact sha, tool-version refusal (v99) and pass (v1).
- `--staging` against a throwaway sshd container: fetched by sha, synced 21 config files, created no data dirs, failed at the image pull (no docker there).
- Generated remote scripts pass `sh -n`. Both workflows and all three compose files parse.
- Not run against the real box; the next release tag is the first live run. Rollback is `deploy --tag <previous>`.

## Rollout notes

- Until a release carries the `manifest.json` asset, the gate takes the safe deploy-last path and `/manifest.json` 404s (the node updater reads that as no update). Attach `ops/manifest.json` to the current release by hand after merging.
- The box's old checkout and `ops/.manifest-hold` are inert.
- `--local` needs the source checkout but no Rust toolchain; use the prebuilt binary.


khaliostr:
tested on staging, works perfectly fine. ready to merge

anacletos (approved):

PR #531, opened 2026-07-22
feat(ui): add Ironsmith-compatible preset decks

Chiplis:
## Summary

- Add 25 preset decks verified for the Ironsmith engine: 18 Modern, 5 Standard, 1 Vintage, and 1 Commander.
- Give Ironsmith its own preset manifest while registering the decks in the shared picker.
- Extend preset loading, expansion, enrichment, and auditing to preserve sideboards.

## Why

The Ironsmith runtime currently returns no preset decks, so players have no curated, known-compatible choices. Every card in these lists is strict-compiled in the Ironsmith status database with semantic similarity above 0.99; the minimum across 289 unique cards is 0.991030693.

## Test plan

- PASS: yarn prettier --check on all changed TypeScript and script files.
- PASS: frontend ESLint, Prettier, generated protocol types, and TypeScript checks from yarn lint:all.
- PASS: manifest audit confirms 25 decks, expected format counts, valid main-deck and sideboard sizes, and all IDs present in the shared picker index.
- PASS: Ironsmith database audit confirms 289 unique cards, 0 missing cards, 0 cards at or below 0.99, minimum score 0.991030693.
- NOTE: the Rust Clippy portion of yarn lint:all reports 52 existing warnings-as-errors in untouched engine files, including dig_until_effect.rs and manabrew-protocol/src/game/mod.rs.

## Demo

No layout or styling changes. Choose the Ironsmith engine and open the preset deck picker; the new Modern, Standard, Vintage, and Commander lists are available for selection.


khaliostr (dismissed):

khaliostr (commented):
Thanks for putting this together, the deck data is in really good shape. I checked all 25 lists: sizes are legal (60/80 + 15, commander exactly 100) and Lathril is extracted correctly. The sideboard plumbing is done properly end to end (type, expansion, enrichment, audit), and threading `section` through the audit occurrences keeps `--apply` correct. Moving the loader into `src/lib/presetDecks.ts` and dropping the worker's private `PresetDeck` interface is a nice cleanup too.

A few things I'd like to sort out before merging.

**The engine manifest has no reachable consumer right now.** Every picker (`CreateGameDialog`, `DeckVsSelector`, `useQuickPlaytest`, `DeckEditor`) plus the appInit cover prefetch goes through `usePresetDecks()`, which calls `getDefaultGameRuntime()`, and that always returns the manabrew runtime. Nothing calls `IronsmithTrustedGameApi.getPresetDecks()`. So the decks show up because of `index.json` alone, for every engine, and selecting the Ironsmith engine doesn't actually change which decks appear.

Given that, I'd suggest dropping `engines/ironsmith.json` and the `ironsmithRuntime.ts` hunk from this PR and keeping `getPresetDecks()` returning `[]` for now. When we wire per-engine pickers for real, I'd rather see a tag in the deck file itself (something like `"engines": ["ironsmith"]`) filtered at load time than a second ID list: single registry, no drift, and both loaders already fetch every deck body anyway. As it stands the manifest and `index.json` have to be kept in sync by hand, and the committed audit script doesn't know the manifest exists.

**Default-engine compatibility.** The lists are verified against Ironsmith, but `index.json` serves them to the default Rust engine picker too, and `deckValidations` only covers user decks. Did you get a chance to check these against manabrew's card coverage? If not, it might be safer to hold the index registration until we have, so players don't hit unimplemented cards vs the AI.

**Startup payload.** The index goes from 43 decks / 2.6 MB to 68 / 3.5 MB, all fetched eagerly at app init. Not a blocker, gzip helps, but it's 25 extra requests for every user. Maybe worth a follow-up to lazy-load deck bodies.

One small latent thing: in `expandPresetDeckDefinition` the commander extraction check now runs during the sideboard pass too. Harmless for these decks, but a commander listed only in the sideboard would get silently promoted. Restricting the check to the mainboard pass closes that.

The clippy failures you noted are the known lint drift on main, not from this PR.

khaliostr:
Merged into #605's branch (cb5e118d4, merge d42a6557a). All 25 decks are in. Changes on top:

- Preset JSONs can declare an `engines` list. Your decks are tagged `["Ironsmith"]`. Presets without the field stay available everywhere except Ironsmith, whose pool is verified per deck.
- Deck pickers filter presets by the engine they launch on. Offline quick play refuses Ironsmith-only presets and points to Multiplayer.
- The play home now lists your decks, starters and published Community decks in one table with engine marks.

This PR can close once #605 merges. Thanks for the verified lists.

khaliostr:
merged! should be live soon. thanks @Chiplis !!

PR #548, opened 2026-07-25
fix(release): keep the github release a draft until installers attach

khaliostr:
## Summary

- `cargo xtask release` now creates the GitHub Release with `--draft`.
- publish.yml's existing attach step (softprops, `draft: false`) finds the draft by tag name, uploads the installers plus `tauri.json` and `manifest.json`, and publishes it in the same call. No workflow logic changes, only comments.
- AGENTS.md release-flow sentence updated to match.

## Why

The release was created and published one second before publish.yml started. From that moment `releases/latest` pointed at a release with zero assets until the release job attached them 25 to 40 minutes later. Three consequences:

- The early-deploy gate curls `releases/latest/download/manifest.json` and got a 404 on every release since #524 landed. `early` has been false on all 8 releases since July 22, so the early web deploy silently never ran.
- The `/manifest.json` and `/tauri.json` Caddy redirects 404'd for installed clients during the same window. Today that window was ~3.5 hours while the macOS build was broken, with v2.1.0 publicly listed and empty.
- A failed publish left a published release with no installers. As a draft it stays invisible until the rerun completes.

With the draft flow, `latest` keeps pointing at the previous complete release for the whole build, so the gate compares against the manifest it was designed to compare against, and clients never see a release without installers.

## Test plan

- `cargo fmt --check -p xtask`, `actionlint`, `prettier --check` pass.
- softprops/action-gh-release resolves existing draft releases by tag name and `draft: false` publishes them (documented behavior; it updates rather than duplicating).
- Next release: the gate log should print a version comparison instead of "no published release manifest reachable", and `early=true` on a compatible release. The Release should appear in the UI only once installers are attached. `curl -I https://manabrew.app/manifest.json` during the build window should still resolve to the previous release's asset.


khaliostr:
merging as hotfix

PR #549, opened 2026-07-26
feat(ui): integrate deck hub across deck flows

fedepoi:
## Summary

- surface account-owned Deck Hub publications in My Decks and the Play Home deck shelf
- add lazy-loaded public Deck Hub choices to offline, multiplayer, and bot deck selectors with source-aware read-only handling
- remove superseded deck UI, inert TanStack Query infrastructure, stale account-save messaging, and unused store APIs

## Why

Deck Hub publications were isolated from the reworked deck and play flows, while account ownership was still inferred from browser-local management tokens. This makes server ownership canonical across devices, keeps remote snapshots distinct from editable local decks, and lets players use community decks directly throughout setup flows.

## Test plan

- `yarn lint`
- `yarn vite:build`
- `cargo fmt --all -- --check`
- `git diff --check`
- `yarn lint:all` reaches Rust Clippy after all frontend checks pass; Clippy remains blocked by existing warnings in untouched `forge-limited`, `manabrew-protocol`, and engine effect files

## Demo

No capture included. The Deck Hub and account surfaces are runtime-flagged and require the live Hub API plus an authenticated account for the owned-publication states.


khaliostr:
Poked around while it was on staging, three small things for later, nothing blocking:

- `useHubDeckSearch` runs even while the dialog is closed. Lobby mounts two `CreateGameDialog`s, so just entering the lobby fires two `/api/hub/decks` fetches and refetches on format changes. Gating the effect on `open` should do it.
- the 401 copy in `hubRequest` says "session expired, sign in again" even when there never was a token, e.g. a signed-out publish attempt. Worth branching the message on `token` so anonymous users get the old "sign in to publish" wording.
- the DEV override in `featureFlags.ts` also force-enables accounts/deckHub in tauri dev, where `getHubApiUrl()` still resolves to prod. The `/hub-api` proxy gate excludes tauri but the flag gate doesn't. It's also active under vitest, so flag-gated paths behave differently in tests.

Plus, pre-existing, the play action on the hub decks and on the precon/own decks differ. Notably the decks from the deck-hub allow playing 4p commander. We should probably normalise and allow own/precon that option too. Also if signed in there is discover decks from the community under my decks that can probably be done without.```

I would probably also default the starter decks chevron to be open (one less click), at least when there are no own/deck hub present.

One last thing, since we're touching decks, should we fold in https://github.com/witchesofthehill/manabrew/pull/531 and gate it to engine selection?

Rest of it looks good, the bearer-then-management-token fallback on delete is nice. I know it's WIP so don't worry too much!

PR #562, opened 2026-07-28
fix(ui): render active card faces on the stack

### LINKED ISSUE #219: Double faced cards not rendering on the stack

fedepoi:
## Summary

- preserve owner and active-face state for stack objects from both engine backends
- render and reorient cards from the displayed face across DOM and Pixi surfaces
- add stack flip controls and the configured F keyboard binding to the pile and stack modal

## Why

Double-faced stack entries were reconstructed without their active-face state, so abilities from transformed cards rendered the front face. Face orientation was also derived from the whole card instead of the displayed face, which mishandled cards such as Battles when switching between portrait and landscape faces.

Fixes #219

## Test plan

- TAURI_CONFIG with no-native development resources yarn lint:all
- yarn build:harness
- cargo check -p manabrew-protocol -p manabrew-agent-interface
- cargo clippy -p manabrew-protocol -p manabrew-agent-interface --lib --bins -- -D warnings
- yarn typecheck
- verify protocol generation and git diff --check

## Demo

No recording attached; this requires an interactive game state containing a double-faced stack source.


manabrew-rs/crates/manabrew-protocol/src/game/mod.rs
```diff
@@ -292,13 +292,59 @@ pub struct StackObjectDto {
     pub id: String,
     pub source_id: String,
     pub controller_id: String,
+    pub owner_id: String,
```
anacletos, line 295: whats `owner_id` for

fedepoi: It preserves the distinction between the card owner and the player controlling/casting the stack object. We reconstruct a `CardDto` from the stack entry when the source is no longer in a visible zone, and the renderers use `ownerId` to select the owning player’s deck for printing/image metadata. Using `controllerId` there would look in the wrong deck for effects that let a player cast an opponent’s card.

forge-harness/src/test/java/forge/harness/host/InteractiveSnapshotExtractorTest.java
```diff
@@ -0,0 +1,25 @@
+package forge.harness.host;
+
+import forge.card.CardStateName;
+
+public final class InteractiveSnapshotExtractorTest {
+    private InteractiveSnapshotExtractorTest() {}
```
anacletos, line 6: This test runs when?

anacletos (approved):

PR #573, opened 2026-07-31
feat(deckhub): add versioned decks and official presets

fedepoi:
## Summary

- add the normalized DeckHub v2 domain: account-owned decks, immutable versions and card rows, public publications, tags, favorites, and dated Top Deck snapshots
- sync the 43 shipped presets into the Hub as ownerless canonical decks with immutable official publications and a ranked Official Presets snapshot
- create one idempotent account-owned fork on the first signed-in preset use, preserve preset provenance, and expose the copy across My Decks, Play, publishing, and version history
- rework discovery and Top Decks with URL-backed filters, source/format/tag/color grouping, grid/list views, exact card snapshots, hover preview, play/save actions, and owner publication editing
- preserve legacy Hub endpoints during the rolling-client compatibility window while preventing stats-only Top Deck rows

## Why

The previous model coupled mutable decks, ownership, publication, and ranking. This PR gives each concern a stable boundary: canonical account decks remain editable, versions preserve publishable content, DeckHub entries carry public metadata, and dated snapshots provide deterministic rankings where every row opens a complete deck.

Official presets now use the same domain instead of living in a disconnected catalog. They remain immutable shared sources, while first use creates an exact private account copy that can be edited, versioned, played, and published without losing its origin.

## Test plan

- `yarn lint:all`
- `yarn vite:build`
- `yarn typecheck`
- `cargo check -p manabrew-hub --all-features`
- `git diff --check`
- run the final Hub binary against the migrated review database and verify capabilities, 43 preset publications, 99+1 commander fork contents, idempotent fork reuse, owner-only publication listing, 43 ranked entries, and `PRAGMA foreign_key_check`
- exercise signed-in discovery, Top Decks, exact-card preview, favorites, Add/Open My Copy, Play/My Decks shelves, publication editing, account-deck opening, and version history in the browser

## Demo

- verified Deck Hub at 1280×800 and iPhone 12 widths
- verified responsive discovery controls, source/format/tag/color filtering and grouping, grid/list results, and scrollbar-free mobile ranking tabs
- verified every discovery and Top Deck card opens the full card browser with search, filters, list/visual/stack views, and hover preview
- verified owners can edit or unpublish a publication, open the canonical account deck, inspect immutable version history, and jump between a preset-derived copy and its official Hub entry


fedepoi:
@anacleto @kaliostr Review follow-up — 4 backend points worth addressing before/around merge (full context in my review):

1. **Deleting an account deck leaves its publications live forever** (`storage.rs:981`, `delete_account_deck`) — the soft delete only sets `decks.deleted_at`; the deck's `deckhub_entries` stay `published`, remain listed (`list_deckhub_entries` has no `d.deleted_at IS NULL` filter), and keep serving the full deck JSON, while the deck disappears from the owner's list with no restore path. Either unlist the deck's entries in the same transaction (and mirror `hub_decks.unlisted = 1` + `update_deck_visibility`), or explicitly document "delete keeps publications" as the intended product behavior.

2. **Preset sync reverts admin moderation on every boot** (`storage.rs:670-691`, `sync_preset_decks`) — the `ON CONFLICT(id) DO UPDATE SET … status = 'published'` unconditionally republishes the current preset entry, so an `admin_unlist`/`admin_delete` (and any admin retagging, wiped by the per-boot `replace_entry_tags`) is silently undone at the next hub restart. Only set `status='published'` on insert, or add a moderation flag the sync respects.

3. **No per-account quota on decks/versions** — any authenticated user can create unbounded decks, and every `PATCH` with a changed hash appends an immutable version plus ~100 `deck_cards` rows forever; `list_owned_decks`/`list_deck_versions` are also unbounded (no `LIMIT`). Suggest capping decks per account and versions per deck (e.g. 200/100) with a 409/422 on overflow, and/or reusing the publish `RateLimiter` pattern.

4. **Smaller follow-ups** (cheap, can be a separate PR):
   - favorites don't migrate when a preset version bump mints a new entry id — favorited presets silently vanish from favorites-filtered lists;
   - mixed timestamp formats (`main.rs` `Utc::now().to_rfc3339()` vs routes' `SecondsFormat::Secs`/`Z`) break lexicographic ordering/comparison of same-second rows;
   - `title`/`summary`/version `notes` lack control-char rejection and `notes` has no length cap;
   - listing endpoints parse every full `snapshot_json` just for a cover URL (plus an N+1 in `get_top_deck_snapshot`) — persist `cover_image_url` on `deckhub_entries` at publish/sync time;
   - five near-identical version-insert blocks in `storage.rs` (~80 lines) that will drift — extract one `insert_version` helper.

khaliostr (dismissed):
Notes from Discord plus my own review.

**On the v0 decks.** Found migrate_legacy_hub_decks in storage.rs. GitHub wasn't rendering the diff so I missed it. One question: it fails the whole migration on the first deck_json that doesn't parse, so one stale row and the hub doesn't boot. I'd skip and log bad rows.

**Leaks:** the prod leak is DeckBuilder's ungated publish item, on main. 517 closes it. I think the leak you meant on 517 is the stored token: hydrate() and api/hub.ts use it without checking the flag, and your hub.ts fix covers half of it. Cherry-pick, not blocker. One decision though: merging 517 dark disables publishing for the people using it on prod today. Fine by me, but let's choose it, not ship it by accident.

**Order is forced:** migration 2 alters tables that only exist once 517 is in. And 517 stands alone: account-wise this PR only adds the email.rs fix, the Community wording and additive columns.

**No fold into 517:** same 11k unreviewed lines, now blocking accounts too, and one giant release we can't debug or roll back.

**Proposed plan:**

1. Merge 517 plus three cherry-picks: email.rs fix, Community wording, token gate. I'll browser-test it flags-off first.
2. Restack this on main: schema + API with the backfill and old endpoints, then presets and forks, then UI. Relay, harness and card preview changes go separate, that code is stable.
3. Tags, favorites and rankings wait until the flag is on.

khaliostr:
@fedepoi further question now that I've played a bit on staging.

Given the decks are now normalised in the db and the presets live in the Hub as community decks too, should we rework the main UI to present our decks, the community decks and the preset decks in the same table? Arguably the presets are now hidden from the main deck view, while inside community they're very visible as presets.

Pointing this out now but could arguably be done as a follow up UI pr.

khaliostr:
after discord chat, ok to merge on condition that #517 lands first as is tested in isolation. item 2 is exercised on staging. @anacletos if you can chime in let us know if you see residual risk or could go in as is.

anacletos (approved):

PR #581, opened 2026-08-01
feat: autopass

### LINKED ISSUE #317: have the ability to pass the whole stack
when playing if the stack gets too big or something comboes out it would be really nice to have the ability to pass the whole stack or pass until something else changes. 

also worth exploring how to visualise huge stacks since as of now it keeps expanding horizontally

### LINKED ISSUE #580: Autopass

anacletos:
## Summary

* Add support full control and autopass mode
* Add "Resolve stack" option to stop going through each stack entry one by one

Closes #317 
Closes #580 

Screenshot 2026-08-01 at 20 46 13
Screenshot 2026-08-01 at 20 46 20
Screenshot 2026-08-01 at 20 45 55
Screenshot 2026-08-01 at 20 41 16
Screenshot 2026-08-01 at 21 27 21


src/components/prompts/ChooseAction.tsx
```diff
@@ -1,43 +1,172 @@
...
-export function ChooseAction({ isWaitingForResponse, onPassPriority }: ChooseActionProps) {
+function AutopassFill({ onDone }: { onDone: () => void }) {
+  const fillRef = useRef<HTMLDivElement>(null);
+  const onDoneRef = useRef(onDone);
+  useEffect(() => {
+    onDoneRef.current = onDone;
+  });
+  useEffect(() => {
+    const ms = Math.round(
+      AUTOPASS_DELAY_MIN_MS + Math.random() * (AUTOPASS_DELAY_MAX_MS - AUTOPASS_DELAY_MIN_MS),
```
fedepoi, line 25: are we totally sure this does not cause race conditions ??? it's a delay inside a use effect not the best (technically ww should avoid use effects in general)

fedepoi (approved):

src/components/prompts/ChooseAction.tsx
anacletos: mmh no not race conds.
Basically this mounts when autopass triggers.
On mount, it randomises a delta on the time it'll take before it autopasses.
It only happens once, and this component will unmount once it "autopasses" or something else blocks it.

fedepoi (dismissed):

fedepoi (approved):

Issue #598, opened 2026-08-03
web rollouts drop every relay connection


khaliostr:
Players report games that disconnect in the middle of a match. I read the relay logs and the Grafana metrics on prod for August 1st-3rd.

What happened:

- On August 2nd at 14:25:10 UTC the release deploy recreated the web container. In that same second the relay dropped 98 connections.
- On August 1st at 14:11 UTC the release deploy restarted the relay container. The effect was the same.
- Each hosted node room reconnected exactly two times in 48 hours. The two deploys above are the only server-side drops in the window.

Cause: the Caddy in the web container owns ports 80 and 443 and fronts `relay.manabrew.app`. Every web rollout recreates this container. The rollout drops every websocket at once: players, bots and nodes. At the current release cadence, every release is an outage for the players in a game.

The recovery path makes the damage worse:

- A client that keeps its session resumes the game. This worked 130 times in 48 hours.
- A client that returns inside the grace window retakes its seat, also after a reload. This path works as designed.
- A client that returns after the grace window gets the `game_already_started` rejection (147 in 48 hours). The client then retries the join every three seconds, without end and without a message to the player. The grace timer forfeited 96 seats in 48 hours.
- The `duplicate_username` retry storm is still active (925 rejections in 48 hours). One client produced 288 rejections in one day. See #362.

Fixes:

- [ ] Split the ingress from the web container, so that rollouts keep connections up (#599).
- [ ] Stop the client retry loop after a forfeit (#600). When the relay rejects a rejoin with `game_already_started`, the seat is gone. Stop the retries, clear the resume marker and tell the player that the seat was forfeited.
- [x] Resolve the duplicate-username storm in the reconnect path (#600). The #366 mechanism is live but does not cover auto-reconnect: `claimTabSession` runs only in `useServerStore.connect`, and the duplicate backoff assumes that the holder dies within the relay's 90s reap. When the holder is the player's other window, that assumption never holds and the client retries every 17 seconds without end. Fix: run the tab-claim on a duplicate rejection, stop with a "signed in in another window" message on refusal, and cap the no-holder backoff at about two minutes.

Correction (August 3rd): an earlier version of this list asked for a seat-reclaim mechanism. The relay already has one. A fresh session retakes its own in-game seat by username inside the grace window, and the client resumes guest seats after a reload (`useGameSessionResume`). The retry loop above is the real defect.

[...]

PR #599, opened 2026-08-03
fix(deploy): keep relay websockets alive across web rollouts

khaliostr:
## Summary

Add an `ingress` service (stock `caddy:2.10-alpine`). It owns ports 80 and 443 and the TLS certificates for every host, and it only proxies. The web container moves behind it: no published ports, three static vhosts on port 80, config in the new `ops/web.Caddyfile`. `ops/Caddyfile` becomes the edge config. Deploys list `ingress` in the `up` but never recreate it. Edge config changes apply with `caddy reload`. The deploy tool contract moves to v2. With this config, an old deploy binary recreates the web container without ports and takes the site down, so the version guard refuses it.

## Why

Every web rollout recreates the container that fronts `relay.manabrew.app`. The rollout drops every live websocket. On August 2nd at 14:25:10 UTC, prod dropped 98 connections in one second, at the moment the release recreated the web container. This fixes the first item in #598.

## Test plan

- `caddy validate` passes for `ops/Caddyfile` and `ops/web.Caddyfile` on caddy:2.10-alpine.
- `docker compose -f compose.production.yml config` resolves. Services: manabrew, ingress, manabrew-hub, manabrew-server.
- fmt and clippy pass on xtask.
- Before the tag: deploy the staging slot with `cargo xtask deploy --staging`.
- After the first prod rollout: run a second web-only deploy. Confirm that an open relay websocket stays connected through it.

## Notes

- The migration deploy drops connections one last time. The old web container must release ports 80 and 443 before ingress binds them. `depends_on` orders this.
- The certificates carry over in the shared `caddy-data` volume. Caddy does not reissue them.
- The staging slot already sits behind the edge proxy by service name. Nothing changes there.


khaliostr:
@fedepoi @anacletos do note that this makes socket-dropping less frequent. The reconnection mechanism works most of the time and the client code is robust, but sometime player lose the game because:

a. they hit the double username wedge
b. the oppo goes away during the deploy window
c. they come back after the 60s grace period and lose the game
d. they think the game is stuck and quit before the deploy is completed

this pr avoids the "freeze" during deploys and reduces the tail of lost games

anacletos (dismissed):

anacletos (approved):

PR #600, opened 2026-08-03
fix(ui): stop duplicate-username and forfeited-seat retry loops

khaliostr:
## Summary

Two reconnect loops retried forever against rejections that can never change. Both now stop and tell the player what happened.

Duplicate username: on an "already taken" rejection during auto-reconnect, the store now probes for a local holder over a new `probeTabSession` (BroadcastChannel, no handover side effect). A held username stops the retry loop with a "signed in in another tab" error, so a background tab never steals the seat from an active one. An unanswered probe (crashed browser, other device) keeps the slow backoff and stops after two minutes, past the relay's 90-second reap. The "already taken" fragment moves to one constant in `types/server.ts`.

Forfeited seat: `rejoinAfterRelayRestart` retried `JoinRoom` every two seconds for the whole window on every failure. A `game_already_started` rejection is terminal: the relay reseats any username with a disconnected seat, so the rejection proves the grace timer forfeited the seat. The loop now ends the game at once with "Your seat was forfeited while you were disconnected."

## Why

Prod shows one player with 288 duplicate-username rejections in one day (925 in 48 hours). Two windows of one browser trade the seat at each idle timeout, and the loser retries every 17 seconds without end. The backoff assumed that the holder dies within the relay's 90-second reap. A living second window never dies. The forfeit case has the same shape: a player who returned after the grace window retried the lost seat every three seconds. The player saw a spinner and no message. See #598.

## Test plan

- tsc, eslint and prettier pass on the touched files.
- Two tabs, same profile, both mid-reconnect after a relay bounce: the loser stops with the "another tab" error instead of retrying. The active tab keeps its seat.
- One tab, connection dropped for longer than the room's reconnect window, then restored: the client shows the forfeit message and returns to the lobby instead of looping on `game_already_started`.
- Explicit connect is untouched: a second tab still takes over via the existing claim handover.


anacletos (commented):
I get the fix and it technically works but I wonder if we're too deep in the shit at this point.

What if we just yank the session if this happens and we invalidate the other tab?
That way any connect is gonna work no matter what, at the risk of yanking the session from another tab.
The only risk with this is somebody "sniping" somebody else out of their session. Which I honestly think it's unlikely, and it'd be impossible as soon as we add validation on relay side?
what you think

khaliostr:
yank on any connect is the storm with the direction flipped. both tabs auto-reconnect after a relay bounce. if any connect wins, A's retry yanks B, B's retry yanks A, forever. the rejection spam becomes seat-swap spam and the active tab keeps losing its seat mid-game.

that's why the PR splits by intent. explicit connect yanks (existing handover, same browser). auto-reconnect never yanks: it probes, and stops with a message if the holder is alive.

your version works if the relay tells the yanked client to stop reconnecting. that's a relay + protocol + client change, and until auth is live it also lets anyone kick a player mid-game by connecting with their name. I'd ship this client-only fix now and do last-connect-wins on the relay once usernames are validated.

anacletos (dismissed):

anacletos (approved):

PR #604, opened 2026-08-03
feat!: allow cancelling during choose_targets

### LINKED ISSUE #577: UI and Harness - Allow cancelling a cast during target selection
The only way to cancel a cast right now, is from the payment prompt.
Targeting should also arguably allow a cancel.

anacletos:
## Summary

You can now cancel during targeting.
Closes #577


khaliostr (approved):

fedepoi (approved):

khaliostr (changes requested):
I'm leaving a blocker so we can merge #613 before if we're all happy. this is ofc still approved

PR #605, opened 2026-08-03
feat(deckhub): add versioned decks and official presets (#573)

anacletos:
resumes #573


khaliostr:
Heads up from the v3.2.0 deploy failure (fix in #606): the rewritten `migrate()` maps a DB without `schema_version` to version 0 and then runs `1_schema.sql` as-is. On a pre-accounts `hub.db` the baseline `CREATE TABLE IF NOT EXISTS hub_decks` no-ops, so `idx_hub_decks_account` fails with "no such column: account_id". That is the exact crash that rolled v3.2.0 back. `migrate_legacy_hub_decks` also selects `account_id`, so version 2 needs the column as well.

#606 adds `Storage::upgrade_legacy_schema` (a `pragma_table_info` check plus `ALTER TABLE`) and calls it at the top of `migrate()`. When this branch merges main, keep that call ahead of the version-gated loop. This diff replaces `fn migrate` wholesale, so a take-theirs conflict resolution drops the guard silently.

Easy repro: staging's current `hub.db` was recreated on 2026-08-03 by a pre-accounts image and lacks the column. Deploying this branch to staging without the guard crashes the hub on boot.

khaliostr:
Merged main in (e112f1ba9) and resolved the conflicts. `Storage::migrate` keeps this branch's version-gated runner with `upgrade_legacy_schema` from #606 called first, so pre-accounts databases survive the version-0 path. The hub row in manabrew-rs/AGENTS.md and the accounts bullets in src/AGENTS.md keep this branch's text; the usernames bullet takes main's newer reconnect-probe wording.

Verified against a copy of the prod hub.db: boot one runs the guard plus migrations 1 through 10 (19 legacy decks migrate, presets seed, `hub_decks` dropped by 6), boot two is a no-op at version 10. `cargo test -p manabrew-hub` passes, fmt and clippy clean.

fedepoi (dismissed):

khaliostr (dismissed):

khaliostr:
Summary of the changes layered on this branch since the Ironsmith merge, in review order.

Decks and engines
- Merged #531 (25 Ironsmith presets) and main.
- Preset JSONs can declare `engines`. Absent means every engine except Ironsmith, whose pool is verified per deck.
- `usePresetDecksStore` is the single preset load path. The worker no longer serves presets and `getPresetDecks` is gone from IGameApi.
- Pickers filter presets by the engine they launch on. Offline quick play refuses Ironsmith-only presets with a pointer to Multiplayer.
- Ironsmith-only decks are hidden everywhere unless the Ironsmith runtime is enabled in Settings.

Play home
- One My Decks area with explicit sections: your decks, preset decks, community decks. Labeled source and format filter rows. Scrolling is confined to the deck area.
- The community section shows the public catalog. Preset entries are excluded there so a preset never lists twice.
- Play on a commander deck asks 1v1 or 4-player pod, same dialog as the editor playtest.
- Engine-specific tiles carry an "Ironsmith engine" chip.

Hub
- Migration 11 adds `decks.engines`. Preset sync stores it and entries expose it in the API and generated types.
- Preset ingestion keeps sideboards. They were dropped before, which corrupted published fingerprints.
- Listings accept an `engines` query param so engine filtering happens before pagination. The old client-side filter emptied first pages when the newest entries were all Ironsmith.
- Listings exclude soft-deleted decks. Deleting a deck now removes its publications from browse, which also covers the duplicate-publication case from legacy republish.

Fixes
- Publish dialog gained a Title field. The title was silently the deck name and the only input was tags, so typed names became tags.
- The engine now initializes explicitly in appInit. It used to start as a side effect of the preset fetch, and removing that left the app stuck on the loading screen.

Migrations are frozen through 11.

fedepoi (approved):

PR #613, opened 2026-08-04
fix(xtask): cap app version bumps at minor

khaliostr:
## Summary

- `cargo xtask plan` now caps the app bump at minor. A breaking commit no longer majors the app.
- Crate bumps do not change. A breaking commit still majors every crate it touches.
- To release an app major, hand-set the version in `Cargo.toml`. The release run already honors hand bumps.

## Why

#604 is a breaking protocol change that also touches `src/`. If it merges first, its squash commit moves the app from 3.3.0 to 4.0.0. Nothing depends on the app's API. An app major is a product decision, so it needs a hand bump.

The early deploy gate does not change. It compares only `manabrew-protocol`, `manabrew-server` and `manabrew-hub`, and a breaking commit still majors those.

Merge this before #604.

## Test plan

- `cargo test -p xtask`: 11 passed
- `cargo clippy -p xtask --all-targets` and `cargo fmt --check`: clean
- `cargo xtask plan` runs end to end against the current tree


khaliostr:
approved on discord

Issue #628, opened 2026-08-05
feature request - Batch-Change Deck To Art of a Chosen Set / Read Parenthesized Set Art Tags When Importing Decklists


fedepoi:
From what I've seen so far trying to import decklists into Manabrew, it's selecting a variety pack that's hard for me to discern of set art. I'd love to be able to recreate decks with set art being as much as possible from a chosen set.

In the Deck Editor, it would be wonderful to be able to, instead of manually one card at a time, change art for *all cards in the deck* to the art for a chosen set *where available*, and if a card doesn't have art for the chosen deck, leave it as-is instead. It'd be great for preconstructed decks where you might want it to actually look like they're all from that precon.

As an alternative / addition to this idea… the impression I have is that right now Manabrew just imports cards fairly plainly when given a text list, but Archidekt and Moxfield do have the option to list things like `(C15)` or whatever on a card's line to show what set it's from. Is there a way you could get the importer to read those set labels in the lines, and (if available) use that set art rather than whatever it's typically pulling by default?

---
Reported by **prusten** in the Feature Requests Discord forum: https://discord.com/channels/1502665088304419008/1534673780985823435

khaliostr:
Closing as completed by the merged editor work. #633 implemented the parenthesized set/collector printing suffixes on text import, and #678 implemented the batch "choose a set and update every matching card, leaving non-matches unchanged" workflow. Together they cover both asks in this ticket.

PR #633, opened 2026-08-06
feat(editor): read printing suffixes in text deck import

khaliostr:
## Summary
Text import now reads Moxfield/Archidekt printing suffixes such as `1 Sol Ring (C15) 231 *F*`. The importer fetches that exact printing and stores the foil flag on the card. Lookup falls back in order: set plus collector number, then set, then default printing, then fuzzy name. A wrong collector number fails a name check instead of importing a different card.

First half of #628. The bulk "change all art to a chosen set" editor action stays tracked there.

## Why
Imports ignored set labels and picked default printings, so decks came in with mixed set art.

## Test plan
- `yarn lint` green.
- Parser exercised against Moxfield, Archidekt and Deckstats sample lines: set only, set plus number, foil with and without number, split names, dash collector numbers, categories after the suffix, `B.F.M. (Big Furry Monster)` left intact.
- Lines without a suffix import exactly as before.

Based on #632 (clippy drift blocks the pre-commit hook); retargets to main when it merges.


khaliostr:
Merged into #634 so the Deck Hub ranking, browse and preset fixes share one branch for staging. Nothing here changed. Close this once #634 lands, or land it first and #634 keeps the merge.

fedepoi (approved):

PR #634, opened 2026-08-06
feat(hub): rank vs-AI games, lead with community decks, fix preset duplicates

khaliostr:
> **Superseded by #641.** The hub half of this branch is merged there as f7447b125, without this branch UI work. See the comment below for what carried over and what did not. Close this when #641 lands.

## Summary

- Hosted games now count towards Deck Hub play rankings. Migration 12 adds `deck_play_reports.hosted`; Highest Win Rate still excludes them.
- Community decks are listed before official presets, grouped into their own sections, and this is the default browse order.
- A preset no longer appears twice once you have played it and it has been forked into your account.
- The Community shelf on the play home is no longer always empty.
- Folds in #633 (text import printing suffixes) so there is one branch to put on staging.

## Why

Deck Hub and accounts went live on 5 August. Three days of production data showed the ranking and browse surfaces were not doing their job:

**Nothing could ever rank.** `relay_deck_game_handler` returned early on any hosted game, and hosted is 93.6% of play (2,337 of 2,498 games over 14 days). Only five play reports existed, all client-side offline ones, so "Most Played" showed five decks that had each been played once and three of the seven buckets were empty. Win rate is kept clean because the AI loses 62-89% of the time, so a board fed by it would rank whoever farmed it hardest.

**Presets buried the community.** The hub holds 68 official presets and 19 user decks, and every preset was published after every user deck. Sorted by newest, real decks never appeared.

**The same preset showed up twice.** First use forks a preset into the account under the same name, and no surface filtered the catalog against decks already forked, so one offline play produced an apparent duplicate. The fork itself is idempotent; this was presentational.

**The Community shelf was always empty.** The play home requested the ten newest entries and then dropped presets in the browser. Since presets are newer than every user deck, page one was pure presets and nothing survived the filter, even though `/hub` listed 19 decks.

## Test plan

- `yarn lint:all` green.
- 54 hub tests pass. `cargo clippy -p manabrew-hub --all-targets -D warnings` clean.
- Migration 12 applied to a copy of the **production** `hub.db` and booted twice: v11 to v12, the second boot skips it, `pragma integrity_check` ok, 27 accounts / 113 decks / 87 entries / 32 sessions intact, existing rows default to `hosted = 0`.
- Empty-shelf bug reproduced against the live API before and after: `?engines=Manabrew,Forge&sort=newest&pageSize=10` returns 10 preset entries and 0 user entries, while adding `&source=community` returns 10 of 19 user entries.
- Not exercised in a browser yet.

## Demo

Not captured yet.

## Notes

Imported decks still cannot rank, because only a published deck carries a `publishedDeckId`. That is inherent, but it caps how full the boards get given "Imported Deck" is the most played deck by a factor of three.

Deleting a preset fork still does not stick: the delete is soft, and both the fork lookup and the unique index exclude soft-deleted rows, so the next play recreates it. Left alone here.

#633 is merged into this branch rather than duplicated. Close it once this lands, or land it separately and this branch keeps the merge.


khaliostr:
The hub half of this PR now lives in #641, merged there as a single commit (f7447b125) with this branch's UI left out. #641 rebuilds the same deck surfaces, so its own presentation wins.

Carried into #641:

- migration 12, the `hosted` flag on relay evidence, and the unhosted-only Highest Win Rate
- `DeckHubSortOrder::CommunityFirst` as the server default sort
- the preset dedupe on the play home, the deck editor and the offline picker
- community-only hub requests in both lobby pickers
- the #633 text import printing suffixes

Left behind, because #641 replaces or contradicts it:

- the shared tile face refactor: `DeckTile`, `DeckTileFace`, and the `DeckGridCard` / `DeckHubEntryCard` / `DeckSelectionCard` rewrites
- the community-first client defaults, the sort select option, and grouping browse by source
- deterministic Community-first section order in `DeckHubResults`
- playing a community deck straight from the play home
- the preset engine badge in My Decks
- the deck preview default card size, 3 to 4

The Community shelf fix does not carry over as such: #641 removed that shelf from the play home. Only the `source` argument it added to `useHubDeckSearch` survives, used by the lobby pickers.

Nothing here needs to land on its own any more. Close this once #641 merges, unless one of the dropped UI items is worth reopening separately.

Issue #467, opened 2026-07-15
ui: player-first surface (rfc)


khaliostr:
The app opens on a room list. Starting a game is six steps: room, deck, engine, bots, ready, start. All of that is engine-dev vocabulary, and the analytics say players route around it.

This issue is the direction, not a spec. #460 + #466 are the first slice.

## The numbers

Relay analytics (events.db), June 15 to July 15, 3,860 games:

- ~90% of games are one human vs bots. 85% Commander. 1,841 two-player tables vs 234 four-player.
- Most played deck name: "Imported Deck", 885 games, plus 419 with no name at all. People import a list and go play the AI.
- 96% of games run hosted Forge.
- Launch weekend brought 394 players; 59 came back the week after. 442 of 911 usernames played exactly one game.
- An average active hour has 6 distinct humans starting games. Same hour, playing alone.

Caveat: `game_over` stopped being recorded on July 6, so completion data is blind until that's fixed.

## The plan

Three pieces, in order. Nothing is removed; the lobby, tabletop and tools stay where they are.

**1. Deck-centric home** (#460 + #466, in review). You land on your decks, or precons plus an import box if you have none. Every tile plays in one click, Commander playtests seat a 4-player pod, the engine picks itself, and hub decks are links that open one button from a game.

**2. Advertise, don't wait** (next, touches the relay). "Play this deck online" posts an advertised game (in-app, plus a Discord announce), counts down, and backfills empty seats with bots so it always starts. A join-and-wait lobby loses at our concurrency; advertised games can catch the overlap that is already there.

**3. Lightweight identity** (open). Decks are localStorage, hub ownership is a local token, usernames are a persisted `@NNNN` tag. That caps 1 and 2: decks die with the browser, stats have no stable player. The question is how far a pseudonymous token or keypair gets us (deck sync, hub ownership, stable name) without building accounts.

## The stuff to actually decide

1. Matchmaking surface: in-app board, a Discord channel the bot posts to, or both. At our size Discord might be the real lobby.
2. Seat handover: can a late-joining human replace a bot mid-game? Neither engine host supports it today. Proposal: v2. In v1 the game advertises for N seconds, then starts with bots.
3. Identity floor: what does 2 strictly need from 3 (reconnecting to your own advertised game, at minimum), and what can wait.
4. Sidebar: home makes the Play / My Decks / Deck Hub entries redundant on paper. Add a nav-click analytics event first and delete based on data, not taste.
[...]

PR #638, opened 2026-08-07
feat!: relay session identity, hub oauth tokens, relay protocol crate

### LINKED ISSUE #653: split manabrew-protocol into relay protocol and engine/UI DTOs
## What

`manabrew-protocol` holds two different vocabularies.

`protocol.rs` is the relay wire protocol. It carries three groups:

- session: `Authenticate`, `AuthResult`, `SessionTakenOver`
- lobby: `ListRooms`, `CreateRoom`, `JoinRoom`, `SetReady`, `SetDeckSelection`, `SetFormat`, `StartGame`, `EndGame`
- game transport: `BroadcastState`, `StateUpdate`, `GameStarted`, `TurnChanged`

The rest of the crate is the engine and UI DTO contract: `game/`, `prompts/`, `display/`, `transport/`, `deck_dto.rs`.

## Why

Review of #638 raised that the session handshake does not belong with the DTOs that drive the UI. That is correct. It also applies to the lobby messages, which are not a UI ruleset either. A crate that holds both makes the relay wire format look like part of the UI contract.

Moving the auth message alone does not fix this. It leaves `CreateRoom` and `SetReady` in the same place.

## Scope

8 crates plus `src-tauri` depend on the crate. 32 Rust files import it. The change also touches three generators:

- `yarn gen:types` (ts-rs to `src/protocol/`)
- `yarn build:harness` (Java prompt DTOs)
- `yarn gen:protocol-docs` (`website/src/generated/protocol-types.json`, committed and gated in CI)

No behaviour change. The risk is a stale hop: a relay, node, harness jar or wasm bundle that misses the move.

## Checklist

- [ ] Decide the boundary: one crate for the relay protocol, or a separate session crate as well
- [ ] Move the `protocol.rs` types out of `manabrew-protocol`
- [ ] Update the 8 dependent crates and `src-tauri`
- [ ] Point the three generators at the new paths
- [ ] Update `manabrew-rs/AGENTS.md` and `src/AGENTS.md`
- [ ] Make sure that `yarn test:networking` passes

khaliostr:
> Was a three PR stack. #655 and #658 are now folded into this branch and closed, so this is the only open PR of the three. Review the three commits in order: relay identity, then hub OAuth, then the protocol split.

### 1. Relay session identity and owner takeover

Give the relay a way to tell "the same player coming back" apart from "someone else with your name", and let the proven owner take its own session over instead of being refused.

`Authenticate` gains an optional identity proof:

- `token`: the EdDSA identity token the hub mints (`POST /api/auth/token`, `aud: manabrew-relay`, 10 minutes), verified against its JWKS.
- `device`: an opaque per-browser secret, stored on the session as its sha256.

Both are sent when both exist. The relay resolves each proof to a `SessionIdentity` and keeps them on the session. On a username that is already held:

- Proof matches the session, live socket: the incumbent is sent `SessionTakenOver` and closed, the new socket inherits the session and its seat.
- Proof matches, socket disconnected or dead: the existing reclaim, unchanged.
- Session has an identity and the proof does not match: refused. This also closes an old hole, where anyone could take a disconnected player's seat by typing their username.
- Session has no identity: exactly the old behaviour, so bots, hosted nodes, and older clients are untouched.

Self-hosting stays first class. With no `MANABREW_HUB_JWKS_URL` set there is no hub to verify against, token proofs do not resolve, and the device secret carries the whole feature. A signed-in player on a hubless relay still gets takeover.

New counter: `manabrew_relay_session_takeovers_total{kind=account|device}`.

### 2. Hub auth on OAuth access and refresh tokens

- `POST /api/auth/token` is an OAuth token endpoint. It takes `grant_type=refresh_token` and an RFC 8707 `resource` value that selects the audience.
- `POST /api/auth/logout` is RFC 7009 revocation. The refresh token goes in the body.
- Sign-in returns an access token and a refresh token.
- Access tokens are 10-minute EdDSA JWTs. The hub verifies them by signature, so `SessionAccount` no longer reads the `sessions` table. That table now holds refresh-token hashes only.
- The client keeps only the refresh token in localStorage. Every other call site uses `getAccessToken()`.

Two audiences, one mechanism: `manabrew-hub` and `manabrew-relay` share one key, one JWKS and one verification path.

### 3. Relay wire protocol in its own crate

`manabrew-protocol` held two vocabularies. The relay wire messages now live in a new crate.

- **`manabrew-relay-protocol`** (new): `ClientMessage`, `ServerMessage`, the session handshake (`Authenticate`, `AuthResult`, `SessionTakenOver`, `IdentityProof`), lobby management (`RoomInfo`, `RoomStatus`, `PlayerInfo`, `RoomPlayerInfo`, `PlayerDeckInfo`, `DraftConfig`, `SealedConfig`, `ResumeRoomRequest`), `PROTOCOL_VERSION` and `DEFAULT_RECONNECT_TIMEOUT_S`.
- **`manabrew-protocol`** keeps the DTOs that drive the UI: decks, game-domain types, prompts, display events, transport envelopes.

The dependency runs one way. The relay crate names `manabrew-protocol` for `Deck` and the domain enums, and nothing in `manabrew-protocol` names the relay crate.

`GameFormat` and `EngineKind` stay behind, moved into `manabrew-protocol::game` next to the other domain types. `manabrew-hub` and `src-tauri` use them and have no business depending on a relay protocol. Moving `protocol.rs` wholesale would have forced exactly that, which is why the split is not a file move.

`PROTOCOL_VERSION` follows the messages. The new crate starts at **4.1.2** so the wire number stays 4 and the rule holds unchanged: the crate major is the single source of wire compatibility, never a separate constant.

Also moved: `gen-protocol` to the new crate (the only place in the protocol graph that can name both sides), with `xtask`, `scripts/harness.mjs` and `publish-protocol.yml` following it; `xtask`'s `GATE_CRATES` deploy gate, which would otherwise stop tripping on a breaking wire change; the `compose.dev.yaml` watchexec mounts on both relay services; the `protocol-docs.yml` path filter.

> [!NOTE]
> The crates.io name is reserved: `manabrew-relay-protocol` 0.0.1 is a stub published under `khaliostr`, following the same pattern as `packages/protocol`. `xtask publish` ships the real version when this merges. Ownership is `khaliostr` only, while `manabrew-protocol` is owned by `khaliostr` and `fedepoi`, so add `fedepoi` if the CI publish token is not khaliostr's.

## Why

425 auth rejections on prod in 24 hours, across 14 of 157 daily players. `J'ae@5969` was refused 256 times over six and a half hours, retrying every 15 to 20 seconds.

Relay identity was a bare username string. The shared server key authenticates the build, not the person, so when a second socket appeared under a live name the relay could not tell owner from stranger, and resolved that by always keeping the incumbent. The two escapes both miss the common case: the `sender_closed` check needs the old channel to be dead already, and the 180s stale sweep needs silence, which a suspended but pinging tab never gives.

The client-side work has gone as far as it can. The `@NNNN` tag (#366) fixed collisions between different people, and `tabSession` (#373) fixed two tabs of one browser. Neither can help across devices, or when the old socket is alive somewhere the BroadcastChannel cannot reach. Every rejected name in the prod logs already carries a tag.

The hub has minted these tokens since #517 and publishes its JWKS at api.manabrew.app/api/auth/jwks. Nothing consumed them. This wires the two ends together.

Review of the identity change asked for signed JWTs on the hub as well, and for one verification path instead of two, which is part 2. The audiences stay separate. RFC 9068 §2.2 makes `aud` required for a JWT access token, and §5 requires a distinct `aud` per resource from one issuer. RFC 9700, the January 2025 security BCP, gives the reason: an attacker can run a resource server and replay the tokens it receives at another one. We ship `compose.selfhost.yml`, so a relay is not always ours. A relay token asserts one account id to one relay and can do nothing on the hub.

Worth stating plainly: part 2 is a nice-to-have, not a fix. Part 1 works without it, and the session lookup it replaces was not slow or broken. The real gain is credential lifetime. Before this change the client attached a 90-day credential to every hub request from every tab. Now it attaches a 10-minute one, so incidental capture in a log, a proxy or a crash report is worth ten minutes instead of three months. XSS is unchanged, because the refresh token is still in localStorage. The cost is that revocation got weaker. Sign-out used to be instant. An issued access token now stays valid for up to ten minutes after sign-out. Account deletion is still immediate, because verification resolves `sub` against `accounts`.

The same review raised that the session handshake does not belong with the DTOs that drive the UI, which is part 3 and closes #653. That applies to the lobby messages too. Moving only the auth message would have left `CreateRoom` and `SetReady` in the same crate and fixed nothing.

## Breaking changes

Wire compatibility for part 1 is intact: `identity` defaults to absent and `SessionTakenOver` is additive, so old clients keep working against the new relay and new clients against an old one. The Rust struct change is source-breaking for anything constructing `Authenticate`.

Part 2 breaks the auth DTOs:

- `AuthSessionResponse` returns `access_token`, `token_type`, `expires_in` and `refresh_token` instead of `token`.
- `/api/auth/token` takes a refresh token in the body instead of a session bearer, and returns `access_token` instead of `token`.
- `/api/auth/logout` takes the refresh token in the body instead of an `Authorization` header.

These DTOs use OAuth's own wire names, so they are snake_case where the rest of the crate is camelCase.

Signed-in users survive the deploy. The old opaque session token is already a valid refresh token, because the `sessions` table did not change, so the client persist migration only renames the key. This is the one part with no gradual failure mode: if it is wrong, every signed-in user is logged out on release.

## Deliberate omissions

An anonymous player could previously retake a disconnected seat from any device by typing the username. That is the same hole that let a stranger take it, so it is now refused unless the session has no identity at all. Signed-in players keep cross-device takeover through the account token.

Relay usernames are still not reserved per account, so two accounts can still pick the same name and collide. Fixing that needs the hub handle to become the relay identity, which is #467's question.

No refresh-token rotation. RFC 9700 recommends it for public clients, but rotation without family-reuse detection breaks concurrent refresh across tabs. That deserves its own change.

## Test plan

- `cargo test -p manabrew-server`: token verification against a generated hub key, plus tampered, expired, wrong-audience and foreign-key rejection, device-secret hashing and bounds, and the hubless fallback to the device proof.
- `cargo test -p manabrew-hub`: 58 passed. Covers the token endpoint per audience, rejection of an unknown `resource` and an unsupported `grant_type`, revocation, and verification of a minted token against the published JWKS.
- `networking-tests` gains `proven_owner_takes_over_its_live_session`: a live seated socket, a stranger with a different device secret refused, the owner with the same secret taking the session over, the displaced socket getting `SessionTakenOver`, and the seat still held afterwards. Suite: 12 passed, 0 failed. This is also the real check on the crate split, because it spawns the built relay, node and bot binaries and drives real sockets across the new crate boundary.
- `cargo check --workspace` clean, including `src-tauri`. `yarn test` 18 passed. `yarn lint:all` and `yarn typecheck` clean.
- `yarn build:wasm` succeeds, so the wasm bundle still compiles against `ServerMessage` in its new home.
- `yarn gen:types` and `yarn gen:protocol-docs` regenerated and committed.
- Staging: sign in, connect from two browsers, confirm the second takes over and the first is signed out with a toast. Then confirm a hubless relay (`MANABREW_HUB_JWKS_URL` unset) still hands over on the device secret alone.
- Staging: confirm an account signed in before the deploy stays signed in, that sign-out revokes the refresh token, and that an access token issued just before sign-out stops working within ten minutes.
- Watch `manabrew_relay_client_rejections_total{reason="duplicate_username"}` against `manabrew_relay_session_takeovers_total` after rollout.

Not verified here: an actual publish of `manabrew-relay-protocol`. That needs the crates.io name first.


manabrew-rs/crates/manabrew-relay-protocol/src/lib.rs
```diff
@@ -43,6 +51,8 @@ pub enum ClientMessage {
         password: String,
         #[serde(default)]
         service: bool,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        identity: Option<IdentityProof>,
     },
```
anacletos, line 60: This shouldn't be here

anacletos (dismissed):
PR looks good but protocol shoulnd't be concerned with identity.
Also we need to test this extensively before sending it. Im happy to do the QA

manabrew-rs/crates/manabrew-relay-protocol/src/lib.rs
khaliostr: Apologies, if not in the protocol, how do we verify identity attestation out of band? Surely we can't roundtrip relay / deckhub on connection or introduce fan out in the hot path

khaliostr:
Ok thanks! Sure, we can deploy it on staging and exercise it there for a while

manabrew-rs/crates/manabrew-relay-protocol/src/lib.rs
khaliostr: see new commit for crypto attestation.I'm looking to remove the identity from the game protocol itself

khaliostr:
Both review points are addressed, and split out so each one can be reviewed on its own.

**Hub JWTs: #655**, stacked on this branch. Hub auth now follows OAuth 2.0. `/api/auth/token` is a token endpoint, `/api/auth/logout` is RFC 7009 revocation, access tokens are 10-minute EdDSA JWTs verified by signature, and `sessions` holds refresh-token hashes only. One key, one JWKS, one verification path.

The two audiences stay separate there. RFC 9068 §2.2 makes `aud` required for a JWT access token, and §5 requires a distinct `aud` per resource from one issuer. RFC 9700 gives the reason: an attacker can run a resource server and replay the tokens it receives at another one. We ship `compose.selfhost.yml`, so a relay is not always ours.

That work is a nice-to-have and I want to be straight about it. This PR does not need it. The device secret is what stops the duplicate-username failures. The real gain in #655 is credential lifetime, from 90 days down to 10 minutes on every hub request. If you would rather not take it, #638 ships without it.

**Protocol placement: #653.** `protocol.rs` holds three vocabularies: session, lobby and game transport. The UI ruleset is `game/`, `prompts/`, `display/` and `transport/`. So the coherent split is the whole file, not the auth message: 8 crates, 32 Rust files, three generators. `Authenticate` predates this branch, which only adds a field to it.

**This PR** now carries the relay change alone. `cargo test -p manabrew-server` and `yarn test:networking` 12 passed, 0 failed, `yarn lint:all` clean.

Staging QA is the next step before it leaves draft, whenever suits you.

anacletos (approved):

anacletos (approved):

PR #641, opened 2026-08-08
feat: unify deck collections and rank vs-AI games

fedepoi:
## Summary

Deck collections:

- simplify the Play homepage deck shelf to owned decks plus a collapsible preset fallback
- render one auth-aware My Decks collection in the deck editor and migrate local decks after sign-in
- turn Community into an infinite Explore feed with varied editorial layouts
- lead Explore with a Top Decks ranking teaser that opens the dedicated `/hub/top` page
- preserve legacy `?tab=top` links through redirects
- consolidate filters into a collapsible desktop rail and one mobile sheet
- share one visual deck-card system across My Decks and Community layouts

Hub, merged from #634 with its UI left out:

- hosted games count towards play rankings. Migration 12 adds `deck_play_reports.hosted`; Highest Win Rate still excludes them
- a preset no longer appears twice once it has been forked into your account
- the lobby pickers ask the hub for community entries only, so a preset is never labelled a Community deck
- the server default browse sort is community-first
- folds in #633 (text import printing suffixes) so there is one branch to put on staging

## Why

Owned decks should form one coherent collection, and Community should encourage layered discovery instead of presenting competing tabs or a paginated table. Rankings now act as the first editorial rabbit hole inside Explore, with the complete comparison experience one level deeper.

The hub half fixes what three days of production data exposed after Deck Hub went live on 5 August. **Nothing could ever rank:** `relay_deck_game_handler` returned early on any hosted game, and hosted is 93.6% of play (2,337 of 2,498 games over 14 days). Only five play reports existed, all client-side offline ones. Win rate is kept clean because the AI loses 62-89% of the time, so a board fed by it would rank whoever farmed it hardest. **The same preset showed up twice:** first use forks a preset into the account under the same name, and no surface filtered the catalog against decks already forked. The fork itself is idempotent; this was presentational.

#634's own UI work is deliberately dropped, because this branch rebuilds the same surfaces. Kept from it: the hub schema and queries, the preset dedupe, community-only picker requests, and the import suffixes. Dropped: the shared tile face refactor, the community-first client defaults and grouping, the play-from-community-home button, preset engine badges, and the deck preview card size.

## Test plan

- `yarn lint:all` green
- `yarn test`, 18 tests
- 54 `manabrew-hub` tests pass. `cargo clippy -p manabrew-hub --all-targets -D warnings` clean
- migration 12 applied on #634 to a copy of the **production** `hub.db` and booted twice: v11 to v12, the second boot skips it, `pragma integrity_check` ok, 27 accounts / 113 decks / 87 entries / 32 sessions intact, existing rows default to `hosted = 0`. The SQL is unchanged here
- browser smoke test at desktop and iPhone 12 Pro widths
- verified curated layouts, list cards, infinite loading, and collapsible filters
- verified the Top Decks teaser is the first editorial entry
- verified `See rankings` navigates to `/hub/top`
- verified legacy `/hub?tab=top&bucket=…` URLs retain the bucket after redirect
- the hub merge itself has not been exercised in a browser

## Demo

Validated locally in the collaborative browser at desktop and mobile widths. Community now flows from a ranking teaser through varied editorial sections into the endless deck grid, while full rankings live on a dedicated nested page.

## Notes

The play home no longer carries a Community shelf, so #634's fix for that shelf always being empty is moot. Only the `source` argument it added to `useHubDeckSearch` survives, in the lobby pickers.

Imported decks still cannot rank, because only a published deck carries a `publishedDeckId`. That is inherent, but it caps how full the boards get given "Imported Deck" is the most played deck by a factor of three.

Deleting a preset fork still does not stick: the delete is soft, and both the fork lookup and the unique index exclude soft-deleted rows, so the next play recreates it. Left alone here.

Close #634 and #633 once this lands.


khaliostr:
@fedepoi this seems to conflict with #634... can we sync?

khaliostr:
#634's UI work was dropped when its hub half was merged here, because this branch rebuilds the same surfaces. Checking what that actually costs: four items are worth doing on top of this branch, not before it.

**1. Lobby deck cards are outside the new card system.** `DeckCardSurface` is wired into `DeckGridCard` and `DeckHubEntryCard` only. `DeckSelectionCard` is untouched, so `CreateGameDialog` and the offline picker still draw decks the old way while My Decks and Community draw them the new way. #634 had folded that card into its shared tile face. Worth the most of the four.

**2. One-click play from a hub deck card.** A community deck goes card, `HubDeckPreviewDialog`, Play. #634 put a Play button on the card that routed straight to the offline picker with the deck preselected. It matches the player-first direction and saves a step. It needs re-placing against the new surface, not porting.

**3. Preset engine tag in the deck editor.** `PresetDeckShelf` passes `engines` to its cards on the play home, the editor's My Decks preset cards have no such prop, so the same deck is tagged in one place and untagged in the other.

**4. Deck preview opens small.** `DeckCardBrowser` defaults to card size 3. #634 used 4. Two lines, independent of either branch.

Nothing here blocks this PR. Two things from #634 are deliberately not on the list: the community-first sort and group-by-source defaults, since this branch drops user-facing sort and grouping controls on purpose, and the play-home Community shelf, which the Explore tile replaces.

khaliostr (dismissed):

khaliostr:
Top job

khaliostr (approved):

khaliostr:
approved on discord

PR #648, opened 2026-08-08
ci: move docker layer cache from actions cache to ghcr

khaliostr:
## Summary

Switches buildkit layer caching in `docker-images.yml`, `cache-warm.yml` and `staging-deploy.yml` from `type=gha` to `type=registry` on GHCR.

## Why

The repo is over its Actions cache quota, so GitHub is evicting continuously:

| Category | Size | Entries |
| --- | --- | --- |
| buildkit blobs | 6.26 GB | 254 |
| Rust caches | 3.29 GB | 9 |
| node/yarn | 1.39 GB | 2 |
| **Total** | **11.77 GB** | 278 |

The limit is 10 GB. `mode=max` saves every intermediate stage, which is why 254 blobs take 53% of the quota, and they compete directly with the Rust and tauri caches that actually shorten builds. The macOS tauri caches (704 MB and 635 MB) only rebuild on some PRs, which makes them prime eviction targets.

Registry storage does not count against the Actions quota and the images already push to GHCR, so this frees ~6.3 GB and leaves the compile caches room to survive between runs.

Two details worth review:

- `staging-deploy.yml` reads main's `:buildcache` but writes to `:buildcache-staging`, so a staging build never overwrites what production pulls from.
- `cache-warm.yml` gained a GHCR login. It never needed one while the cache was local, but pushing a cache manifest is a registry write even though `push: false`.

`image-manifest=true,oci-mediatypes=true` is required for GHCR to store the cache manifest.

## Test plan

- All three workflows parse, `prettier --check` clean, and no `type=gha` remains outside a comment.
- Cache figures above read from `actions/cache/usage` and a full paginated `actions/caches` dump.
- Real verification is the first `main` build after merge: it will be a cold miss while it seeds `:buildcache`, and the run after should hit. Worth watching `actions/cache/usage` drop as the old blobs age out.

The existing 6.26 GB of blobs are not deleted here. Nothing will write them after this merges, so they age out on their own, or can be purged to reclaim the space immediately.


khaliostr:
Verified end to end by dispatching `cache-warm` on this branch twice.

**A real bug the first review pass missed:** `cache-warm.yml` ran with only `contents: read`. That was correct while the layer cache was local, but exporting to GHCR is a package write, so it would have 403'd on merge. Fixed in e8f70a983, which also marks every cache export `ignore-error` so a registry hiccup can never fail a release build.

**Auth and export confirmed.** All four `:buildcache` manifests are live in GHCR (anonymous `GET /v2/witchesofthehill/<image>/manifests/buildcache` → 200 for web, server, hub, node). Checked directly rather than trusting the green check, since `ignore-error=true` could otherwise mask a failed export.

**Cache reads hit.** Same jobs, cold seeding run vs warm run:

| Image | Run 1 (cold) | Run 2 (warm) |
| --- | --- | --- |
| manabrew-web | 10.1 min | 0.7 min |
| manabrew-node | 6.6 min | 0.7 min |
| manabrew-hub | 3.9 min | 0.8 min |
| manabrew-server | 3.5 min | 0.6 min |
| **Total** | **24.1 min** | **2.8 min** |

Run 1 also confirms a missing cache ref is a miss and not an error, which is what the first build after merge will do.

The two `Warm macOS/Windows tauri` failures in those runs are the #647 checkstyle bug, not this change: this branch is cut from `891b32f6b`, which predates that fix.

PR #649, opened 2026-08-09
perf(ci): cache wasm, harness jar and native-image outputs

khaliostr:
## Summary

Caches the three expensive artifacts the desktop and mobile jobs rebuilt from scratch on every run.

| Artifact | Per-job cost | Cached by |
| --- | --- | --- |
| wasm bundle + cardset | ~4.9 min | existing key, now wired everywhere |
| Forge harness jar | ~1.1 min | `harness.mjs checksum` |
| GraalVM native library | ~4.3 min | `harness.mjs native-checksum` |

Up to ~10.3 min per job, on the runners GitHub bills at 10x (macOS) and 2x (Windows).

## Why

Measured on `Warm macOS tauri` (a warm run, 20.9 min total): `Build WASM + bundle cards` 4.9 min, `Build native Forge library` 4.3 min, `Build Forge harness jar` 1.1 min. None of it was cached.

**wasm** already had a proven cache in this repo, but only 2 of the 12 `build:wasm` sites used it. The other 10 now reuse that key verbatim, including its content-addressed `hashFiles` (its own `ensure-wasm.mjs` stamp can't be used in CI because it hashes card-data mtimes, which a fresh checkout always changes).

**native-image** is the one needing a new key. `computeChecksum()` hashes what goes *into the jar*; native-image also reads the FFI entrypoint, `build-native.{sh,ps1}`, `frozen-config`, `extra-config` and Forge's `res/languages`. `computeNativeChecksum()` layers the jar checksum with those.

Worth reviewing carefully: **nothing revalidates a restored native library.** The jar cache has a second guard, since `ensure:harness` recomputes its checksum at runtime and rebuilds on mismatch. The native cache has only the key, so an input missing from the hash would ship a stale `libforgeharness` in a release. `scripts/AGENTS.md` gained a note to that effect.

`publish.yml` uses `actions/cache/restore@v4` throughout, matching the existing `save-if: false` policy that stops tag builds evicting main-scoped caches.

## Test plan

Key sensitivity, verified by perturbing each input and diffing both hashes:

| Input touched | jar key | native key |
| --- | --- | --- |
| `ForgeNative.java` | same | CHANGED |
| `build-native.sh` / `.ps1` | same | CHANGED |
| `frozen-config/*`, `extra-config/*` | same | CHANGED |
| forge `res/languages/*` | same | CHANGED |
| harness Java source | CHANGED | CHANGED |
| protocol `.rs` | CHANGED | CHANGED |

Native-only inputs leave the jar key alone (no wasted Maven runs); jar inputs correctly cascade.

Staleness path, via `harness.mjs check`: no jar → stale; jar without checksum → stale; jar + matching checksum → fresh (Maven skipped); perturb a harness source → stale again.

Also: both checksums are deterministic across runs, all three workflows parse, every `build:wasm` and `build-native` step is guarded by its cache-hit condition, `prettier --check` clean, `yarn lint:all` green.

Cache behaviour itself is exercised on this PR with the `ci:test-desktop` label, which turns on the GraalVM macOS/Windows jobs: first run cold, second run should hit.


khaliostr:
Exercised on this PR with `ci:test-desktop`, which turns on the GraalVM macOS/Windows jobs.

**The first pass found a bug that the green checkmarks hid.** The jar cache restored and then did nothing:

```
Cache restored from key: harness-jar-0bd8e7c9…
harness: sources changed (checksum mismatch)
[INFO] BUILD SUCCESS
```

`computeChecksum()` hashes `forge-harness/src`, which holds the generated `forge.harness.protocol` classes. The key is computed before codegen; `updateChecksum()` writes its value after. On a fresh checkout those can never agree, so Maven ran every time. It only works locally because the generated files persist between runs.

Fixed in two commits, and the second is the one that matters: caching the generated sources alone would have changed nothing, because cache entries are immutable — the old key would have kept restoring the entry saved without them and never re-saved. Hence `harness-jar-v2-`.

**After the fix**, per GraalVM job:

| Step | Cold | Warm |
| --- | --- | --- |
| Build wasm + cardset | 1.3 min | SKIPPED |
| Build Forge harness jar | 1.4–1.6 min | 0.1 min |
| Build native Forge library | 3.0 min | SKIPPED |

Warm log confirms the jar path rather than just a fast run:

```
Cache restored from key: harness-jar-v2-0bd8e7c9…
Cache restored from key: forge-native-macOS-ARM64-3fa86bcb…
harness: JAR is up-to-date
harness: staged Tauri runtime at src-tauri/resources/forge-runtime
```

No Maven, no `BUILD SUCCESS`. Both keys match the values `harness.mjs` computes locally. The residual 0.1 min is `stageRuntime`, which has to run regardless since it stages the jar and Forge's `res/` into the Tauri bundle.

Note on the totals: whole-job time also dropped (macOS 17.3 → 8.6 min, Windows 21.8 → 13.0 min), but I would not attribute all of that here — the earlier attempt also warmed the cargo cache, which speeds the tauri build independently. The clean attributable wins are the three rows above.

These jobs don't build wasm, so the wasm row comes from `TypeScript & Wasm` on the same runs. Its full weight lands in the `publish.yml` and `cache-warm.yml` tauri jobs, which a PR can't trigger.

anacletos (approved):

PR #651, opened 2026-08-09
fix(ui): preserve token printing identity

### LINKED ISSUE #646: triage - Deck Editor showing wrong same-named tokens “produced by this deck” in Deck Analysis
I'm seeing something going wrong here in the Deck Builder, I think.
This doesn't seem to occur in gameplay.

It's not affecting gameplay, but in Deck Builder, the `deriveTokens` function isn't doing enough to specify a specific token by anything but name.
> `/src/lib/decks.ts` (function `deriveTokens`)
> `/src/hooks/useDerivedTokens.ts` (function `mergeDerivedAndCustomized`)
that are used in `/src/components/editor/DeckBuilder.tsx` seem to use merely the name from the pulled json from Scryfall. So the Dragon or Knight or Assassin or Elemental just uses a token that's *named that*, and not the token that Scryfall would *associate* with that card.

```const key = part.name.toLowerCase();```
This here. If the name is all that's being used to identify the part, then it's just going to go grab  a 'Knight' token, or a 'Dragon' token, or 'Angel' no matter what its ability is, and there are a variety! … but following the `part.uri` in the Scryfall json would lead to the json of the *specific token card that the card is referring to.* … which would have the same 'name'…

I dunno, a few functions would have to change here. You get where I'm going with this, though?

---
Reported by **prusten** in the Issue Tracker Discord forum: https://discord.com/channels/1502665088304419008/1535756019257581659

### LINKED ISSUE #650: triage - Killian, Decisive Mentor being listed as a Goat Token
I believe I'm using forge. I'm adding a new deck, and everytime I do, it lists Killian Decisive Mentor as my commander, but when I hover over it, it lists a goat token instead. The same for Eiganjo Dynastorian now being a Dragon Illusion token. Ghostly Dancers is an insect token, I'm sure there are others as well.

---
Reported by **baritonebiden** in the Issue Tracker Discord forum: https://discord.com/channels/1502665088304419008/1535895362215215154

### PR DESCRIPTION
**Pull request by fedepoi**
## Summary

- mirror Forge `TokenScript` extraction in the Rust card database and carry the script identity through Manabrew and Forge game snapshots
- generate and bundle a Forge token-script to Scryfall printing index for web, WASM, and desktop builds
- let users select an equivalent token printing and reuse that exact choice when the corresponding token appears in-game
- keep exact ordinary-card lookups separate from Forge token-set aliases

## Why

Closes #646. Forge identifies produced token kinds by token script, while display names and Scryfall relationships are not precise enough to distinguish every token. Using the same semantic identity across deck construction and gameplay makes printing choices stable.

Closes #650. Forge token-set aliases such as `tsoc #4` were also indexed as `soc #4`, causing ordinary cards with the same collector number to render as unrelated tokens.

## Test plan

- [x] `yarn lint:all`
- [x] `yarn test`
- [x] `yarn build:harness`
- [x] `yarn build:wasm`
- [x] `yarn vite:build`
- [x] validate the generated archive: schema 4, 2,835 token printings, 3,310 source-card mappings, and 714 token-script printing groups
- [x] verify Banish into Fable resolves `w_2_2_knight_vigilance` with 17 equivalent printings
- [x] verify Hunted Dragon resolves `w_2_2_knight_first_strike` with 2 equivalent printings
- [x] verify the chosen printing is reused only for the matching runtime token script
- [x] verify `soc #4`, `soc #13`, and `dsk #13` retain their ordinary card identities

## Demo

A deck containing Banish into Fable and Hunted Dragon now displays two distinct same-named Knight tokens based on Forge token scripts. Each picker offers only semantically equivalent printings, and the chosen printing is reused when that exact token script appears during gameplay. The token-script data is committed in `public/token_archive.json`, while the Forge token scripts themselves remain included in the shared cardset used by WASM and packaged desktop builds.

### CODE REVIEW

File: manabrew-rs/crates/manabrew-protocol/src/deck_dto.rs [hunk 1]
```diff
@@ -115,6 +115,9 @@ pub enum CardPartComponent {
 pub struct CardPart {
     pub name: String,
     pub component: CardPartComponent,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub scryfall_id: Option<String>,
```
anacletos: emmmh no this can't be necessary...

anacletos (changes requested):

manabrew-rs/crates/manabrew-protocol/src/deck_dto.rs
fedepoi: Scryfall `all_parts` gives us the exact related object only as `id`/`uri`, but `scryfallToDeckCard` currently reduces each part to `{ name, component }`. Once that `DeckCard` is saved (including through Hub/presets), the frontend has no way to distinguish same-named tokens such as the different Knight objects from #646. This optional field preserves the exact relationship across serialization; token Oracle identity is then resolved from the archive. Is there another existing persisted identifier you would prefer this to use?

khaliostr: Does forge use this mechanism too?

fedepoi: No, Forge does not use Scryfall IDs. It identifies the semantic token by the Forge `TokenScript` (for example `w_2_2_knight_vigilance` versus `w_2_2_knight_first_strike`), then `TokenInfo.getProtoType` asks the host edition for `getTokenSet(script)` and `TokenDb.getToken(script, edition)` selects the paper token/art. It also pins the selected edition by token script for the game. See `forge-game/.../card/token/TokenInfo.java:251-272`, `forge-core/.../token/TokenDb.java:194-221`, and `forge-core/.../card/CardEdition.java:467-474`. The shared principle is semantic identity rather than display name, but this deck-editor path has Scryfall `all_parts` data and no Forge `TokenScript` metadata. Its exact related-object ID is therefore the available persisted association; we resolve that to Oracle identity to group equivalent printings. Gameplay still begins with Forge’s emitted exact token set/collector identity before applying the user’s selected equivalent printing.

fedepoi:
ok the new token lookup works with the integrated tokenscript 
image

anacletos (dismissed):

anacletos (approved):

anacletos (approved):

PR #655, opened 2026-08-09
feat(hub)!: move auth onto oauth access and refresh tokens

khaliostr:
> Stacked on #638. Review that one first. Base branch is `khaliostr/relay-identity`, so the diff here is only the auth commit.

## Summary

Move hub authentication onto OAuth 2.0.

- `POST /api/auth/token` is an OAuth token endpoint. It takes `grant_type=refresh_token` and an RFC 8707 `resource` value that selects the audience.
- `POST /api/auth/logout` is RFC 7009 revocation. The refresh token goes in the body.
- Sign-in returns an access token and a refresh token.
- Access tokens are 10-minute EdDSA JWTs. The hub verifies them by signature, so `SessionAccount` no longer reads the `sessions` table. That table now holds refresh-token hashes only.
- The client keeps only the refresh token in localStorage. Every other call site uses `getAccessToken()`.

Two audiences, one mechanism: `manabrew-hub` and `manabrew-relay` share one key, one JWKS and one verification path.

## Why

Review of #638 asked for signed JWTs on the hub as well, and for one verification path instead of two. This does that.

The audiences stay separate. RFC 9068 §2.2 makes `aud` required for a JWT access token, and §5 of the same RFC requires a distinct `aud` per resource from one issuer. RFC 9700, the January 2025 security BCP, gives the reason: an attacker can run a resource server and replay the tokens it receives at another one. We ship `compose.selfhost.yml`, so a relay is not always ours. A relay token asserts one account id to one relay and can do nothing on the hub.

Worth stating plainly: this is a nice-to-have, not a fix. #638 works without it, and the session lookup it replaces was not slow or broken. The real gain is credential lifetime. Before this change the client attached a 90-day credential to every hub request from every tab. Now it attaches a 10-minute one, so incidental capture in a log, a proxy or a crash report is worth ten minutes instead of three months. XSS is unchanged, because the refresh token is still in localStorage.

The cost is that revocation got weaker. Sign-out used to be instant. An issued access token now stays valid for up to ten minutes after sign-out. Account deletion is still immediate, because verification resolves `sub` against `accounts`.

## Breaking changes

- `AuthSessionResponse` returns `access_token`, `token_type`, `expires_in` and `refresh_token` instead of `token`.
- `/api/auth/token` takes a refresh token in the body instead of a session bearer, and returns `access_token` instead of `token`.
- `/api/auth/logout` takes the refresh token in the body instead of an `Authorization` header.

These DTOs use OAuth's own wire names, so they are snake_case where the rest of the crate is camelCase.

Signed-in users survive the deploy. The old opaque session token is already a valid refresh token, because the `sessions` table did not change, so the client persist migration only renames the key. This is the one part with no gradual failure mode: if it is wrong, every signed-in user is logged out on release.

## Deliberate omission

No refresh-token rotation. RFC 9700 recommends it for public clients, but rotation without family-reuse detection breaks concurrent refresh across tabs. That deserves its own change.

## Test plan

- `cargo test -p manabrew-hub`: 58 passed. Covers the token endpoint per audience, rejection of an unknown `resource` and an unsupported `grant_type`, revocation, and verification of a minted token against the published JWKS.
- `yarn test:networking`: 12 passed, 0 failed.
- `yarn test`: 18 passed. `yarn lint:all` and `yarn typecheck` clean.
- Staging: confirm an account signed in before the deploy stays signed in.
- Staging: confirm sign-out revokes the refresh token, and that an access token issued just before sign-out stops working within ten minutes.


khaliostr:
deploying stack tip on staging so QA can start @anacletos

khaliostr:
Folded into #638, which now carries all three commits and is merged up to `main`. Closing this so the stack is one PR against `main`. Branch `khaliostr/hub-oauth` is left in place.

PR #658, opened 2026-08-09
refactor(protocol)!: split the relay wire protocol into its own crate

### LINKED ISSUE #653: split manabrew-protocol into relay protocol and engine/UI DTOs

khaliostr:
> Stacked on #655, which is stacked on #638. Review those first. Closes #653.

> [!NOTE]
> The crates.io name is reserved: `manabrew-relay-protocol` 0.0.1 is a stub published under `khaliostr`, following the same pattern as `packages/protocol`. `xtask publish` ships the real version when this merges. Ownership is `khaliostr` only, while `manabrew-protocol` is owned by `khaliostr` and `fedepoi`, so add `fedepoi` if the CI publish token is not khaliostr's.

## Summary

`manabrew-protocol` held two vocabularies. The relay wire messages now live in a new crate.

- **`manabrew-relay-protocol`** (new): `ClientMessage`, `ServerMessage`, the session handshake (`Authenticate`, `AuthResult`, `SessionTakenOver`, `IdentityProof`), lobby management (`RoomInfo`, `RoomStatus`, `PlayerInfo`, `RoomPlayerInfo`, `PlayerDeckInfo`, `DraftConfig`, `SealedConfig`, `ResumeRoomRequest`), `PROTOCOL_VERSION` and `DEFAULT_RECONNECT_TIMEOUT_S`.
- **`manabrew-protocol`** keeps the DTOs that drive the UI: decks, game-domain types, prompts, display events, transport envelopes.

The dependency runs one way. The relay crate names `manabrew-protocol` for `Deck` and the domain enums, and nothing in `manabrew-protocol` names the relay crate.

`GameFormat` and `EngineKind` stay behind, moved into `manabrew-protocol::game` next to the other domain types. `manabrew-hub` and `src-tauri` use them and have no business depending on a relay protocol. Moving `protocol.rs` wholesale would have forced exactly that, which is why the split is not a file move.

`PROTOCOL_VERSION` follows the messages. The new crate starts at **4.1.2** so the wire number stays 4 and the rule holds unchanged: the crate major is the single source of wire compatibility, never a separate constant.

## Why

Review of #638 raised that the session handshake does not belong with the DTOs that drive the UI. That is right, and it applies to the lobby messages too. Moving only the auth message would have left `CreateRoom` and `SetReady` in the same crate and fixed nothing.

## What moved with it

- `gen-protocol` moves to the new crate. It is the only place in the protocol graph that can name both sides, and it still emits the whole TypeScript tree in one pass. `xtask`, `scripts/harness.mjs` and `publish-protocol.yml` follow it.
- `xtask`'s deploy gate gains the new crate. `GATE_CRATES` is what delays a deploy until installers exist after a semver-incompatible bump. Without this entry a breaking wire change would stop tripping it, because the wire version no longer lives in `manabrew-protocol`.
- `compose.dev.yaml` mounts the new crate on both relay services, so watchexec still live-rebuilds on a protocol edit.
- `protocol-docs.yml` gains the path filter.

## Cost

24 files, and almost all of it is one-line import changes. The 426 references the issue predicted never materialised: `manabrew-server/src/protocol.rs` and `manabrew-agent-interface/src/protocol.rs` re-export the whole module with `pub use`, so six import sites absorbed everything downstream.

`website/src/generated/protocol-types.json` gains 12 lines. `EngineKind` and `GameFormat` now generate into `game/index.ts` instead of `lobby/index.ts`. No frontend file imports `@/protocol/lobby` directly, so the barrel makes that invisible.

## Test plan

- `cargo check --workspace` clean, including `src-tauri`.
- `yarn lint:all` clean: eslint, prettier, tsc, `cargo fmt --check`, `cargo clippy -D warnings` across the workspace.
- `yarn test:networking`: 12 passed, 0 failed. This is the real check, because it spawns the built relay, node and bot binaries and drives real sockets across the crate boundary.
- `cargo test -p manabrew-hub` 37 passed, `cargo test -p manabrew-server --lib` 5 passed.
- `yarn build:wasm` succeeds, so the wasm bundle still compiles against `ServerMessage` in its new home.
- `yarn gen:types` and `yarn gen:protocol-docs` regenerated and committed.

Not verified here: an actual publish. That needs the crates.io name first.


manabrew-rs/crates/manabrew-protocol/src/game/mod.rs
```diff
@@ -417,3 +417,29 @@ pub struct PlaymatSettings {
...
+}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Default, TS)]
+#[ts(export, export_to = "game/index.ts")]
+pub enum EngineKind {
+    #[default]
+    Manabrew,
+    Forge,
+    Ironsmith,
+}
```
anacletos, line 445: Not a blocker but I was wondering where the notion of "engine kind" belongs.
If someone wants to add their own engine ot their stack, it feels weird that they would need a change in protocol to support the new one.

Maybe down the line this needs a more thought of system, perhaps free naming / manifest of  an engine could be a thing

khaliostr:
Folded into #638, which now carries all three commits and is merged up to `main`. Closing this so the stack is one PR against `main`. Branch `khaliostr/protocol-split` is left in place.

PR #673, opened 2026-08-10
fix(harness): isolate multiplied trigger targets

### LINKED ISSUE #502: Forge Harness - Multi-Triggered Abilities act like copies
Forge engine.
Playing a Yarok, the Desecrated deck and playing something like Reclamation Sage doesn't work as expected. The game acts like the first trigger was copied, rather than actually triggering the ability a second time and letting you choose other targets. This means the extra trigger just targets the same thing and is effectively wasted.

---
Reported by **graklinghunter** on Discord: https://discord.com/channels/1502665088304419008/1528210113041793034

fedepoi:
## Summary

- detach a clean cached trigger template before target selection mutates the live wrapped ability
- keep multiplied triggers from inheriting target choices selected for an earlier firing
- add a harness regression and run all harness regression executables during builds

## Why

Forge models Yarok/Panharmonicon effects as separate firings of the original trigger, but the first wrapped firing can directly reference the trigger cached overriding ability. Harness target selection mutates that object, allowing later firings to clone previously selected targets instead of prompting from independent target state.

Fixes #502.

## Test plan

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) yarn build:harness`
- `yarn lint:all`
- `git diff --check`


anacletos (changes requested):
Does this new test run on CI?

fedepoi:
Addressed the review feedback in d45a4a591:

- PR CI now invokes `node scripts/harness.mjs test`, which compiles and runs both harness regression executables without staging runtime assets.
- The trigger regression now models two firings selecting different cards and verifies the second firing starts clean and neither target choice mutates the other firing.
- Merged the latest `main` into the branch.

Verified locally with `node scripts/harness.mjs test` and `yarn lint:all`.

anacletos (approved):

PR #683, opened 2026-08-11
feat(ui): overhaul deck editor and collection workflows

fedepoi:
## Summary

- overhaul the deck editor with richer card actions, grouping, shortcuts, collapsible analysis panels, printing optimization, collection coverage, and a reusable preview rail
- add authenticated collection management with account sync, exact-printing and foil tracking, quick add, import/export, deletion, filtering, and grid or text views
- route deck-editor card data through the shared Scryfall store with batched, paced requests and exact set-plus-collector-number resolution
- repair deck launch, copy, community preview, navigation, and editor entry-point wiring

## Why

The deck-building flow had accumulated fragmented controls, inconsistent navigation, incomplete collection integration, and direct card-data requests that became unreliable for large decks and imports. This consolidates those workflows into a faster editor cockpit while preserving exact printing data and safe account synchronization.

## Test plan

- `yarn lint:all`
- `yarn test`
- `git diff --check`
- verify deck-editor quick add, card actions, multi-select drag and drop, undo/redo, grouping, sorting, collapsible panels, and keyboard commands
- verify printing optimization for owned, cheapest, foil, and non-foil printings, including cancellation during paced Scryfall requests
- verify collection quick add, large-file import preview, exact-printing and foil preservation, account sync, export, filtering, and deletion
- verify My Decks, preset, copied, and community deck CTAs open the intended editor or play flow

## Demo

UI changes span the deck editor, My Collection, My Decks, and community deck preview flows.


anacletos (dismissed):

khaliostr (dismissed):
super work

khaliostr:
Since this PR already touches `useDeckTextImport.ts`: the fallback name is still leaking to prod at a steady rate.

Last 21 days on prod: 314 games played on a deck literally named "Imported Deck", 10-20 per day, no downward trend. 264 of those are Commander games with a commander recorded, so the deck knows its commander and still carries the fallback name. New distinct decks with that name keep appearing (8 on 08-11, 8 on 08-14, 3 on 08-16), so it is not just old decks lingering.

The name is only derived at two moments:

- `useDeckTextImport.ts:91`, `customName || commanderName || DEFAULT_IMPORT_NAME`, where `commanderName` is non-empty only if the pasted text marked a commander section.
- `useDeckStore.ts:526`, inside `setCommander`, and only while the name is still the default.

A plain decklist paste misses the first. Any route that fills `commanders` without going through `setCommander` misses the second, including the legacy singular `commander` migration in `normalizeDeck`.

Worth deriving rather than assigning: when the stored name is `DEFAULT_IMPORT_NAME` or `DEFAULT_DECK_NAME` and the deck has commanders, resolve the display name from the commanders at read time (save, play, hub publish). That covers every route without adding another event hook.

Out of scope here is fine, say so and I will file it separately.

khaliostr:
could you also check if when you create.a new deck, and _then_  import a CMD list in the editor itself, the format is switched to CMD automatically?

khaliostr (approved):

PR #688, opened 2026-08-16
fix(build): make the harness build work on GNU/Linux

allentiak:
Problems:

* Building the harness on GNU/Linux with the documented `mvn -pl
forge-harness -am clean package` failed: `forge.harness.protocol.*`
was
  generated from the Rust protocol crate and gitignored, and only
  `scripts/harness.mjs` generated it before Maven runs.

* Every doc that reached for Maven directly produced a wall of `cannot
  find symbol` and constructor mismatch errors against those classes.
  The parity regression workflow made the same bare-Maven call, so it
  would fail the same way once re-enabled, and forge-harness's own
  AGENTS.md instructed agents to run `mvn -pl forge-harness -am clean
  package` after a submodule bump — which is how the bypass kept coming
  back.

* Platform assumptions were baked in well beyond the codegen: several
  defaults were macOS-only, the documented JDK path hardcoded both the
  Debian layout and `amd64`, and setup-linux-runner.sh hard-exited on
  anything that was not Debian/Ubuntu.

* No workflow ever executed setup-linux-runner.sh, so its package sets
  could only rot silently — nothing would have caught a wrong package
  name on any distro.

Solution:

* Point every entry point at `yarn build:harness`, and give it a
  `--clean` flag so the post-Forge-bump clean rebuild can no longer skip
  the codegen step. The regression workflow now runs
  `node scripts/harness.mjs test` rather than calling Maven itself.

* Make the macOS-only defaults OS-aware:

  - build-native.sh defaulted GRAALVM_HOME to a `Contents/Home` bundle
    path that does not exist on Linux; it now falls back to the plain
    directory and names the env var in its error.
  - The parity JDK auto-detect only scanned
/Library/Java/JavaVirtualMachines; it now also scans /usr/lib/jvm
and
    handles both layouts.
  - parity_tools.rs and parity-repair-agent.py hardcoded a macOS zulu-18
    JAVA_HOME; both now use the environment.
  - build-wasm.mjs suggested `brew install binaryen` on every platform.

* Derive JAVA_HOME from `javac` on PATH instead of naming a distro
  layout and an architecture.

* Teach setup-linux-runner.sh the Debian/Ubuntu (apt), Fedora/RHEL (dnf)
  and Arch (pacman) families, with per-family package sets and Node
  installs, and no x86_64 assumption anywhere — so it also provisions an
  aarch64 host such as Fedora Asahi Remix. Yarn falls back to a global
  npm install when corepack is absent: Fedora ships no corepack package
  and strips the copy bundled in the upstream Node tarball, and Arch
  splits it into its own package.

* Add a `linux-runner-setup` matrix to build-checks.yml that runs the
  provisioning script to completion in a container per family and
  asserts the toolchain it promises, plus that the documented JAVA_HOME
  derivation resolves against each distro's JDK layout. It is gated on
  the script changing, so it costs nothing on unrelated PRs.

* Add a `harness-entrypoint` guard that fails any workflow or Dockerfile
  invoking Maven on forge-harness without the codegen — through
  harness.mjs, or from a protocol-builder stage. That bypass is the root
  cause fixed here and it had already regressed once, in regression.yml,
  so the class of bug needs a guard rather than another one-off fix.

* Correct the JDK range wherever it was stated: Forge's enforcer
  requires 17+ and nothing caps it at 21, verified by building the full
  reactor on both 21 and 25. forge-harness's AGENTS.md claimed 18-21.

* Bring the docs back in step with the code they describe — README,
  CONTRIBUTING, PARITY_TESTING, the parity and native READMEs, the
  self-hosting guide, and the scripts and forge-harness AGENTS.md — so
  no remaining page tells a reader or an agent to invoke Maven directly,
  hardcode a JDK path, or cap the JDK at 21.

Verified by running that matrix body locally against debian:trixie,
ubuntu:24.04, fedora:41 and archlinux, all green: JAVA_HOME resolves to
/usr/lib/jvm/java-21-openjdk-amd64 on the Debian pair and
/usr/lib/jvm/java-21-openjdk on Fedora and Arch. Asahi maps to the dnf
branch by ID but is untested on real hardware, where the 16K page size
may still affect GraalVM native-image.

Code written with AI assistance.


khaliostr:
Thanks again for the work here. I went through the PR in more detail and I think there is a real issue being fixed, but the PR currently mixes that fix with a broader expansion of our Linux support surface.

The root cause of the original build failure seems to be the harness build path rather than Linux itself: forge.harness.protocol.* is generated, so invoking Maven directly on a fresh checkout can fail, whereas yarn build:harness / scripts/harness.mjs performs the required code generation first. The fact that this was not clear in the docs, and that some workflows were bypassing the canonical harness entrypoint, is worth fixing.

I am also in favour of removing genuine macOS-only assumptions where they exist. In particular, things such as assuming a GraalVM Contents/Home layout or hardcoding a specific Linux JDK path are reasonable portability fixes, provided the behaviour remains platform-agnostic.

One additional concern I have is around the environment-variable handling. I am not convinced that Java/GraalVM discovery and validation belongs in several lower-level .rs, .sh and other files.

We already have a canonical harness wrapper, and I would prefer that layer to own developer-environment validation and, where appropriate, discovery. In other words: check JAVA_HOME / GRAALVM_HOME there, fail early with a useful error if the required environment cannot be resolved, and then invoke the lower-level build with an explicit environment.

Otherwise we risk spreading slightly different Java/GraalVM discovery rules across Rust, shell scripts and CI. That makes the build harder to reason about and is likely to introduce another set of platform-specific assumptions over time.

There may still be small platform-normalisation fixes that belong in the bridge itself — for example handling a platform-specific directory layout after a Graal installation has already been identified — but I would distinguish that from making java_bridge.rs itself responsible for discovering the user’s JDK/GraalVM installation.

So for me there are three separate things here:

* fixing concrete platform-specific assumptions;
* validating the required environment at the canonical build entrypoint;
* automatically discovering/defaulting Java/GraalVM installations throughout the codebase.

The first two seem useful. I am much less convinced we want the third.

Where I am also less convinced is the additional distro provisioning and CI matrix. Supporting and bootstrap-testing Debian, Ubuntu, Fedora, Arch, etc. as first-class native development environments is a separate maintenance commitment from fixing the original build problem.

Our supported/recommended Linux deployment path remains Docker, and the images we build and run in production are already Linux-based. A native Linux source build should of course be possible, and improving the documentation around it is welcome, but I don’t think that automatically means the project should take ownership of provisioning and continuously validating multiple distro-specific development environments.

For the same reason, I would probably leave the broader Linux setup/provisioning scripts and distro CI matrix for a separate PR, where we can evaluate independently whether that is a support surface we want to maintain.

I have a similar reservation about the grep-based CI guard around Maven invocations. I understand what it is trying to prevent, but I would prefer the actual harness build/test in CI to be the source of truth unless we have a concrete regression that demonstrates we need an additional static policy check.

So my preference would be to keep this PR focused on:

* making the harness entrypoint and code-generation requirement explicit;
* fixing workflows/docs that invoke the harness incorrectly;
* fixing concrete platform-specific assumptions that prevent a native build with a correctly configured environment;
* centralising environment validation at the appropriate wrapper/build boundary;
* documenting the expected Java/GraalVM setup clearly.

Then, if you are interested, the broader native-Linux provisioning and CI work can be discussed separately.

Overall I think the original fresh-checkout build problem is legitimate and there are useful fixes in this PR. I would just like us to separate the necessary build/portability fixes from the broader proposal to expand what the project officially provisions and continuously validates.

khaliostr (changes requested):
Thanks again for the work here. I went through the PR in more detail and I think there is a real issue being fixed, but the PR currently mixes that fix with a broader expansion of our Linux support surface.

The root cause of the original build failure seems to be the harness build path rather than Linux itself: forge.harness.protocol.* is generated, so invoking Maven directly on a fresh checkout can fail, whereas yarn build:harness / scripts/harness.mjs performs the required code generation first. The fact that this was not clear in the docs, and that some workflows were bypassing the canonical harness entrypoint, is worth fixing.

I am also in favour of removing genuine macOS-only assumptions where they exist. In particular, things such as assuming a GraalVM Contents/Home layout or hardcoding a specific Linux JDK path are reasonable portability fixes, provided the behaviour remains platform-agnostic.

One additional concern I have is around the environment-variable handling. I am not convinced that Java/GraalVM discovery and validation belongs in several lower-level .rs, .sh and other files.

We already have a canonical harness wrapper, and I would prefer that layer to own developer-environment validation and, where appropriate, discovery. In other words: check JAVA_HOME / GRAALVM_HOME there, fail early with a useful error if the required environment cannot be resolved, and then invoke the lower-level build with an explicit environment.

Otherwise we risk spreading slightly different Java/GraalVM discovery rules across Rust, shell scripts and CI. That makes the build harder to reason about and is likely to introduce another set of platform-specific assumptions over time.

There may still be small platform-normalisation fixes that belong in the bridge itself — for example handling a platform-specific directory layout after a Graal installation has already been identified — but I would distinguish that from making java_bridge.rs itself responsible for discovering the user’s JDK/GraalVM installation.

So for me there are three separate things here:

* fixing concrete platform-specific assumptions;
* validating the required environment at the canonical build entrypoint;
* automatically discovering/defaulting Java/GraalVM installations throughout the codebase.

The first two seem useful. I am much less convinced we want the third.

Where I am also less convinced is the additional distro provisioning and CI matrix. Supporting and bootstrap-testing Debian, Ubuntu, Fedora, Arch, etc. as first-class native development environments is a separate maintenance commitment from fixing the original build problem.

Our supported/recommended Linux deployment path remains Docker, and the images we build and run in production are already Linux-based. A native Linux source build should of course be possible, and improving the documentation around it is welcome, but I don’t think that automatically means the project should take ownership of provisioning and continuously validating multiple distro-specific development environments.

For the same reason, I would probably leave the broader Linux setup/provisioning scripts and distro CI matrix for a separate PR, where we can evaluate independently whether that is a support surface we want to maintain.

I have a similar reservation about the grep-based CI guard around Maven invocations. I understand what it is trying to prevent, but I would prefer the actual harness build/test in CI to be the source of truth unless we have a concrete regression that demonstrates we need an additional static policy check.

So my preference would be to keep this PR focused on:

* making the harness entrypoint and code-generation requirement explicit;
* fixing workflows/docs that invoke the harness incorrectly;
* fixing concrete platform-specific assumptions that prevent a native build with a correctly configured environment;
* centralising environment validation at the appropriate wrapper/build boundary;
* documenting the expected Java/GraalVM setup clearly.

Then, if you are interested, the broader native-Linux provisioning and CI work can be discussed separately.

Overall I think the original fresh-checkout build problem is legitimate and there are useful fixes in this PR. I would just like us to separate the necessary build/portability fixes from the broader proposal to expand what the project officially provisions and continuously validates.

allentiak:
Thanks for the quick and thorough review, @khaliostr .
I think you are generally right. My initial aim was to simply fix the build on GNU/Linux, and to update the docs and CI accordingly. After the comments on support for specific distros and archs on the chat, I went too far with the changes.
I will review it and let you know.
Thanks again,
Leandro

allentiak:
@khaliostr A short question. 

You say the support for local builds for different distro is not intended. However, the local build currently makes three local builds: RPM, DEB, and AppImage.

Even though this is clearly outside de scope of this PR...
What do you think about this? Should (all, or some of) them go away?
I have to confess that their existence was what initially made me think they were also a recommended way of trying the system locally...

khaliostr:
Thanks, Leandro — and yes, I think your read on the scope is right.

On RPM/DEB/AppImage: I don’t think they need to go away. Tauri outputs those Linux bundle formats by default when building the desktop app on Linux, so their presence is mostly a consequence of the underlying toolchain.

We don’t currently consider Linux desktop stable or officially supported, mainly because of the Tauri/WebKit limitations we discussed. That means we don’t actively test, release, or guarantee those packages across distros.

At the same time, I don’t see any reason to prevent them from being built. If someone wants to build the Tauri client on Linux and it works for them, that’s perfectly fine.

So the distinction is really: buildable/best-effort vs supported/maintained.

For Linux, the maintained path remains the web stack/Docker. For this PR, I’d still keep the scope to the concrete harness/build fixes and portability issues.

allentiak:
@khaliostr

Thanks again for the detailed review. As agreed, I've narrowed the PR to focus on the fix itself.

What's in this overhauled PR:

- **All non-essential parts are gone**: the distro provisioning, the distro CI matrix, the grep-based Maven check and the JDK auto-detection in Rust.

- **Harness entrypoint.** Every doc and workflow that called Maven directly now uses `yarn build:harness` / `node scripts/harness.mjs`. That covers the README, `forge-harness` AGENTS.md, the parity README, the `pom.xml` comment, the self-hosting docs, `regression.yml` and `parity-repair-agent.py`. The README explains why: `forge.harness.protocol.*` is generated and gitignored. `harness.mjs` also gets a `--clean` flag, so the rebuild after a Forge bump no longer needs `mvn clean package` either.

- **Environment checks in `harness.mjs` only.** Before codegen, the wrapper asks the JDK from `JAVA_HOME` (or `java` on `PATH`) for its `java.home` and version. If there's no usable JDK, the JDK is older than 17, or Maven is missing, it stops and says which one. Maven then runs with that `JAVA_HOME` set explicitly. Nothing searches for a JDK in any other file, and `java_bridge.rs` is unchanged.

- **Platform-specific paths.** I removed the hardcoded macOS zulu-18 `JAVA_HOME` from `parity_tools.rs`, `parity-repair-agent.py` and `.env.example`, and the Debian JDK path from `regression.yml`. `build-native.sh` no longer defaults to a macOS `Contents/Home` GraalVM path. It now requires `GRAALVM_HOME`, the same as `build-native.ps1`. CI and the self-hosted-node Dockerfile already set that variable, so the only people affected are local macOS users who relied on the default.

- **Docs.** They now say JDK 17+ (Forge's own minimum; CI uses 21) and explain when `GRAALVM_HOME` is needed.


I tested it on Debian with JDK 21. A clean build passes, including the harness regression tests. It also passes with the generated protocol sources deleted, which is the fresh-checkout case this started from. The `regression.yml` job is still `if: false`, so that change hasn't run in CI. Once it's re-enabled, it will use the runner's own JDK and needs Node on the runner.

allentiak:
One question for later: would a local-dev env file (direnv or similar) be welcome in a separate PR? Right now the .env files are all for deployment.

khaliostr (commented):
Thanks a lot for the work on this, and for taking the time to clean it up after the previous review. This looks much better now.

Just one small thing: resolveJdk() checks for java but not javac, so a JRE could pass the check and then fail later in Maven. Could we check for a working javac there too?

Other than that, looks good to me. Thanks again!

Issue #684, opened 2026-08-11
hosted games: per-decision latency grows 10x within a session


khaliostr:
In a hosted 3-human Commander game, prompt latency went from 0.6s to 15s over 80 minutes. A comparable game on the same node two hours earlier stayed under 1.3s for its whole run.

All numbers below come from relay-side capture (`ops/observability/data/captures/`), which timestamps every envelope in both directions. None of it is node-side. We have no node logs (see "what we cannot see").

Code defects found while investigating this are in #685.

## The game

- game `a7ec7e17-d2ff-4437-aa48-3c05e8df395c`, room `8f09379b`, 2026-08-10 22:35 UTC
- node `self-hosted-node-b4070326`, engine Forge, hosted, official
- 3 humans, Commander: Atla Palani / Witherbloom, the Balancer / Ramses, Assassin Lord

One metric throughout: the time from a client action arriving at the relay to the first `state` or `prompt` the node sends after it. Heartbeats (`roomRelay`, every 30s) do not count.

## The slowdown

| game age | median | p90 | max |
| --- | --- | --- | --- |
| 0-10 min | 0.63s | 1.19s | 10.8s |
| 10-20 min | 0.96s | 2.79s | 3.4s |
| 20-30 min | 1.28s | 3.39s | 7.2s |
| 30-40 min | 1.75s | 6.55s | 10.7s |
| 40-50 min | 1.80s | 8.72s | 12.4s |
| 50-60 min | 9.42s | 16.04s | 29.1s |
| 60-70 min | 3.65s | 26.32s | 26.3s |

39 stalls of 8s or more. The wait happens before the node sends anything. Once it starts, it sends 4 states and 1 prompt in 40 to 500ms.

```
23:31:44.419  CLIENT player-2 pass
23:31:53.838  +9.42s  state -> player-0 (42944 B)
23:31:53.871  +0.03s  state -> player-1
23:31:53.908  +0.04s  state -> player-2
23:31:53.913  +0.01s  state -> ALL
23:31:53.915  +0.00s  prompt -> player-0
```

## Two effects, not one

**Effect 1: the action-space scan.** anacleto and fedepoi identified this. `ActionSpace.getPossibleActions` re-runs per prompt and probes mana payability per candidate ability, so cost scales with untapped permanents and hand size. The capture agrees. Two games on the same node the same evening:

| probe candidates (untapped + hand) | game 3a7bb349 | game a7ec7e17 |
| --- | --- | --- |
[...]

Issue #669, opened 2026-08-10
engine: rollback, tabletop mode and state injection are one problem (epic)


khaliostr:
Rollback, tabletop mode and state injection are one operation, used three ways. Most of it is already built. Forge ships all three upstream, and the Rust engine records a checkpoint every turn and can restore one. What is missing is that the rollback never reaches an online player, and that the two engines speak different vocabularies for the edit. The same system also gives a table with no rules engine at all.

Forge is first class today and under 1% of games run on the Rust engine. So an answer that works only on the Rust path serves almost nobody, and the Forge path leads in everything below.

This issue describes the problem space. No plan, no order of work.

## What people ask for

| Ask | Source |
| --- | --- |
| Table host rolls back to the start of the turn, or an earlier turn | #617, rustydawwwg |
| Room controller restores an earlier snapshot | #313 |
| Save a game state and load it | #304 |
| Undo or cancel a committed action | #592 (closed), and the untracked half of #424 |
| Tabletop mode: the table unlocks, players edit life, counters and zones, play resumes | #424, u/Dimmins2. Never filed on its own |
| Recover a commander stuck in the graveyard after a mis-clicked prompt | #424, kamigoroshi |

The reasons are the same in all six: friendly rule-breaking, recovery from our own bugs, teaching new players.

Below, "tabletop mode" is the name for all of the manual-edit asks. The prototype code calls the same thing maintenance mode.

## One operation

Remove the interface and the same four steps remain:

1. Stop at a point where the state is consistent.
2. Write to base state.
3. Recompute derived state and run state-based actions.
4. Send the new state to every seat, record the edit, resume.

They differ in one place only: where the new state comes from. Rollback takes it from the past of the same game. Tabletop mode takes an edit from a person. Save, load and injection take it from a document. The parity harness in #426 is the fourth case, where a machine writes a captured Forge state into the Rust base state to stop divergences cascading.

## The engine-free case

A table with no rules engine under it, the untap.in or Tabletop Simulator model, is a thing we want to offer. it was implemented but deprioritised. It is also what several of the asks above describe once you remove the rules enforcement.

It is not a hybrid and not a second mode. The shared editable state is the base of the system. The rules engine is one writer on top of it, the fastest and most consistent writer, and not the only one. A player is a writer. A rollback is a writer. The parity harness is a writer. "With engine" and "without engine" is a rule about who is allowed to write, not two products.

[...]

PR #690, opened 2026-08-17
feat(harness): load a forge dev-mode game state into a running game

khaliostr:
## Summary

- Adds `GameStateInjector`: reads a Forge dev-mode game-state file and applies it to a running game.
- `-Dharness.state.benchScans=N` then times N `ActionSpace.getPossibleActions` calls over the injected board and exits.
- Both properties are opt-in. Unset, the hook returns immediately and a normal parity or host run is unchanged.

## Why

Action-space scan cost is superlinear in board size, hand size and the number of continuous static abilities in play. There was no way to measure it against a stated board, because a bot will not assemble a late-game board on demand: across a day of trying, self-play plateaued well below the boards real games reach.

This came out of #684. Being able to state a board and measure it turned a chain of extrapolations into single measurements, including reconstructing the board from a production capture and timing the scan on it directly.

The format is Forge's own, the one `IDevModeCheats.dumpGameState` writes and `setupGameState` reads:

```
p0life=40
turn=23
activeplayer=p0
activephase=MAIN1
p0battlefield=Jungle Shrine;Forest;Atla Palani, Nest Tender;...
p0hand=Penance;Serra's Emissary;...
```

`GameState` lives in forge-game, so none of this depends on the GUI.

This is also the injection third of #669, which notes Forge has the primitive upstream and manabrew has not wired it. Scoped deliberately narrow here: the parity path only, driven by a system property. If the epic wants tabletop and rollback, the hook belongs in the host controller and on the protocol, which is a different and larger change.

## Test plan

```
java -Dpreset.decks.dir=parity_decks \
     -Dharness.state.file=state.txt -Dharness.state.benchScans=10 \
     -jar forge-harness/target/forge-harness-jar-with-dependencies.jar \
     --deck1 red_burn --deck2 green_stompy --seed 1 --max-turns 40 \
     --forge-home forge/forge-gui/
```

Verified:

- Injects and measures: `BENCH permanents=55 hand=15 actions=10 scans=5 median=31.7ms max=42.3ms`.
- Sweeps cleanly. Static abilities against hand size, median ms per scan:

  | statics \ hand | 5 | 15 | 30 |
  | --- | --- | --- | --- |
  | 10 | 5.8 | 9.2 | 17.4 |
  | 40 | 11.5 | 42.7 | 50.9 |
  | 80 | 29.0 | 72.8 | 151.5 |

- With neither property set, a normal parity game runs to completion unchanged.
- `yarn lint:all` clean.

Two traps are documented in `forge-harness/.../AGENTS.md`: `GameState.applyToGame` defers through `GameAction.invoke`, so a caller already on the game thread sees nothing happen and has to use the protected `applyGameOnThread`; and the state file must declare as many players as the game has.


khaliostr:
Worth recording what this tool produced. Injecting the actual board from the 134s stall in #684 and timing the scan on it gave 60ms, against 9420ms observed in production. That number is what ruled the engine out and sent the investigation to the transport, where the cause turned out to be TCP retransmit timeouts. See #695.

Bots never build a board like that on demand, so without injection every estimate came from a board that was not the one in the incident.

khaliostr:
Superseded by #902. Its `GameCheckpoint` loads a Forge `GameState` into a running engine on the host path, with a sidecar for what the format drops, and works around the same `GameAction.invoke` deferral this branch documented. The benchScans fixture is not needed now that #924 stresses the wasm engine directly.

Keeping the one result this branch produced, for the record: injecting the actual board from the 134 s stall in #684 and timing the action-space scan gave 60 ms against 9420 ms observed in production. That ruled the engine out and sent the investigation to the transport (#695).

PR #691, opened 2026-08-17
perf(harness): stop re-running the static-ability pass per candidate ability

khaliostr:
## Summary

- `ActionSpace.getPossibleActions` brackets one enumeration pass with `setHoldCheckingStaticAbilities` and restores once, instead of letting `getAlternativeCosts` re-run the CR 613 layer pass twice per alternate-host ability.
- Three smaller costs on the same path: target enumeration now runs after the payability guards, the life-payment fallback returns early unless the cost has a shard it could pay with life, and the LKI restriction host is memoised per host card per pass.
- `Card.getAllPossibleAbilities` builds its survivor list in one pass; `ArrayList.removeAll` over another `ArrayList` is quadratic and `removeUnplayable` removes most entries.
- Sizes the engine JVM explicitly and adds `manabrew_node_decision_latency_seconds`.

Contributes to #684. **It is not the fix for it**, see below.

## Why

The scan re-runs on every priority prompt and on every mana undo. Instrumented, on a board with 60 modal double-faced cards:

| | before | after |
| --- | --- | --- |
| per scan | 24.94ms | **4.88ms** |
| `checkStaticAbilities` per scan | 35.2 | 8.5 |
| time in those passes | 24.01ms | 5.46ms |
| `getAllPossibleAbilities` | 23.04ms | 3.11ms |

The in-scan layer recomputations drop from about 34 to 1. The residual calls are real resolution work outside the scan.

On the JVM side, `SubprocessBridge::spawn` passed four `-D` flags and nothing else, so about 30 engine JVMs each sized heap and GC threads from the whole machine. Their own `gc,init` line says `CPUs: 10 total` and `Memory: 32768M`. New knobs: `SELF_HOSTED_NODE_JAVA_HEAP_MB` (1024), `_ACTIVE_PROCESSORS` (2), `_GC_LOG` (`stderr` or a directory), `_OPTS`. `-XX:+ExitOnOutOfMemoryError` is always set, which the endstep patches already assumed.

`manabrew_node_decision_latency_seconds` times a hosted player's answer reaching the engine until the engine asks for the next decision, on the existing push-gateway path. The production fleet ships no logs, so this is the only fleet-wide view of engine time. Staging's node gets GC logging to stderr, which alloy already ships to Loki, and a `mem_limit` that fits the heap.

## This does not explain #684

Measured on the actual stalled board from the capture, reconstructed with #690 and timed directly: **60ms per scan**, against 9420ms observed in production. The engine accounts for well under 1%. Details and the eliminated hypotheses are in #684.

So this is an optimisation of a real superlinear cost, worth having, and the JVM sizing plus the metric are what that issue actually needs next.

## Test plan

- Parity decision logs byte-identical before and after: 6 games, 4314 decisions (token_swarm ×2, comprehensive_test, trigger_expanded, planeswalkers_spaceships, keyword_advanced).
- Plus 3 games on decks made entirely of modal double-faced cards, 1393 decisions, byte-identical. Those exercise the path the hold suppresses, which the standard parity decks never touch.
- `yarn lint:all` clean. `cargo clippy -p self-hosted-node -D warnings` clean with no features, `--features java-forge`, `--features graal-forge`.
- JVM flags verified end to end: spawn line carries them and `gc,init` reaches the node log with `SELF_HOSTED_NODE_JAVA_GC_LOG=stderr`.
- `docker compose -f compose.staging.yml --profile hosted-ai config` renders the node service.

Submodule: `forge` moves to `khaliostr/actionspace-static-hold` on the fork, which makes `setHoldCheckingStaticAbilities` public and fixes the quadratic filter. Merge that into `manabrew` and `manabrew-staging` together when this lands.


khaliostr:
Reframing this PR now that #684 has a root cause in #695.

The stalls in #684 are TCP retransmit timeouts on the link between the node fleet and the relay. Latency there is `k * RTO + ~130ms`, with the observed 134s stall being 254 timeouts. No engine change can move that term.

What this PR does is still real and still worth landing. Holding the static ability pass across the enumeration cuts the action space scan from 24.94ms to 4.88ms, about 34 layer passes down to one. Treat it as a scan cost improvement, not a latency fix, and do not expect it to change the hosted numbers.

anacletos (dismissed):
I think must be rebased with main. Also some comments a bit confusing

khaliostr:
Updated against main (merge, not rebase) now that #701 has landed, and reconciled with it.

**The conflict and how it is resolved.** #701 added `-Xlog:gc:stderr` unconditionally so its metrics would work out of the box. This PR makes the GC log configurable, so the hardcoded flag is dropped and `SELF_HOSTED_NODE_JAVA_GC_LOG` now **defaults to `stderr`**. Behaviour is unchanged for anyone who sets nothing, and the metrics #701 exports keep flowing.

Also added two parser tests, because this PR asks for a richer log format than #701 was tested against (`-Xlog:gc*:stderr:time,uptime,level,tags` rather than plain `-Xlog:gc`): one for a real production line with full decorators, one asserting the extra `gc,heap` and `gc,metaspace` lines that `gc*` emits are not mistaken for pauses. That is a silent-breakage risk otherwise.

## Sizing, measured on the fleet host

Numbers from the box, which change what heap value to pick:

```
32 cores, 62GB RAM (33GB used, 28GB available)
32 java processes, 19.5GB total RSS, mean 623MB, max 868MB
```

So each JVM costs roughly 110MB beyond its 512MB heap. That rules out the obvious move: 32 JVMs at 1536MB would be about 53GB of JVM on a 62GB box.

| config | JVMs | RAM | headroom per game after the ~400MB card database |
| --- | --- | --- | --- |
| today | 32 x 512m | 20GB | ~100MB |
| `GAMES_PER_JVM=2`, 2GB | 16 | ~34GB | **~825MB** |
| `GAMES_PER_JVM=4`, 4GB | 8 | ~33GB | ~925MB |

`GAMES_PER_JVM=2` gives about eight times today's headroom at half the blast radius of 4. Suggested deployment:

```
SELF_HOSTED_NODE_MAX_GAMES=2
SELF_HOSTED_NODE_GAMES_PER_JVM=2
SELF_HOSTED_NODE_JAVA_HEAP_MB=2048
```

Sharing is not merely more efficient here. With 62GB the box cannot give 32 separate JVMs a heap large enough to hold the card database and a four player game, so consolidating is the only way to get real headroom. See #703 for why the floor exists and #706 for bounding one runaway game's effect on the games sharing its JVM.

## Merge prerequisite

This PR moves the `forge` submodule pointer. The fork branch `khaliostr/actionspace-static-hold` must be merged into **both** `manabrew` and `manabrew-staging` when this lands, or staging will fail to build.

anacletos (approved):

PR #694, opened 2026-08-18
fix(harness): remove blocking action trace

fedepoi:
## Summary

- remove the unconditional synchronous `/tmp/harness-trace.log` write before every Forge action is enqueued
- expose low-cardinality timing stages for bridge locking, action submission, next-prompt production, total decision latency, and snapshot fan-out
- move node websocket writes onto a dedicated bounded writer task and explicitly keep production/staging at one game per JVM

## Why

Issue #684 records hosted-game decisions taking 9–134 seconds while the reconstructed action-space scan takes about 60ms. The harness still performed an obsolete open/write/close cycle before `actions.offer`, exactly inside the unmeasured boundary. Even if that file write is not the complete cause, it is blocking production I/O on the critical path and provides no consumed output.

The stage metric makes a recurrence attributable without restoring per-action file logging. Separating websocket writes prevents state fan-out from directly occupying the node's inbound read loop. Explicit JVM isolation removes cross-game bridge contention. This complements the JVM sizing and ActionSpace work in #691 rather than duplicating it.

Contributes to #684.

## Test plan

- `yarn build:harness`
- `cargo check -p self-hosted-node --features java-forge`
- `cargo check -p networking-tests --tests`
- `cargo clippy -p self-hosted-node --features java-forge -- -D warnings`
- `cargo clippy -p networking-tests --tests -- -D warnings`
- `cargo test -p networking-tests --test networking_regression -- --ignored --nocapture --test-threads=1 brief_disconnect_reclaims_seat`
- `yarn lint:all`
- verified no `MANABREW_HARNESS_TRACE` / `harness-trace` references remain and a Forge-backed run does not create `/tmp/harness-trace.log`
- Forge-backed reconnect diagnostic reached game start, prompt handling, disconnect, reconnect, resync, and another answered prompt; the existing Rust-oriented scenario later timed out waiting for its final post-grace prompt, so it is not claimed as a passing Forge regression


khaliostr:
Both of your structural criticisms are correct. One of them invalidates a number I published, and I have measured the quantity your analysis left open.

## The trace costs about 180us per action

Timed it directly on the Forge path: `System.nanoTime()` either side of the `trace` call in `submitAction`, three-player Commander with the real decks from the incident, 400 actions.

```
ERROR-PROBE actions=400 trace_total_us=81602 last_trace_us=184
```

**204us mean, 184us at the end, flat across the run.** No growth, which matches your reasoning: append mode makes the open O(1) and close only flushes to the page cache.

For context, on the same run the whole JVM-side window from `submitAction` entry to the next prompt being published averaged **10.1ms**. So the trace is about 2% of it on a warm local SSD.

That supports your qualification rather than the defect being the cause. 180us needs to become ~50,000x worse to reach 134s. It would take genuinely pathological host I/O, and as you say the phase distribution does not fit writing the same pass line either. Removing it is still obviously right: unbounded synchronous I/O on the decision-delivery path with no consumer.

## You are right that my networking-tests measurement was invalid

I ran the local probe without `SELF_HOSTED_NODE_ENGINE_BACKEND=forge`, so it exercised the Manabrew Rust engine and never touched the subprocess, the bridge, or the trace. The "3.2ms locally against 150ms in production" comparison I posted above is therefore meaningless and I retract it. Good catch, and it is exactly the sort of default that makes a green test prove nothing.

Re-running the same shape with the Forge backend gives the 10.1ms above, which is a real number for that boundary.

## You are also right that my 60ms overreaches

The benchmark brackets `ActionSpace.getPossibleActions` only. `PhaseHandler.checkStateBasedEffects()` runs before `chooseSpellAbilityToPlay` and covers a static-ability pass and trigger processing, none of which my bracket sees. So the correct claim is narrower than the one I made: **ActionSpace is not the missing 9 to 134 seconds**, and I should not have generalised that to all rules-engine work. I will correct it in the issue.

The 10.1ms `submitAction` to prompt figure is the better bound, since it does include state-based effects, triggers and the scan. Still nowhere near seconds, but measured at the right boundary this time.

## Overlap with #691

Your stage metric is strictly better than the single gauge I added, so I have removed mine (`2a7f546e3`). #691 now keeps only what you do not cover: explicit JVM heap and `ActiveProcessorCount` sizing, `-Xlog:gc*` to stderr so a containerised node's GC log reaches Loki, the staging `mem_limit` raised to fit the heap, and the ActionSpace and `getAllPossibleAbilities` work.

We still both touch `compose.production.yml`, `compose.staging.yml` and `docs/OBSERVABILITY.md`, but the hunks are disjoint now. Happy to rebase on yours once #694 lands, or to land mine first if that is easier for you.

## On your ranking

Agreed, and I would add one thing from the capture side. The strongest constraint on any remaining hypothesis is that **board size does not predict latency at all**: on node `b4070326`, a 302-permanent game ran at 0.18s median and a 2-permanent game at 4.28s, same node, same day. Whatever the mechanism is, it is close to independent of how much game there is to compute, which fits a blocking or scheduling explanation far better than a rules-work one. Your stage split should separate those on the first hosted game that stalls.

khaliostr:
**#684 reproduces locally.** Your backend catch is what unlocked it. Same probe, same machine, one variable:

```
SELF_HOSTED_NODE_ENGINE_BACKEND=manabrew   median=3.3ms    p90=24.6ms   max=130.6ms
SELF_HOSTED_NODE_ENGINE_BACKEND=forge      median=112.3ms  p90=115.0ms  max=6960.1ms
```

Two-player hosted game, basic decks (40 lands, 20 creatures), trivial board, on a laptop with no network between relay and node. A **7 second stall** and a **34x median penalty**.

The metric is #684's exactly: time from broadcasting our response to the next relay message of any kind. I also measured time to our own next prompt (which spans the other seat) and the two agree at the max, so the stall is the node's first output, not another seat waiting.

## Recipe

`networking-tests` with `Sim::spawn` + `Client`, plus:

```
SELF_HOSTED_NODE_ENGINE_BACKEND=forge
SELF_HOSTED_NODE_FORGE_HARNESS_JAR=<fat jar>
SELF_HOSTED_NODE_FORGE_ASSETS_DIR=<repo>/forge/forge-gui
REGRESSION_NODE_BIN=<target/release/self-hosted-node built --features java-forge>
```

The support harness needs the backend env plumbed through `spawn_node`; without it every existing test runs the Rust engine, which is why none of them ever showed this.

## What I have already excluded, in this local reproduction

- **No bridge call is slow.** Timed every `SubprocessBridge::call`; nothing above 50ms during a 7.4s stall.
- **Not the JVM game thread.** Timed `actions.offer` to `actions.poll` returning, and that take to the next `publishAgentPrompt`. Neither exceeded 200ms during a 6.1s stall.
- **Not the node's async loop**, at least in production: heartbeat intervals across the whole incident game are min 29.962s, median 30.000s, **max 30.039s**, with 52 landing inside stalls and none late. `MissedTickBehavior::Delay` means a one second block would show as a ~31s interval, so that loop was never blocked beyond 39ms. That also clears the outbound state fan-out and `handle_server_message`.
- **Not Forge stdout flooding the bridge.** `Main` redirects `System.out` to a sink in server mode, so the bounded stdout channel cannot fill from engine chatter.
- **Not the trace.** ~180us per action, measured.

So the time is somewhere between the JVM publishing a prompt and the node emitting its first envelope, and it is invisible to per-call timing on both sides. The 34x median gap is the more tractable half and is present at p90, not just in the tail.

Your `manabrew_node_forge_decision_stage_seconds` stages are the right cut. If you point them at this reproduction rather than production you get a feedback loop in minutes, and `snapshots` plus `next_prompt` should immediately show which side of the boundary the 112ms sits on.

Happy to push the probe as a scratch test if useful, or leave it out of the PRs and hand you the diff.

khaliostr:
Correction to my comment above: **the multi-second stall does not reproduce.** I was wrong.

Logging per-decision instead of aggregates shows the 7s is always **decision 0** and nothing else:

```
[probe] decision 0: NEXT-OUTPUT 6755ms
[probe] NEXT-OUTPUT n=50 median=112.3ms p90=113.1ms max=6754.6ms
```

Seven runs, one spike each, always the first decision, always 6.1–7.6s. That is Forge/JVM start-up cost, not the #684 tail. Reporting only the max hid it, which is my mistake and the second time today I have drawn a conclusion from a summary statistic instead of the distribution.

## What does reproduce, and it is worth having

The **constant Forge penalty**, present at p90 and not a tail at all:

```
backend=manabrew   median=3.3ms     p90=24.6ms    (excluding decision 0)
backend=forge      median=112.3ms   p90=113.1ms
```

34x, on every decision, on a laptop with no network between relay and node.

The likely cause is structural rather than mysterious. `run_hosted_engine_game_inner` polls:

```rust
loop {
    // drain client responses -> submit_action
    if let Some(prompt_json) = engine.get_prompt(&session_id, 0)? { ... }
    if engine.is_game_over(&session_id)? { ... }
    std::thread::sleep(Duration::from_millis(50));
}
```

Each decision pays that sleep about twice: once before the loop notices the client's response, once before it notices the resulting prompt. Two half-intervals average ~50ms, and the measured 112ms is consistent with that plus the snapshot fan-out. The Rust backend is in-process and event-driven, which is why it sits at 3.3ms.

That also matches production: the fleet-wide baseline on small boards is ~150ms, which is ~100ms of polling plus the relay-to-node round trip. So this explains the **baseline**, and I think it is worth fixing on its own (condition variable or a blocking read rather than a 50ms poll), but it is not the 9 to 134 second tail.

## What I have now excluded for the tail, in the local rig

During the stalls I did see, with probes on both sides:

- no `SubprocessBridge::call` above 50ms
- `actions.offer` to `actions.poll` returning: never above 200ms
- that take to the next `publishAgentPrompt`: never above 200ms
- `get_prompt` never returned `None` for more than 1s
- the prompt-id dedup never held one id for more than 1s

So on the reproduction I have, nothing on either side of the bridge is slow. The tail remains unreproduced locally, and I would not claim otherwise again.

Your staged metric in production is still the thing that settles it, and the poll interval above is a separate, cheap, provable win it will also make visible.

khaliostr (dismissed):

Issue #696, opened 2026-08-18
send state deltas instead of a full board per viewer per decision


khaliostr:
The node sends a full game state to every viewer on every decision. This is the largest single contributor to hosted latency. Root cause analysis in #695.

## The numbers

One state is 30-46KB. The node sends one per viewer, so a four seat game moves about 184KB per decision. Game `a7ec7e17` sent 2147 states at a 29.6KB median, roughly 63MB for one game.

At a 1448 byte MSS that is about 120 packets per state. Play is bursty, so `tcp_slow_start_after_idle` resets the congestion window to about 14.5KB before each decision, and 106 of 118 measured peers sit at `cwnd:10`. A 46KB message therefore needs roughly three round trips to leave the machine. At the measured 130-260ms RTT that alone is 0.4-0.8s per decision, before any packet loss.

Sending a delta instead takes a decision from about 184KB to about 2KB. That fits inside the initial congestion window, which removes the multi round trip push completely, and it cuts packets per decision from about 120 to about 2, which cuts loss exposure by the same factor. Median and tail both improve from one change.

## Two problems, not one

**Duplication across viewers.** Four near identical copies of mostly public information go out per decision. A shared public view plus a small private overlay per player removes most of that on its own, before any diffing.

**Retransmitting unchanged state.** One card changes zone and we ship the whole board.

Either alone is a large win. Both together are the target.

## Design sketch

Not a decision, a starting point.

- Every state carries a monotonic `stateVersion` per game.
- The node sends `stateDelta` messages carrying a patch from version `n` to `n+1`, plus the version pair.
- A client that receives a delta whose base version it does not hold asks for a full state and the node sends one.
- Full states are still sent on join, on reconnect, and on request. The full state path stays as the fallback and is what a delta is validated against.
- Split the payload into a public view shared by all seats and a private overlay per seat, so the public part is diffed once rather than per viewer.

## Open questions

- Patch format. A typed diff over the existing view structs, or a generic JSON patch. A typed diff is smaller and checkable at compile time. A generic one is less code.
- Where the diff is computed. The node holds both versions, so it can diff there. The alternative is to have the engine emit change events directly, which is more accurate but touches more of the harness.
- Resync policy. How far behind a client may fall before it gets a full state rather than a chain of deltas.
- Whether the observer seat needs the same treatment.

## Acceptance

- Median bytes per decision under 4KB for a four seat Commander game with a developed board.
- A client that misses a delta recovers without a visible break.
[...]

Issue #695, opened 2026-08-18
hosted stalls are TCP retransmit timeouts on the node to relay link


khaliostr:
Hosted prompt stalls are not engine time. They are TCP retransmit timeouts on the link between the node fleet and the relay. This issue records the evidence and the work that follows from it. It is the root cause for #684.

## The finding

Stalled decisions are not one long computation. They are a whole number of identical waits:

```
latency = k * T + ~130ms of real work
```

`T` is constant within a game and differs between games:

| game | n | best period | Rayleigh R |
| --- | --- | --- | --- |
| a7ec7e17 | 452 | 530.8ms | 0.80 |
| c1b1fa63 | 253 | 523.2ms | 0.86 |
| bdcb62c8 (fast) | 314 | no lock | 0.28 |

The 134s stall is `k` = 254. A code constant would be the same in every game. A per-game constant that lands between 520ms and 531ms is the retransmit timeout of that TCP connection.

## Evidence

All from the 2026-08-10 relay capture unless stated.

**The stall is one leg of the round trip.** Response to next state: median 1249ms, p90 10750ms, max 134295ms. That state to prompt: median 80ms.

**The node sends nothing during a stall.** Zero errors, zero retries, zero extra responses. One `unknownActionId` in the whole game.

**Stalls are per room, never fleet wide.** During a7ec7e17's 107 stalls, decisions in other games ran at median 0.15s and p90 0.62s. Their all day numbers are 0.18s and 1.02s. No shared resource can stall one room and leave two to four neighbours untouched. This rules out box CPU, GC, the relay, disk and uplink saturation together.

**Two plateaus that do not depend on payload size.** Consecutive state to state gaps, n=140837 across 155 games:

| state size | n | median gap | p90 | p99 |
| --- | --- | --- | --- | --- |
| 0-10KB | 10518 | 2ms | 37ms | 42ms |
| 20-30KB | 35945 | 5ms | 38ms | 459ms |
| 40-50KB | 13521 | 8ms | 38ms | 486ms |
| 90-100KB | 1054 | 33ms | 42ms | 426ms |

The median moves 2ms to 33ms across a tenfold size range, so serialisation is not the cost. p90 pins at 37-42ms and p99 at 412-486ms in every bucket. 40ms is the delayed ACK timer. 200ms is the kernel minimum retransmit timeout, and 400ms and up is that timer after backoff.

[...]

PR #698, opened 2026-08-18
feat(node): send state patches instead of a full board per viewer

khaliostr:
## Summary

- Add `state_delta`: a patch format that keys zone entries by `zone`/`ownerId` and cards by their existing `id`, so moving or tapping one card does not resend the zone.
- The node can send `stateDelta` against the last state it sent that seat, behind `SELF_HOSTED_NODE_STATE_DELTA`, off by default.
- Full state envelopes now carry an optional `fingerprint`. A patch names the fingerprint it applies to, so a receiver holding something else can tell.
- Client applies patches at the socket edge and emits an ordinary `state` envelope, so nothing downstream changes.

## Why

Closes part of #696, and the byte half of #695.

A decision sent one 30-46KB state to every viewer, about 184KB for four seats. At a 1448 byte MSS that is roughly 120 packets per state. Play is bursty, so `tcp_slow_start_after_idle` resets the congestion window before each decision and a 46KB message needs about three round trips to leave the machine. At the 130-260ms RTT measured in #695 that is 0.4-0.8s per decision before any packet loss.

Measured on two recorded production games:

| | median | p90 | p99 | total per game |
| --- | --- | --- | --- | --- |
| full state | 30720B | 46365B | 52080B | 62.6MB |
| patch | 63B | 1106B | 2720B | 1.0MB |

A decision drops from about 184KB to roughly 250B at the median, which fits inside the initial congestion window.

Two things were measured and rejected rather than built. Dropping static card data (`text` and `identity` are 56% of card bytes) only moves 65x to 77x and would break copy and layer effects. Splitting public from private views earns very little once patches are this small.

## Test plan

- **8677 consecutive states from two recorded production games round-trip exactly.** `recorded_states_round_trip` is gated on `STATE_DELTA_FIXTURE`; the capture holds player data so no fixture is checked in. Run: `full 372.9MB, delta 3.2MB, 115x smaller, worst patch 9681B`.
- **511 patches generated by the Rust differ were applied by the TypeScript applier and compared against the expected state. All 511 matched.** This is the check that matters, because the differ and the applier are in different languages.
- Unit tests both sides for the traps: real `null` fields surviving, explicit removals, reordering, arrays without stable keys, and an object replacing a scalar.
- `cargo test -p networking-tests -- --ignored`: 12 passed with the flag off and 12 with it on. That suite covers room lifecycle rather than state contents, so read it as "the node still works", not as evidence about patch correctness.
- `yarn lint`, `yarn lint:rust`, `tsc` clean.

## Notes for review

**Receivers never compute a fingerprint, they only compare what the sender declared.** Making both sides hash would mean Rust and TypeScript agreeing on canonical JSON down to number formatting, which is a poor thing to depend on. The cost is that this detects a patch applied to the wrong base, not an applier bug; those are covered by the cross-language check above.

**On a base mismatch the client drops the patch and keeps the board it has**, rather than showing a wrong one. It recovers on the next full state. There is no explicit resync request yet; adding one is a protocol change and wants its own pass. Worth a view on whether dropping is the right behaviour.

The snapshot used for resume stores the fingerprinted full state, so a client that reconnects lands on a state later patches can attach to.

Stacked on #697, which touches the same file. Merge that first.


khaliostr:
The early deploy gate does not catch this, and that turns out to be a hole worth fixing on its own.

`GATE_CRATES` in `xtask/src/deploy.rs` watched `manabrew-protocol`, `manabrew-relay-protocol`, `manabrew-server` and `manabrew-hub`. `StateEnvelope` lives in `manabrew-agent-interface`, which is none of those. `protocol.rs` says as much: the wire types moved to `manabrew-relay-protocol` and `StateEnvelope` stayed behind because it carries engine DTOs.

`cargo xtask plan` for this branch:

```
manabrew-agent-interface   0.7.1  -> 0.8.0   (minor)
manabrew-server            0.10.1 -> 0.10.2  (patch)
manabrew-protocol          not bumped
manabrew-relay-protocol    not bumped
manabrew-hub               not bumped
```

A 0.x minor bump is already incompatible by the gate's own rule, but no gated crate moves incompatibly, so the release would evaluate `early=true` and ship the web stack ahead of the installers.

Added `manabrew-agent-interface` to `GATE_CRATES`. Simulated against the released manifest, this release now takes the deploy-last path:

```
manabrew-agent-interface   0.7.1 -> 0.8.0   INCOMPATIBLE -> forces deploy-last
```

The hole predates this branch and applies to any envelope change, so it is worth a look independently of the rest of this PR.

It does not remove the need for `SELF_HOSTED_NODE_STATE_DELTA`. The gate orders a deploy so an installer exists before the relay rolls; it cannot make anyone install it. And the failure mode here is silent rather than loud: an old client's `switch (envelope.kind)` has no `stateDelta` case, so the message is dropped and the board simply stops updating, with no error anywhere. The flag is what lets the emission be turned on deliberately once installs have caught up, and off again in one restart.

PR #700, opened 2026-08-18
feat: account niceties and minor fixes

anacletos:
## Summary
* UI overhaul of the account section and navbar
* Handle = username - A signed-in account plays under its claimed handle everywhere; the separate, editable username in server settings is gone.
- Hub refuses to hand out any name already claimed as a handle
- Guests can't take claimed names 
- The in-game name comes from a Hub-signed token
- UI niceties:
- dev niceties to test accounts
- deduplicate same useraname accounts on the multiplayer tab

## Screenshots
Screenshot 2026-08-18 at 20 41 17
Screenshot 2026-08-18 at 20 42 05
Screenshot 2026-08-18 at 20 42 16


manabrew-rs/crates/manabrew-hub/src/dto.rs
```diff
@@ -571,6 +468,14 @@ pub struct TokenRequest {
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, TS)]
+#[serde(rename_all = "snake_case")]
+#[ts(export, export_to = "authTypes.ts")]
+pub struct GuestTokenRequest {
+    pub name: String,
+    pub guest_id: String,
+}
+
```
khaliostr, line 581: I am not convinced we should do this. The original idea was to make the 'hub' _optional_ to attest possession of an user name, with guest flow not need the hub and its deployment. this way we tie _every_ game to the presence and deployment of the hub

khaliostr (changes requested):
Looking good, but I am not convinced on the guest token. Relay should be viable in isolation and not depend on the hub for operation. So attested identity should be an upgrade rather than a requirement

khaliostr (dismissed):
on second thought, approved as self-hosted does not depend on the hub

khaliostr:
Hey @anacletos, i thought a bit about this. To move the hub off the relay hot path, could we arrrange it so that when teh _frontend_ tries to set a handle/username the _hub_ checks it against a claim table. so that if a user previously claimed the handle it refuses to give it away. this way we guarantee that only "approved" names can ever getinto the relay but moving the check off the relay code

khaliostr (approved):
very nice work. thanks!

Issue #626, opened 2026-08-05
Data protection follow-ups


khaliostr:
Accounts and server-side decks are live, so the service now stores personal data. This issue tracks the data-protection work.

Blockers:

- [ ] #622 Account deletion and data export
- [ ] #623 Data retention sweeper

High:

- [ ] #624 Privacy policy: update it and link it

Small:

- [ ] Drop the client IP from the relay connection-error log (`server.rs`). It ships to Loki.
- [ ] Add auth checks inside the hub `/admin/*` handlers. Today only ingress basic auth protects them.
- [ ] Add an age statement to sign-up or to the terms.
- [ ] Fix `docs/OBSERVABILITY.md`: Alloy ships logs from all containers, not only relay and node.
- [ ] Remove the stale `compose.production.yml` comment about logging magic-link codes.
- [ ] Add a CSP for web and Tauri.

Decision: game captures stay, under the existing 20 GB size cap. An erasure request scrubs the username. The games stay.

fedepoi:
@anacletos this was all fixed in your pr right ??
canwe close this ?

Issue #622, opened 2026-08-05
Account deletion and data export


khaliostr:
Accounts store personal data with no way to delete or export it.

Current state:

- No delete-account endpoint. `storage.rs` has no `delete_account`.
- `decks.account_id` is `ON DELETE RESTRICT`. Even a manual `DELETE FROM accounts` fails once the user owns a deck.
- Deck deletion is soft only. `deck_versions.snapshot_json` keeps the full contents forever.
- `GET /api/auth/me` returns a small subset. No export covers sessions, decks, publications, or play history.

Tasks:

- [ ] `DELETE /api/auth/me`: delete the account with its sessions, identities, decks, versions, and favorites.
- [ ] Rework the RESTRICT foreign keys so account deletion can cascade.
- [ ] Decide the fate of published entries on deletion: delete them, or anonymize the author.
- [ ] `GET /api/auth/export`: one JSON file with account, identities, decks, versions, publications, and play reports.
- [ ] Settings UI: delete button with confirmation, export button.
- [ ] Operator script that scrubs a username from relay analytics (`events.db`, `events-*.jsonl`) and game captures. The games stay. Only the name goes.
- [ ] Update the privacy page.

GDPR art. 17 gives us one month to answer an erasure request. A self-serve flow is the cheapest way to comply.

PR #713, opened 2026-08-18
feat(observability): expand product analytics dashboards

fedepoi:
## Summary

- materialize privacy-safe Hub aggregates into the existing analytics SQLite database
- add curated Executive Health, Engagement & Gameplay, and Decks & Collections dashboards while retaining the broad Analytics Explorer
- classify relay clients as web, PWA, desktop, or mobile and visualize distinct platform usage
- add an optional local Grafana profile for end-to-end dashboard testing

## Why

Grafana previously saw only relay gameplay tables and exposed a mostly all-time panel collection. Product, auth, deck, publication, collection, and ranking data lived exclusively in the operational Hub database. This keeps Grafana isolated from core and sensitive data while providing curated decision-making and diagnostic views with consistent time semantics.

## Test plan

- `yarn lint:all`
- `docker compose -f compose.dev.yaml --profile observability config --quiet`
- `MANABREW_SERVER_KEY=validation GRAFANA_ADMIN_PASSWORD=validation docker compose -f compose.production.yml --profile observability config --quiet`
- ran every SQLite panel and template-variable query against a fresh exported analytics database
- verified idempotent repeated exports, all 20 Hub table coverage rows, and `PRAGMA integrity_check`
- `cargo check -p manabrew-server -p manabot -p self-hosted-node -p networking-tests`
- `yarn typecheck`

## Demo

Started local Grafana with `docker compose -f compose.dev.yaml --profile observability up -d events-ingester grafana`, logged in at `http://localhost:3000`, and visually verified the provisioned Executive Health dashboard, navigation, sections, and SQLite-backed panels.


khaliostr:
Really like this one, the privacy design and the `#[serde(default)]` rollout safety are both right. Blocking: `networking-tests/tests/support/mod.rs:257` uses `manabrew_relay_protocol::ClientPlatform` but that crate is not a dev-dependency, so CI cannot compile it. Import `ClientPlatform` from `manabrew_agent_interface::protocol` like the rest of that file. I pushed that one line fix to your branch as 98a8025fe, hope that is ok, revert it if you would rather do it your way.

`scripts/ingest-events.py:325`: the `schema_version` `.fetchone()[0]` raises `TypeError`, which `except sqlite3.Error` does not catch, so the process exits and `restart: unless-stopped` makes it a loop that also stops relay ingest. Please broaden to `except Exception` and guard the `fetchone()`.

`scripts/ingest-events.py:329`: all 47 hub queries share one `try`, so one renamed table writes zero metrics and every panel goes stale. Wrap each query individually so a break degrades to one missing series, and a small test running the exporter against a freshly migrated hub.db would stop this drifting.

Smaller: `executive.json:370` (same in `product.json:1011`, `live-ops.json:621`) reads `hub_sync_state`, which is empty until the first successful export, so the staleness panel shows grey "No data" instead of red, needs a `coalesce(..., 0)`. And `ingest-events.py:408` uses `coalesce(v.format, '')` where the siblings use `'unknown'`, which double counts cards from NULL format versions.

Optional: `clientPlatform.ts:10` uses `window.matchMedia` unguarded although line 3 guards it, and `connection.rs:274` builds the label from `Debug` while the enum already declares `snake_case`. Separate follow up, not this PR: `client_connections` stores usernames unpruned, so the #622-#626 erasure work needs to reach it.

khaliostr (approved):

Issue #715, opened 2026-08-19
triage - Players Keep Disconnecting and Reconnecting


fedepoi:
Forge

This wasn't a problem last time we played. It keeps happening a lot. The players that disconnect don't seem like they're disconnected themselves.

### Engine
**Forge**

---
Reported by **devon7654** in the Issue Tracker Discord forum: https://discord.com/channels/1502665088304419008/1539441406836346970

khaliostr:
fixed by  #719

PR #719, opened 2026-08-19
fix(relay): answer the client keepalive so dead sockets are caught

khaliostr:
## Summary

- The relay answers `ClientMessage::Ping` with a new `ServerMessage::Pong` instead of dropping it.
- The web client pings every 4s and fails the socket after 10s of silence, going straight into the existing reconnect path.
- A throttled background tab re-baselines instead of tripping the watchdog.
- The heartbeat is kept out of the comms ring buffer that bug reports dump.

## Why

Relates to #715. Players reported opponents vanishing for long stretches mid game.

The relay logs show the cause. A middlebox on some players' path resets the connection roughly every 5 minutes, and the reset reaches only the relay. The client keeps writing into a socket that is already gone. In one 4 player Forge game the reconnect gaps were 21, 23, 39, 40, 47 and 48 seconds, with zero TCP connection attempts during the gaps. The client was not backing off, it had not noticed. Once a reconnect did start, session resume finished in 370ms.

Nothing on our side could detect this. WebSocket ping and pong frames are answered by the browser and never surface to page JS, so the app level `Ping` was the only usable signal, and the relay never replied to it.

This does not stop the resets. It cuts the time to notice one from 20 to 50 seconds down to about 12.

Additive on the wire. Only the web client sends `Ping`, so only the web client receives `Pong`, and older clients ignore unknown message types.

## Test plan

- Ping and pong round trip against a local `manabrew-server` build over a real WebSocket: `Authenticate` then `Ping` returns `Pong`.
- `WebServerApi` driven with a stub socket and fake timers:
  - 12 answered cycles keep one socket open and send exactly 12 pings.
  - Silence tears the socket down at 12s and opens a replacement.
  - A 60s clock jump delivering a single tick does not trip the watchdog, and the next cycle stays healthy.
- `yarn lint` and `yarn lint:rust` clean.

### DIFF
--- a/manabrew-rs/crates/manabrew-relay-protocol/src/lib.rs
+++ b/manabrew-rs/crates/manabrew-relay-protocol/src/lib.rs
@@ -150,6 +150,8 @@ pub enum ClientMessage {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(tag = "type")]
 pub enum ServerMessage {
+    Pong,
+
     AuthResult {
         success: bool,
         player_id: Option<String>,
--- a/manabrew-rs/crates/manabrew-server/src/connection.rs
+++ b/manabrew-rs/crates/manabrew-server/src/connection.rs
@@ -680,7 +680,7 @@ fn handle_client_message(
             );
         }
 
-        ClientMessage::Ping => {}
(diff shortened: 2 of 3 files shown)


anacletos (approved):
nice

PR #720, opened 2026-08-19
feat(node): give every deploy preview a GraalVM hosted-AI node

khaliostr:
## Summary

- Every staging deploy starts the hosted-AI node, previews included. The `branch == "staging"` gate in `deploy_staging` is gone.
- The node image compiles Forge with GraalVM `native-image` and links it in-process (`--features graal-forge`) instead of shipping a JRE and spawning one JVM per game. It runs no Java process at all.
- Every room attaches to one shared isolate, so the 34,000-card database loads once per node rather than once per game.
- Staging sizing: 4 concurrent games, `mem_limit: 1g`, `cpus: 1.0`, `cpu_shares: 256`.
- New CI job builds the Linux native library and smokes the shared isolate, so a broken native build fails a PR instead of a release.

## Why

A `deploy-preview` PR got web, relay and hub but no node, so the one thing hardest to review by reading a diff — actually playing a game — was the one thing a preview could not do. The reason was cost: the node was a Forge JVM per game on the box that also runs production.

GraalVM removes the cost. The card database is what the memory is (#684: ~450MB of `PaperCard`/`CardRules` against 240 live `Card` objects), and a shared isolate pays for it once. Measured on this branch, a node with the database loaded and a game session live is **534MB RSS, up in 6.3 seconds** — against roughly 600MB per JVM, four JVMs, each ~30s to warm.

The same change fixes something worse. `up --remove-orphans` does not reclaim a service whose profile is off — Compose treats a profile-gated service as defined, not orphaned — so the last staging-branch node was still running on the production box, **4 weeks old, version 0.6.0, 91.7% of one of two cores, 47h51m of accumulated CPU time**. The deploy comment claiming the profile-off container gets reclaimed was simply wrong. With the profile always on, the node is recreated on every deploy and cannot go stale.

Closes nothing on its own; makes #698 and the rest of the node-transport stack reviewable by playing them.

### Can the prod box take it

`root@37.27.253.153`, 2 vCPU, 3819MB RAM + 4GB swap, measured 2026-08-20.

**Memory: yes, with room.** 1942MB available with the orphan gone (1430MB before), memory pressure 0.00 across avg10/60/300. Containers use 1.56GB in total. The new node's `mem_limit: 1g` *replaces* the orphan's 2g, so the committed ceiling drops by 1GB; actual use is roughly flat at idle and rises a few hundred MB with four games actually running. 70% of the cgroup limit is baked in as the max heap at build time, so 1g means ~716MB of heap against a ~420MB live set — versus the 512MB heap that produced the death spiral in #684.

**CPU: it is the constraint, and this PR improves it.** Load average was 2.1 on two cores at `some avg10=16.49` CPU pressure; with the stuck orphan JVM gone it sits at 0.83 and 5.0. Essentially all of the load was that one process. The node is now capped at `cpus: 1.0` and given `cpu_shares: 256` against production's default 1024, so under contention it yields 4:1. Four Forge AI seats deciding at once will want more than one core and will simply run slower; production does not lose the core.

That orphan has since been removed by hand (`docker rm -f manabrew-staging-self-hosted-node-staging-1`), which is where the "after" numbers above come from, and the graal node has replaced it. The always-on profile is what stops it happening again.

## Test plan

Measured on this branch, macOS arm64, GraalVM CE 21.0.2:

- `./forge-harness/build-native.sh` against the branch's harness jar: **built clean**, `libforgeharness.dylib` 79.6MB, 14,052 reachable types, 2m2s wall, 6.2GB peak build heap. That peak is why the Docker stage passes `-J-Xmx8g`.
- Shared-isolate smoke on the release binary — `SELF_HOSTED_NODE_GRAAL_SMOKE=1 SELF_HOSTED_NODE_SHARED_ISOLATE=1`: `shared graal isolate initialized` → session started → completed. **6.27s wall, 534MB peak RSS** with the full card database resident.
- `-R:MaximumHeapSizePercent` confirmed against `native-image --expert-options-all`: it exists, defaults to 80, and applies to Serial and Epsilon — which is what GraalVM CE has.
- `cargo clippy -p xtask -D warnings`, `cargo fmt --check`, `yarn lint`, `yarn lint:rust`, prettier on every touched yaml/md: clean.
- Compose and both workflows parse.

CI is green, including the new `GraalVM node (Linux)` job — that was the real unknown, since the Linux amd64 native build had never run anywhere.

**Deployed to the staging slot and played against.** Everything that was open is now measured on the box:

| | result |
| --- | --- |
| Processes in the container | **one** — `self-hosted-node`. No `java`. |
| cgroup limit seen inside | `1073741824` — 1GiB, not the host's 3.8GB |
| Heap reservation with a game live | VSZ **1043MB** — sized from the cgroup. Had it read host RAM the cap would be 2.7GB and the container would OOM-kill at 1GiB. |
| RSS with a game live | **500MB / 1GiB**, no OOM, no restart |
| Rooms | 4 up and joinable |
| Idle RSS before the first game | **19.9MB** |

The game itself plays quickly — in-process Forge has no IPC hop.

**One real regression, fixed in this PR.** The first game took 45 seconds to appear:

```
22:45:30.127  starting hosted engine thread
22:46:14.790  shared graal isolate initialized   ← 44.7s
22:46:14.882  session started
```

That is the card database loading, and `init_engine` was a no-op on the graal path, so the isolate was built lazily by whoever asked first — a player. (The same load is 6.3s on a 10-core laptop; this box gives the node one core.) `init_engine` now builds the shared isolate at startup, so the cost lands at deploy time and rooms are not advertised until the engine can serve them. Games after the first always reused the isolate and paid nothing; now the first one doesn't either.

**The preview node now emits state patches** (`SELF_HOSTED_NODE_STATE_DELTA=1` in `compose.staging.yml`). The first deploy ran with the flag at its default, so #698's differ never executed and that game went over full states — the deploy validated #697's send path and that #698 is inert when off, nothing more. Staging can carry the flag where production cannot: patches need a client that can apply them, and the staging web bundle is built from the same ref as the node while installed desktop clients are not.

Still unmeasured: four games running at once on 1g. The floor and a single game are measured; the ceiling is not.

If the Linux native build turns out to be a problem, the fallback is two lines: `--features java-forge` in the Dockerfile and the Temurin runtime base back. The subprocess backend is untouched.

## Notes for review

**This changes the released `manabrew-node:latest` image too**, not only staging — self-hosters pulling it get the native engine and no JVM. `website/src/content/docs/self-hosting.mdx` is updated to match, including `SELF_HOSTED_NODE_SHARED_ISOLATE=0` for anyone who wants a per-game isolate back.

**The JVM GC metrics go quiet for this node.** `record_jvm_gc` (#701) is fed by the engine JVM's `-Xlog:gc` stderr pump. A native image has no such log, so `manabrew_node_jvm_gc_*` stays at zero on a graal node. The java-forge fleet on ripper is unaffected. `docs/agents/SELF_HOSTED_NODE.md` says so.

**Stacked on #698, which is stacked on #697.** The whole stack has been brought up to date with main by merge, not rebase, per `AGENTS.md`. main's `SELF_HOSTED_NODE_JAVA_*` block in `compose.staging.yml` conflicted with the node service rewritten here and was resolved as "take this branch's" — those knobs configure a subprocess JVM this image no longer has.

**The orphan is gone.** Removed by hand on the box while writing this. Load average dropped 2.1 → 0.83, CPU pressure `some avg10` 16.5 → 5.0, available memory 1418MB → 1942MB, idle 29% → 64%.


khaliostr:
## Verified on the staging slot

Deployed to the preview slot on the production box and played against. Numbers are measured on that box, not projected, unless said otherwise.

**The node.** One process in the container, `self-hosted-node`, no `java`. cgroup limit seen from inside is `1073741824`, so 1GiB and not the host's 3.8GB. With a game live, VSZ 1043MB: the 70% heap cap is sized from the cgroup. Had it read host RAM the cap would be 2.7GB and the container would be OOM-killed at its limit. Zero restarts, no OOM, across the whole session.

**Card database load moved to boot.**

```
23:03:29.105  initializing forge engine backend
23:03:57.758  shared graal isolate initialized      28.7s
23:03:57.982  starting room host
23:03:58.007  room created
```

The database is resident before the first room is advertised, so nobody can join a node that cannot serve. First game start to engine session is now **105ms**. Before this fix it was **44.7s**, paid by the player, because `init_engine` was a no-op on the graal path and the isolate was built lazily by whoever asked first.

**State patches, 11m03s game, one human viewer.**

| | bytes |
| --- | --- |
| Node container out, whole game | **1.58MB** |
| Node container in | 1.09MB |
| #696 fixture, patched | ~1.0MB/game |
| #696 fixture, full states | 62.6MB/game at four seats, ~15MB scaled to one viewer |

That 1.58MB includes relay chatter, heartbeats, room updates and the deck payload, so patch traffic is below it. An order of magnitude under full states, on top of the patched figure. No fingerprint mismatch, no dropped patch, no resync in the log. A client failing to apply patches shows as full-state fallback and there is none. This is the Rust differ, the relay replay cache and the TypeScript applier running against a real game rather than a fixture.

**Memory, and why `mem_limit` moved to 1536m.**

```
447MB   idle, database resident
658MB   one game, peak
```

One game adds about 212MB over the resident database. Four games project to roughly 1.3GB, which does not fit the 1GiB this PR originally set. RSS is not the live set and a tighter heap would compress rather than fail outright, but compressing is the #684 failure mode exactly: too little heap, back to back full GCs, 89% of wall clock stopped. 1536m gives about 1075MB of heap against that projection. The box has 1942MB available.

**Four concurrent games is still projection, not measurement.** One game and the idle floor are measured. The ceiling is not. That is the remaining risk in this PR.

**Also fixed here.** Staging had no `MANABREW_GAME_CAPTURE_DIR`, so the only instrument for the patch work above was container byte counters. It has one now, capped at 2GB since the disk is production's too. And alloy labelled every container `host="production"`, staging included; containers now carry an `env` label from the compose project. Node logs already reached Loki, they were just indistinguishable from production's.

Note the alloy change lands on the next production deploy, or `cargo xtask deploy --only alloy`, since prod's alloy does the shipping for both stacks.

## The orphan

`up --remove-orphans` does not reclaim a profile-gated service whose profile is off. Compose treats it as defined, not orphaned. The staging node from 21 July was still running four weeks later at v0.6.0, holding 91.7% of one of two cores with 47h51m of accumulated CPU. It was the entire load on the box.

| | before | after |
| --- | --- | --- |
| load average | 2.10 | 0.83 |
| CPU pressure some avg10 | 16.49 | 5.04 |
| available memory | 1418MB | 1942MB |
| idle | 29% | 64% |

Removed by hand. The always-on profile is what stops it recurring, since the node is recreated on every deploy. Nothing alerted on this for four weeks. A check that flags any running container older than the last deploy would have caught it and is worth filing separately.

fedepoi (approved):

PR #735, opened 2026-08-22
fix(node): never exit a stale node while a room is live

### LINKED ISSUE #734: node: publishing a release ends live games — the fleet exits mid-game and games die as host_lost
Publishing a release ends games that players are in the middle of. The whole fleet exits within the same second, systemd restarts it 30s later, and every game that was running dies. The relay records them as `host_lost`.

## What players got

Real seats, not load test ones:

```
18:04:49  53a0231c  seats=2  humans=['Mr EMOLH@8614']
23:46:44  7b6740c5  seats=4  humans=['damionking18']
23:47:31  e302f534  seats=4  humans=['damionking18']
23:53:41  36697f47  seats=4  humans=['damionking18']
23:54:15  5da66ecb  seats=4  humans=['damionking18']
23:54:46  8b15e9de  seats=4  humans=['damionking18']
```

Five games in eight minutes for one player, immediately after the 23:44 fleet bounce.

`host_lost` per day, and 2026-08-21 is the day with several releases:

| day | host_lost |
| --- | --- |
| 2026-08-21 | 12 |
| 2026-08-20 | 2 |
| 2026-08-19 | 4 |
| 2026-08-18 | 4 |
| 2026-08-17 | 3 |

For scale, the last three days hold 692 `game_over` against 22 `host_lost`.

## The sequence

Observed on the prod fleet with 16 load-test games live across four nodes, plus the player games above:

```
23:21:08  load starts, 16 games live
23:44:37  six of eight nodes vanish, all within the same 10s sample
23:45:17  six nodes back, new pids
23:46:08  the load run's games would have ended on their own
```

The nodes went down 91 seconds before the games they were hosting would have finished. From the journal:

```
23:44:30 manabrew-node@2.service: Control process exited, code=exited, status=1/FAILURE
23:44:30 manabrew-node@2.service: Failed with result 'exit-code'.
23:45:00 manabrew-node@2.service: Scheduled restart job, restart counter is at 3.
23:45:15 Started Manabrew stable rooms node #2 (docker, graal engine).
```

Nobody restarted anything by hand: "Scheduled restart job" with an incrementing counter is `Restart=always` reacting to the unit failing, and `systemctl restart` does not increment it. The same signature appears at 17:38:38 with the counter at 2, so this happens on every release.

The likely shape is that the node process exits, `docker run --rm` removes the container, and the unit's `ExecStop=docker stop -t 60` then fails against a container that is already gone. The FAILURE line is the cleanup, not the cause.

The same-second alignment across nodes follows from the previous bounce: they were all restarted together at 17:39, so their manifest polls have been in lockstep since.

## Why the cause is not yet pinned

`run_stale_monitor` in `updater.rs` is written to avoid exactly this. It exits only under `if is_idle()`, and otherwise logs `"self-hosted-node is stale but a game is in progress — deferring shutdown until idle"`. Which branch it took is unknowable after the fact: the line goes to the container's stdout and `docker run --rm` destroys it on exit.

So either `registry_idle()` reports idle while games are live, or the process leaves by some other path. Each room does push its `engine_session` slot into the registry at `host.rs:456`, and `registry_idle` requires every slot to be `None`, so the first case would mean the slot is not `Some` during a Forge game.

Note `sessions.iter().all(...)` is also `true` for an empty registry, which is worth checking on the paths where rooms have not registered yet.

## What needs to change

1. **Never exit with a live game.** Whatever the current path is, it did.
2. **Keep the logs of a node that exits.** A node that terminates itself throws away the only record of why. `--rm` plus the default json-file driver means there is nothing to read afterwards.
3. **Jitter the stale poll.** Even a correct drain bounces all eight nodes within seconds of each other, because they have been synchronised since the last bounce. Spreading them turns a fleet-wide outage into a rolling one.

Related: the node exposes `SIGUSR1` as a drain signal (refuse new games, close each room once idle) which would have been the safe path here.

khaliostr:
## Summary

- A room now counts as live while its `engine_session` is set **or** the relay's last `RoomUpdate` still reports it `InGame`. An empty or poisoned registry reads as unknown, never as idle.
- A stale node drains instead of cancelling its rooms: it stops taking new games, each room closes once it stops being live, and the process only leaves when nothing is left. The updater's own `exit(0)` is now a backstop that re-checks after the grace.
- The stale poll carries up to 25% jitter, so a fleet last restarted together stops polling in lockstep.
- Every stale check records `manabrew_node_stale_checks_total{decision="exit"|"defer"|"unknown"}` and `manabrew_node_stale_live_rooms`, and the decision log now carries the live-room count.

## Why

Fixes #734.

The updater exited only `if is_idle()`, and `is_idle` was `registry_idle`: every registered `engine_session` slot is `None`. Two things are wrong with that.

`sessions.iter().all(...)` is `true` for an empty registry. The registry is empty before any room registers, and `host_room`, the crate's public entry, never registers at all.

More importantly the engine session is `None` in every window where the relay still has a game in the room: before `maybe_start_hosted_engine` runs, on each of its early returns (missing deck, engine disabled, unsupported backend, fewer than two players, node not in player order), and after `abort_engine_session` clears it when a `ResumeRoom` is rejected. In all of those the relay's room is `InGame` with players in it, and the node reported itself idle.

The exit path itself was also a kill, not a drain. `shutdown_rooms()` cancelled every room and then slept ten seconds without re-checking, so a room could take a `GameStarted` inside that window and be killed anyway.

The timing on the box fits the updater. The units last started 17:39:15, the poll is 300s, and `tokio::time::interval` fires its first tick immediately, so ticks land at 17:39:15 + k*300. k=73 is 23:44:15, plus the 10s grace is 23:44:25, and the journal logs the `ExecStop` failure at 23:44:30. That is also why the fleet moves as one: it has been in lockstep since the last common restart.

I could not prove from the code which of those branches the six nodes took, and I cannot recover it after the fact either, which is the second half of the issue. All the branches are closed, and the decision is now recorded in Prometheus rather than only in a container log that `--rm` throws away.

## Test plan

- `cargo fmt --all -- --check`, clean.
- `cargo clippy -p self-hosted-node -- -D warnings`, and again with `--features graal-forge` and `--features java-forge`. Clean.
- `cargo clippy -p networking-tests --tests -- -D warnings`, clean.
- New system test `publishing_a_release_never_ends_a_live_game` in `networking-tests`. It serves a real release manifest over TCP that advertises this node's own build until the test publishes a newer one, runs the node with `SELF_HOSTED_NODE_SHUTDOWN_ON_STALE=1` against it, starts a real game, publishes the release mid-game, and asserts the node is still alive and still serving prompts twenty seconds later, then that it leaves within sixty seconds of the room returning to the lobby.
- The assertion is load-bearing: with `registry_live_rooms` temporarily stubbed to return `Some(0)`, the test fails on "the node exited on top of a live game". Reverted before committing.
- Full suite green: `yarn test:networking`, 16 passed, 0 failed, 179s.

## Follow-ups (ops, not in this repo)

- The node containers run `docker run --rm` with the default json-file driver, so a node that exits destroys the only log of why. The metric added here covers the updater's decision, but nothing else it prints survives. Worth dropping `--rm` or shipping the container's output.
- Consider whether the units want `RestartSec` staggered per instance as well. The poll jitter de-synchronises the decision, but eight nodes that do go down together still come back together.


anacletos (changes requested):
mmh

anacletos (changes requested):
No I think this is allucinated.

If true, I'd like to be able to reproduce this myself. Which right now. I can't

khaliostr:
Ok fair enough, all the better. I did see games going down while testing but could be unrelated. Closing ticket and PR

Issue #736, opened 2026-08-22
upstream the AiCache game-scoping fix to Card-Forge


khaliostr:
`AiCache` is upstream code (Card-Forge #10149, March 2026) and the leak is upstream's, not ours. Our fork only made the map thread safe for concurrent games in `19-concurrent-safe-singleton-caches.patch`; the lifetime problem came with the original design.

The fix is merged on our fork as witchesofthehill/forge#1: entries live in a per game map, finished games are dropped when the cache is next used, and `clear()` takes the game to clear.

It applies upstream unchanged. Card-Forge's desktop client keeps one JVM across a whole session, so every game a player finishes stays reachable there too, with the same `[result, args...]` lists pinning `Game` and `Player`. Nobody notices because a desktop session plays a handful of games and the heap is large; a host running games back to back in one process does notice.

Sending it upstream is deferred, not dropped. When someone picks it up:

- the diff is five files in `forge-ai`, no API change outside `AiCache`
- the evidence is in #729: heap dump, histogram, and the reference path back to `AiCache.dataMap`
- worth mentioning the second issue in the same review, that the cache was global across concurrent games, so two games could collide on a key and one game's `clear()` wiped another's entries

Blocked on nothing. It needs someone to open it against Card-Forge and follow their review.

Issue #729, opened 2026-08-21
node: finished games are retained by the static AiCache, filling the isolate heap


khaliostr:
Graal nodes run to their 2GB cgroup cap during play and then sit in permanent reclaim. Two separate causes: heap sizing that does not account for native memory, and a smaller amount of genuine per-game retention.

Measured on the prod fleet 2026-08-21, 11h35m after the units started. The card database is the constant 66-67MB `file` column on every node, so none of this is the archive.

| pid | anon | file | swap | cgroup used | `memory.events max` |
| --- | --- | --- | --- | --- | --- |
| 2901668 | 1891MB | 67MB | 18MB | 1912MB | 103 |
| 2905032 | 1062MB | 66MB | 51MB | 1073MB | 364973 |
| 2900535 | 1040MB | 67MB | 32MB | 1053MB | 141 |
| 2900368 | 1003MB | 67MB | 173MB | 1018MB | 2277933 |
| 2900514 | 965MB | 67MB | 0 | 1008MB | 0 |
| 2900520 | 867MB | 67MB | 0 | 908MB | 0 |
| 2900407 | 853MB | 66MB | 0 | 893MB | 0 |
| 2900457 | 635MB | 66MB | 0 | 675MB | 0 |

A fresh node sits at 446MB. Games completed since boot run 10 to 32 per node.

## The memory sawtooths, it does not only climb

One client, one node, repeated 2.5 minute games, all eight nodes sampled every 10s. Anon, in MB:

```
pid 2900368  1299 1963 2033 2034 1237  971  980 1037    <- ran to the cap, then a full GC returned 1063MB
pid 2901668  1554 1383 1038 1814 1602 1439 1440 1681    <- oscillates about 900MB peak to trough
pid 2900535  1030 1030 1030 1030 1032 1034 1034 1034    <- idle, flat, peaked at 1913 earlier in the day
pid 2900514   935  935  935  935  935  935  935  935    <- idle, flat
pid 2900520   868  868  868  868  868  868  868  868    <- idle, flat
pid 2905032  1063 1063 1063 1064 1064 1064 1064 1064    <- idle, flat
```

Idle nodes do not move at all. Nothing grows with time or with connections, only with games. Threads are flat (34 to 42), so this is not a thread or session leak.

## 1. Heap sizing ignores native memory

[...]

PR #737, opened 2026-08-22
chore(forge): bump the submodule for the AiCache game-scoping fix

khaliostr:
## Summary

- Bumps the `forge` submodule to pick up witchesofthehill/forge#1.
- Nothing else changed in the fork between the old pointer and this one.

## Why

`AiCache` stored `[result, args...]` in one process wide multimap, and the arguments are game scoped (`Game`, `Player`). Every entry pinned the game it came from, and the only `clear()` caller ran mid turn, never at the end of a game. On a shared isolate that means finished games are never collected.

Measured on a prod node after about 20 games: 14 live `forge.game.Game` objects on a node whose `MAX_GAMES` is 4, with 28 `Player`s, 420 `PlayerZone`s and 8,485 `Card`s, while the card database stayed byte identical to a fresh node. Roughly 34MB of live heap per retained game. Full trace in #729.

The fork change keys the cache by game, drops finished games when the cache is next used, and makes `clear()` take the game to clear so one game no longer wipes a concurrent game's entries.

`AiCache` is upstream code (Card-Forge #10149) and upstream leaks the same way; sending it there is tracked in #736.

## Test plan

- `mvn -pl forge-ai -am -DskipTests package` and the full `forge-harness` jar build clean against this pointer.
- The submodule tracks the fork's `manabrew` branch, per `.gitmodules`; no tag pin.
- Java only change, no card data touched, so the cardset archive is unaffected.
- The real confirmation is after this deploys: dump a node that has played a while and check `forge.game.Game` stays at or below `MAX_GAMES`, with `notes/loadtest/hprof_histogram.py`. Not something this PR can prove.

## Follow-up

The fork's `manabrew-staging` branch is behind `manabrew` after this merge and should be synced by whoever owns the release flow.


khaliostr:
Runtime verification of the fork change this bump carries, done locally with a before/after on the same machine (full detail in #729).

Relay, node and load client on loopback, shared isolate, `MAX_GAMES=2`, 26 games in each arm:

| object | PRE-fix | POST-fix |
| --- | --- | --- |
| `forge.game.Game` | 14 | 2 |
| `forge.game.card.Card` | 6,499 | 654 |
| `forge.item.PaperCard` | 118,579 | 118,579 |
| live heap | 445MB | 203MB |

The pre-fix arm reproduced the fleet's 14 retained games exactly. Card database identical in both, so nothing else moved.

Incidentally, the same run measures the engine without any network: p50 124ms over loopback against 227ms on prod, which puts about 103ms of the median in the wire and matches the 112ms floor measured earlier (71ms client to relay, 41ms relay to node).

anacletos (approved):

Issue #640, opened 2026-08-08
Bugfix - Creature with multiple enchantments can only see one


fedepoi:
I have no way to target the enchantment i need to get rid of this guy. He has Commander Plate i think is what it's called which give creature protection from colors other than blue. It's basically making that guy completly unkillable for me because I can't see the other enchantment to target it.

Forge

---
Reported by **devon7654** in the Issue Tracker Discord forum: https://discord.com/channels/1502665088304419008/1535520645138620476

khaliostr:
[[commander plate]]

PR #739, opened 2026-08-22
fix(ui): preserve dragged battlefield placement

### LINKED ISSUE #61: ui: bad auto placement of cards
Problem is of double nature:

1. When I play a card via shortcut (press 1 to cast), the card is missplaced in the battleground, and should instead follow top bottom (creatures, artifacts, lands)
2. When I play a card, and I place it in a specific slot, and a trigger causes another card to be placed first, the new card is taking the slot, and the card I played gets auto placed. It should be the other way around.

fedepoi:
## Summary

- Bind a manually selected battlefield cell to the permanent that the player dragged.
- Keep the cell through intervening triggers, then release it when the cast is canceled, countered, attached, or otherwise does not produce a top-level permanent.

## Why

Triggered permanents could take the selected cell before the dragged permanent entered. The dragged permanent then fell back to automatic placement.

Fixes #61

## Test plan

- yarn lint:all
- yarn build:web
- yarn test: 74 tests pass and 3 skip. useScryfallStore.test.ts has the existing __APP_VERSION__ test-environment failure.

## Demo

No recording. The collaborative browser is not cross-origin isolated, so its game runtime cannot create SharedArrayBuffer.

### DIFF
--- a/src/hooks/useHandDrag.ts
+++ b/src/hooks/useHandDrag.ts
@@ -14,6 +14,7 @@ interface UseHandDragOptions {
   battlefieldContainerRef: React.RefObject<HTMLDivElement | null>;
   handDropExclusionPx?: number;
   onCastSpell: (cardId: string) => void;
+  onBattlefieldDrop?: (card: CardDto, position: { clientX: number; clientY: number }) => void;
   dismissHover: () => void;
   onLongPress?: (card: CardDto, pos: { x: number; y: number }) => void;
 }
@@ -22,6 +23,7 @@ export function useHandDrag({
   battlefieldContainerRef,
   handDropExclusionPx = 0,
   onCastSpell,
+  onBattlefieldDrop,
   dismissHover,
   onLongPress,
 }: UseHandDragOptions) {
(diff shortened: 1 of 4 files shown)


khaliostr:
Hey @fedepoi , nice work. Since we're touching the battlefield we could fix the fact that multiple auras/equipment cannot be selected when attached to a permanent. only the first of the pile can be selected and interacted with. There are a couple of open issues in relation to this but i've noticed during playtesting too

anacletos (approved):

fedepoi:
> Hey @fedepoi , nice work. Since we're touching the battlefield we could fix the fact that multiple auras/equipment cannot be selected when attached to a permanent. only the first of the pile can be selected and interacted with. There are a couple of open issues in relation to this but i've noticed during playtesting too

@khaliostr this is intended behaviour ill open an issue to improve this UX flow

fedepoi:
Tracked the attachment-pile UX follow-up in #742. It links the existing player report in #640 and defines mouse, touch, targeting, and host-selection acceptance criteria.

PR #741, opened 2026-08-22
fix: style onaccount and minor fix

anacletos:
## Summary

* Fix minor missmatch between name and connected handle when changing username
* Improve UI on account


fedepoi (approved):
awesome

PR #755, opened 2026-08-24
fix(node): build the engine image with G1 instead of Serial

khaliostr:
## Summary

Two lines of build config: Oracle GraalVM base image, `--gc=G1`, and an absolute
heap cap in place of the percentage one.

## Why

`native-image` defaults to Serial and Community Edition offers nothing else —
`native-image --help` there lists exactly `serial` and `epsilon`. A Serial
*complete* collection traces the entire live set, and this live set is Forge's
card database: roughly 450MB that stays reachable for the life of the process
and never becomes garbage. At the ~1.3ms per MB measured in #684 that is about
half a second per complete collection **whatever it reclaims**.

Which means the pause has nothing to do with the game being played. A two-player
game on a 31-card board pays the same as a four-player game on 200 cards, and
that is why the stalls in the current fleet concentrate in *small* games.

This is #684. It was fixed by naming G1 on the JVM in #691 — the period after
that deploy is the cleanest in the 40-day window, 0 of 418 games with a p90 over
two seconds — and the move to a native image reverted it without anyone choosing
to.

## Verification

Built against the real 26MB harness fat jar with the existing `build-native.sh`
flags plus `--gc=G1`, then linked a C driver and called the production entry
point:

```
 Java version: 21.0.12.1+1-LTS, vendor version: Oracle GraalVM 21.0.12.1+1.1
 Garbage collector: G1 GC (max heap size: 25.0% of RAM)
        9 native libraries: dl, g1gc-cr, m, pthread, rt, stdc++, z
Produced artifacts:
 /w/out-shared/forgeharness.so (shared_library)

graal_create_isolate rc=0
Read cards: 33645 files in 40 ms (25 parts) using thread pool
forge_initialize -> {"ok":true,"result":""}
```

Plus 8 concurrent `graal_attach_thread` calls on that isolate, all `rc=0`, which
is the shape `create_in_shared_isolate` uses. Negative control: the same build
without `--gc` reports `Serial GC` and links no `g1gc-cr`.

## The catch: one isolate only

G1 permits a single isolate per process and aborts on the second:

```
guarantee(SVMIsolateData::_heap_base == nullptr) failed:
G1 doesn't support multiple isolates at the moment.
```

`SELF_HOSTED_NODE_SHARED_ISOLATE=1` — the image default — creates one and
attaches a thread per room, so this is fine. **`SHARED_ISOLATE=0`, the
isolate-per-game fallback, now aborts on the second room** and is only usable on
a Serial build. Documented; there is no in-process guard because the collector
is not visible at runtime.

## The heap cap has to become absolute

`-R:MaximumHeapSizePercent` is rejected outright under G1, which is the good
failure mode:

```
The option 'MaximumHeapSizePercent' can only be used together with the serial
('--gc=serial') or the epsilon garbage collector ('--gc=epsilon').
```

So `FORGE_MAX_HEAP=1100m` via `-R:MaxHeapSize`. This is a real ergonomic
regression: the cap no longer tracks `mem_limit` and has to be retuned by hand
when that moves. Leaving it unset is not an option — G1 defaults to 25% of RAM,
which on a 2GB limit is ~512MB against a ~450MB live set, i.e. the #684 heap.

## Test plan

- `--gc` values confirmed from `native-image --help` on both distributions.
- Shared library built with G1 from the production jar and flags; isolate
  created, card database loaded, `forge_initialize` returned ok; 8 attached
  threads.
- Executable form also builds under G1 and reaches `Interactive server mode
  ready` in a 2GB container, so an IPC route exists if this one ever fails. Not
  used here.
- Licence: the image ships GFTC. Internal business operations and server-side
  production use are covered, redistribution permitted where no fee is charged.
  Images pull anonymously.

Not done: no end-to-end `docker build` of the image. Oracle's base is Oracle
Linux 10 and its glibc aborts under Docker emulation on Apple Silicon with `CPU
does not support x86-64-v3`, so amd64 cannot be built or run on a Mac any more
and this needs a CI runner. Every stage-2 command was run directly on arm64;
the amd64 image ships the same `libg1gc-cr.a` the arm64 build linked, and the
Community image ships none.

Costs: image +18.6MB (+21%), build +11%. G1 is Linux AMD64/AArch64 only, so
desktop builds stay on Serial and no Mac can reproduce the fleet's collector.


khaliostr:
**Do not merge.** The image build fails on amd64. Dispatched `docker-images.yml` on this branch (run 32745750081) because no PR check builds this Dockerfile.

The configuration half is confirmed working:

```
#64 [native-builder 7/7] RUN ... ./forge-harness/build-native.sh -J-Xmx8g --gc=G1 -R:MaxHeapSize=1100m
#64 8.332  Garbage collector: G1 GC (max heap size: 1.07GB)
```

So the Oracle base image pulls from CI, `--gc=G1` is accepted, and `-R:MaxHeapSize` sets the ceiling as intended. Then native-image dies compiling the harness:

```
#64 199.8 Fatal error: org.graalvm.compiler.debug.GraalError:
  com.oracle.svm.core.util.VMError$HostedError:
  Did not find a matching CFunctionEpilogueNode in same block: [25|CFunctionPrologue, 96|If]
#64 ERROR: process "/bin/sh -c ... build-native.sh ..." did not complete successfully: exit code: 1
```

An internal compiler error in the native-call transition machinery, which is what `--shared` generates around the `@CEntryPoint` layer in `forge-harness/native/forge/harness/ffi/ForgeNative.java`.

Architecture-specific: the same `--shared --gc=G1` build against the same jar and flags succeeds on **arm64** (`Garbage collector: G1 GC`, links `g1gc-cr`, isolate created, card database loaded, `forge_initialize -> {"ok":true,"result":""}`). It is the **amd64** build, which is what production runs, that fails.

Also confirmed on this run: the earlier `ARG GRAALVM_IMAGE` scoping bug is fixed, and the other three images build fine.

Next thing worth trying is the executable form rather than the shared library. `forge.harness.Main --interactive-server` is already a multi-session server over stdin/stdout and `SubprocessBridge` already speaks it, so it is a change of artifact shape rather than of architecture — and it carries far less `@CEntryPoint` surface, which is where this error lives. That builds under G1 on arm64. Whether it survives amd64 is the open question.

Failing that: a newer GraalVM, or leave the collector alone and attack the live set instead — the pause is 1.3ms per MB and the card database is ~450MB of permanently reachable objects, so halving it halves every complete collection whatever the collector.

khaliostr:
**Green.** Run 32748449494 on the branch: all four images built and pushed, `manabrew-node` included.

```
#64 9.427  Garbage collector: G1 GC (max heap size: 1.07GB)
#64 324.6 Finished generating 'forgeharness' in 5m 22s.
```

Native amd64 runner, default `-O2`, no workaround. What it took to get there, since none of it was visible to any existing check:

| | |
| --- | --- |
| `ARG GRAALVM_IMAGE` used in `FROM` before being declared | blank base name |
| GraalVM 21 + G1 + amd64 | codegen abort on a synthesised reflective factory method |
| same at `-O1` | identical failure, same node ids — a codegen bug, not an optimisation artefact |
| GraalVM 24 | builds at `-O2`; the `-O1` workaround and its throughput cost are retired |
| runtime stage copied only the harness | 24 emits `libawt`, `libfontmanager`, `libjava`/`libjvm` shims beside it |
| copying the directory instead | would have shipped `forgeharness.so`, a second 121MB copy build-native.sh leaves for the `-l` link name |
| GraalVM 24 is Oracle Linux 10 | harness links glibc 2.38; both Debian stages were bookworm at 2.36 |

The last one is the one to keep in mind: it failed the link here only because Rust links against the library. Had it not, it would have passed the build and failed `dlopen` in production.

Still not covered by this PR:

- **No container smoke test.** A pushed image proves the link, not the load. #756 adds a PR job that builds this file and starts the container, requiring it to reach `shared graal isolate initialized`. It would have caught three of the seven above.
- **G1 at runtime is unverified.** The build banner says G1; nothing yet confirms the running fleet is on it. `manabrew_node_engine_gc_pause_millis_total{collector=...}` from #754 (now merged, releasing as v3.21.0) is what answers that, and it is the first time the fleet reports its own collector at all.
- `FORGE_MAX_HEAP` no longer tracks `mem_limit` and must be retuned by hand when that moves. Unset is unsafe: G1 defaults to 25% of RAM, ~512MB on a 2GB limit, below the ~450MB live set.
- `SELF_HOSTED_NODE_SHARED_ISOLATE=0` is now ignored with a warning rather than honoured — G1 aborts on a second isolate.

PR #756, opened 2026-08-24
ci: build and start the node engine image on a PR

khaliostr:
## Summary

A path-gated job that assembles `manabrew-rs/crates/self-hosted-node/Dockerfile`
and starts the container, gated on the Dockerfile, `build-native.sh` and the
native config.

## Why

Nothing covers that file today. `GraalVM node (Linux)` builds the harness with
`distribution: graalvm-community` straight from `build-native.sh`; the Dockerfile
is only ever assembled by `docker-images.yml`, which triggers on `v*` tags. So a
change to the engine image is first exercised by the release that ships it to
production.

That gap is not theoretical. Moving the image to G1 (#755) produced five
distinct failures, none of which any existing check could see:

| | Caught by |
|---|---|
| `ARG` used in `FROM` before declaration, expanding to a blank base | `docker build` |
| native-image codegen abort on amd64 under G1 | `docker build` |
| JDK libraries emitted beside the harness, left out of the runtime layer | **only running it** |
| a second 121MB copy of the harness a directory copy would ship | review, or image size |
| glibc 2.36 runtime against a harness linked at 2.38 | link here, `dlopen` in production |

Three of those five build green. The harness is `dlopen`ed, so a missing JDK
library or a glibc mismatch surfaces on start, not on build. That is why the job
runs the container and requires it to reach `shared graal isolate initialized`
rather than stopping at a successful build.

## Test plan

- Job is `needs: changes` and fires on `run_node_image`, matching the existing
  `run_rust` / `run_native` / `run_forge_room` pattern.
- Paths: the Dockerfile, `forge-harness/build-native.sh`,
  `forge-harness/native/`, and this workflow.
- Builds `linux/amd64` with `push: false, load: true`. This is a check, and the
  registry copy stays `docker-images.yml`'s job on a tag. Reuses the existing
  `buildcache` ref so it is not a cold build.
- Starts the container with a deliberately unreachable relay URL and a 2g limit
  (production's `mem_limit`), then polls the log for the isolate line, failing
  early if the container exits and dumping logs either way.
- Counts the card scripts in the runtime layer and fails under 30k. This used to
  read the engine's `Read cards:` line, which no longer exists: #759 turned on
  `LOAD_CARD_SCRIPTS_LAZILY` in `ManaBrewEngineAdapter`, which is what
  `forge_initialize` calls, and `CardStorageReader.loadCards` returns before it
  logs when that is set. Counting files in the image is what the check was for
  anyway, and it does not depend on a log line.
- 45 minute cap; native-image dominates and takes 3 to 6 minutes on a warm
  cache.
- YAML validated with `yq`, and re-validated after the merge.
- `find` is present in `debian:trixie-slim`, checked against the image.

This PR does not change the image. #755 does that, and would have been caught by
this job three times over.

Review comments addressed: both blocks anacleto flagged are cut down, and the
two other long ones in the job went with them.


.github/workflows/build-checks.yml
```diff
@@ -878,3 +892,81 @@ jobs:
...
+  # the PR checks build the harness with Community Edition straight from
+  # build-native.sh, and the Dockerfile itself is only ever built by
+  # docker-images.yml on a `v*` tag — after merge, on the way to production.
+  #
+  # It earns its runtime because the failures here are the kind that build
+  # green: a native-image codegen abort, JDK libraries emitted beside the
+  # harness and left out of the runtime layer, and a glibc mismatch between the
+  # GraalVM base and the Debian stages that fails `dlopen` rather than the link.
+  # `docker build` alone would catch only the first, so the container is started
+  # and has to reach a listening engine before this passes.
```
anacletos: Comment

.github/workflows/build-checks.yml
```diff
@@ -58,6 +59,18 @@ jobs:
               '.github/workflows/build-checks.yml',
             ];
 
+            // The node image assembles four stages nothing else exercises: the
+            // GraalVM base, native-image's own output, the glibc the harness is
+            // linked against, and which artefacts reach the runtime layer. Every
+            // one of those has produced an image that builds and then fails to
+            // run, so the only useful check is building it.
```
anacletos: Inline comment hahahahha

anacletos (dismissed):
Looks good just comments that I dont like :)

.github/workflows/build-checks.yml
khaliostr: cut to one line: "Nothing else builds the node image before a release tag does."

.github/workflows/build-checks.yml
khaliostr: cut to three lines. also trimmed the two other long ones in that job while I was in there.

khaliostr:
merged main. one thing that came out of it: #759 turned on `LOAD_CARD_SCRIPTS_LAZILY` in `ManaBrewEngineAdapter`, which is the class `forge_initialize` calls. `CardStorageReader.loadCards` returns before it prints `Read cards:` when that is set, so the card assertion I added would have failed on every run from now on. It now counts card scripts in the runtime layer instead (`docker run --entrypoint find`, fails under 30k). Same thing it was checking, no log line involved.

Worth knowing separately: the hosted nodes now load card scripts lazily in production too.

anacletos (approved):

PR #759, opened 2026-08-24
feat(forge): run Forge in the browser via GraalVM Web Image

khaliostr:
> **Experimental, behind a flag and a Settings opt-in.** Dark in production, on in staging. Nothing changes for a player who does not turn it on.

Forge, the Java rules engine, compiled to WebAssembly with GraalVM Web Image and playing inside the real Manabrew client. Same bytecode through a different Native Image backend, so the callback stream is byte-identical to the JVM and there is no parity surface to defend.

```
33.5 MB  forgeharness.js.wasm      8.3 MB brotli
 462 ms  boot, assets included
1-10 ms  engine think time, p50 per decision
```

## What a player gets

| | |
|---|---|
| Play vs AI, offline | Constructed, commander, brawl, oathbreaker |
| Hosting a table | Up to four humans over the relay, game runs in the host's browser |
| Autopass | Dead windows and the held pass, our implementation, not Forge's |
| Limited, draft, sealed | Served by the Rust worker, which is what answers it today |
| Rollback | Not available. The harness refuses it, so the control is disabled |

Against the hosted node: a cheap decision costs about 67 ms hosted and 46 ms in the browser, so the win there is small. The prize is the tail. A hosted ability activation is a fixed 385 ms charge; the browser samples 54 activations at a p50 near 20 ms.

## Verified on staging

Every check runs against https://staging.manabrew.app, a production build.

```
forge-wasm-toggle       the Settings choice takes effect mid-session
forge-wasm-offline      62 prompts, 62 states, board mounts
forge-wasm-commander    Ramses at 40 life, no unsupported cards
forge-wasm-limited      193 sets listed with the engine on
forge-wasm-rematch      a second game starts after walking out of the first
forge-wasm-table        two and four humans, Forge host, every seat acts
forge-wasm-autopass     dead windows pass, the held pass is honoured
forge-wasm-telemetry    a finished game reports its own timings
```

The telemetry line is the browser engine measuring itself on the deployed build: 7 decisions, turnaround p50 54 ms, engine think p50 21 ms. On a dev server, over a longer game: turnaround p50 38 ms, engine p50 8 ms, against 271 ms for the Rust engine on the same script.

Engine log for a game: 14 lines, no stack frames, no complaints. Boot 462 ms.

## What this had to fix

- **Commander decks played as constructed.** The worker hardcoded `variant: "Constructed"`, and Forge takes commander rules from the variant, so a commander deck played as a 100 card pile at 20 life with the commander absent from the game.
- **Accented card names were invisible.** Forge loads scripts lazily and a lazy lookup goes by filename, accents stripped. The bundle wrote its own spelling, so `Lim-Dûl's Vault` was framed as `lim_d_l_s_vault` while Forge looked for `lim_duls_vault`. Five presets carry such a card.
- **Everything the engine does not answer.** Forge owns a game and nothing else, and the worker answered every other command with `null`. Opening Limited took the page down. Those commands now go to the Rust worker, and the Forge worker returns an explicit error rather than a null for anything unexpected.
- **Hosting a table did nothing.** `start_multiplayer_game` returned null. Every seat now has its own buffer and its own view of the board, and the board is published to every seat on every exchange: Forge asks one seat at a time, so a guest whose board arrived with their own first prompt watched a blank screen.
- **A missing image killed the Play view.** Renderers index `deckCard.uris[resolution]` and five presets ship cards with no `uris` at all. The engine then waits forever for a response the crashed UI never sends, which reads as the engine hanging.
- **Boot shouted.** No home directory, so four preferences files failed to read and printed a stack trace each; the AI profiles were missing from the archive, so every opponent played on built-in defaults.

## Production

`docker-images.yml` now builds the engine into the release image, the same step the staging leg runs, so production can turn the flag on when it wants to. It stays off: `FORGE_WASM` has no default in `compose.production.yml`, and the engine is off until a player opts in on top of that. The entrypoint also refuses to advertise an engine the image does not carry, which is the failure this feature hit three times.

## Licensing

Reviewed. Oracle GraalVM and Web Image are GFTC, EA terms included, which permits redistributing runtime portions embedded in compiler output as long as distribution and use are not for a fee. GPLv3's System Library exception covers compiler runtime components under separate terms, so this does not conflict with Forge's GPLv3+ or ManaBrew's AGPLv3+. A paid distribution or paid service needs a fresh review.

## Already under review elsewhere

Two commits here are also open against main on their own, because they are not wasm bugs:

- `fix(forge): resolve a deck's flavor name to the card Forge knows` — #766. A hosted bug too, proven against the JVM Forge the nodes run.
- `feat(telemetry): measure the engines that run on players' machines` — #773, with the two fixes that came out of testing it against a deployed build: a game the player walks out of, and a game whose tab is closed, both used to lose their report. This branch adds the browser engine's label and its own think time on top.

## Still open

- No multiplayer game played to completion. The table tests prove every seat acts, not that a game ends cleanly.
- Chrome only in tests. It is played in Firefox, but nothing guards it.
- If the host backgrounds their tab the relay poll stops, which parks the engine. True of the Rust engine too, but hosting in a browser makes it matter more.
- The missing-`uris` fix is proven by unit test, never against the deck that first hit it.

## Try it

```bash
npx vite --port 5199                      # the flag auto-enables in browser dev
# Settings, "Forge engine in the browser", On

BASE=http://localhost:5199 node tests/e2e-ui/forge-wasm-offline.mjs
BASE=http://localhost:5199 node tests/e2e-ui/forge-wasm-commander.mjs
BASE=http://localhost:5199 node tests/e2e-ui/forge-wasm-autopass.mjs
MANABREW_SERVER_KEY=forge cargo run --release -p manabrew-server
BASE=http://localhost:5199 SEATS=4 node tests/e2e-ui/forge-wasm-table.mjs
```

Rebuilding the engine needs Oracle GraalVM 25.1+ EA and binaryen:

```bash
node scripts/harness.mjs build && forge-harness/build-wasm.sh
cp forge-harness/native/wasm/forgeharness.js* public/forge/
```

The engine binary does not rebuild with a Java edit on its own. A stale one is how a local run ends up testing an older engine than staging.

## Traps

- `Dockerfile.web` must copy `formats`, `defaults`, `effects`, `ai` and all of `lists`. A missing directory is skipped silently and the archive ships without it.
- Dropping `res/lists` looks harmless. It changes card legality and parity diverges immediately.
- `window.__gameStore` is stripped from a production build. A test that reads it reports an empty board against staging and looks like an engine hang.


khaliostr:
@anacletos ready to go in. FF on in follow up branch

forge-harness/native/web/index.html
```diff
@@ -0,0 +1,449 @@
+<!doctype html>
```
anacletos: maa what is this for?

src/lib/decks.ts
```diff
@@ -12,6 +12,20 @@ function normalizeTokenName(name: string): string {
...
+
+/**
+ * Renderers index `deckCard.uris[resolution]` directly, and the type says a
+ * deck card always has them. A user-authored deck does not have to: a card with
+ * no `uris` at all reaches the board and takes the whole Play view down with
+ * "can't access property border_crop". Give every resolved card the object.
+ */
+function withImages(card: DeckCard): DeckCard {
+  return card.uris ? card : { ...card, uris: {} as DeckCard["uris"] };
+}
```
anacletos, line 16: rmmmh what? This can never happen and we can't have fallbacks for this stuff

anacletos (changes requested):

tests/e2e-ui/forge-wasm-activation.mjs
```diff
@@ -0,0 +1,303 @@
+// Measures browser engine turnaround split by the kind of action submitted,
```
anacletos, line 1: idk surely we're supposed to clean these up?

anacletos:
I'd like to clean this PR up before we send it in. It's really full of slop

src/lib/decks.ts
khaliostr: It does happen and was crashing, hence the fix

tests/e2e-ui/forge-wasm-activation.mjs
khaliostr: Do you mean the tests or the measurements or the comment? tests and telemetry would need to stay i reckon comment can go

forge-harness/native/web/index.html
khaliostr: ?

src/lib/decks.ts
khaliostr: Might need to be fixed better that i agree

forge-harness/native/web/index.html
khaliostr: ah i think it's a stale throwaway, can be removed

anacletos (approved):

PR #763, opened 2026-08-25
perf(relay): stop copying the whole board for every state envelope

khaliostr:
## Summary

- Fold a state patch by taking the cached board out of the map and swapping its state in place, instead of borrowing it and cloning it twice.
- Only copy the folded board when a seat that cannot apply patches is actually receiving it.
- `state_patch_audience_ready` now takes the room the caller already holds, so that decision can be made before the copy rather than after it.

## Why

The node emits one state envelope per seat plus an observer snapshot, so a four-seat room sends five per decision where a two-seat room sends three. Every one of those made three full deep copies of the cached board, and all of them ran while the room guard was held, so they serialised rather than overlapping.

- `replay.rs` cloned `previous` twice per patch (once for `base`, once for `rebuilt`) purely to produce a value it then put straight back in the map.
- `connection.rs` cloned the folded board for every patch envelope, and used it only when some client was too old to apply patches. Every current client applies them, so in the normal case it was a full copy of a Commander board allocated and dropped.

That is roughly 15 whole-board copies per decision in a four-seat room against 9 in a two-seat one, and each copy costs in proportion to board size, so it grows through the game.

This matches the measurement. Splitting the response-to-prompt interval per decision across two days of captures, by `engineMs` stamped in the node:

| seats | wire p99 | engine p99 | rest p99 | rest >2s |
| --- | --- | --- | --- | --- |
| 2p (479 games, 136k decisions) | 0.461 | 0.194 | 0.270 | 0.0% |
| 4p (59 games, 49k decisions) | 2.856 | 0.473 | **2.298** | 1.7% |

Forge grows 2.4x from two seats to four. Everything around it grows 8.5x. At p99 in a four-seat game, 2.3 of the 2.9 seconds are outside the rules engine, and this is relay-side work in that window.

Format is not involved: Commander 2p and Standard 2p both sit at an `engineMs` p99 of 0.17-0.19s. It was never a Commander problem, it is a seat-count problem.

## On the audience check

`state_patch_audience_ready` used to look the room up through `state.rooms`, which is why it could not be called while the guard was held, which is why the copy was made unconditionally in the first place. Re-locking after `drop(room)` would have worked but opens a window in which another envelope folds a newer board between the decision and the copy. Passing in the room the caller already has avoids that: the check and the copy stay in one critical section, exactly as before. It has one caller, so the signature change is contained.

## Test plan

- `cargo check`, `cargo clippy` zero warnings, `cargo fmt` clean.
- `cargo test -p manabrew-server`: 16 pass. Three cover this fold directly, including `seats_do_not_share_a_board`, which is the case the take-and-restore change could have broken, and `a_patch_for_a_seat_with_no_board_leaves_nothing_to_hand_back`.
- Full `yarn lint:all` green via the pre-commit hook.

Behaviour is intended to be unchanged. `&&` short-circuits, so the audience check still runs only for patch envelopes; a downgrade still emits `record_state_patch_downgrade` and still sends the folded board.

**Not covered by any test:** the downgrade branch itself, i.e. an old client that cannot apply patches. There is no integration coverage for it in `networking-tests` either. It is the branch that now decides whether the copy happens, so it is worth a look in review.


manabrew-rs/crates/manabrew-server/src/replay.rs
```diff
@@ -73,19 +73,22 @@ impl GameReplayCache {
             // reads gameOver out of it.
             Some("stateDelta") => {
                 let slot = envelope.get("forPlayer").and_then(Value::as_str);
+                // Taken by value, not borrowed and cloned: this is the whole
+                // cached board, it is rebuilt once per seat per decision, and a
+                // four-seat Commander room does that five times a decision.
```
anacletos: comment

manabrew-rs/crates/manabrew-server/src/connection.rs
```diff
@@ -1317,16 +1324,26 @@ fn handle_client_message(
...
-                            .and_then(|replay| replay.state_after(&game_state).cloned())
-                    })
-                    .flatten();
+                // `observe` has already folded this patch into the cached board,
+                // so an older seat can be handed the state the patch would have
+                // produced rather than an envelope it drops. Only take a copy
+                // when such a seat is present: the fold is a whole board, and
+                // copying it for every envelope meant copying it once per seat
+                // per decision and throwing it away, because current clients all
+                // apply patches.
```
anacletos: I know there was a comment already, but I think it's not very useful, the code is pretty clear

manabrew-rs/crates/manabrew-server/src/connection.rs
```diff
@@ -221,24 +221,32 @@ fn room_player_id(
...
+            .map(|p| p.player_id.clone())
+    }
 }
 
 /// Whether every client that would receive this envelope ships the `stateDelta`
 /// applier. One seat that does not is enough to fall back to a full state: a
 /// dropped patch leaves that player's board frozen for the rest of the game.
+/// Takes the room rather than looking it up, so it can be asked while the room
+/// guard is held. That is what lets the caller decide whether it needs the
+/// folded board before cloning one.
```
anacletos: meta commentary comment

anacletos (changes requested):
Few comments I think we can do without.
Also merge conflicts.

Looks good though

manabrew-rs/crates/manabrew-server/src/replay.rs
khaliostr: Removed.

manabrew-rs/crates/manabrew-server/src/connection.rs
khaliostr: Removed.

manabrew-rs/crates/manabrew-server/src/connection.rs
khaliostr: Removed.

khaliostr:
addressed and merging

PR #766, opened 2026-08-25
fix(forge): resolve a deck's flavor name to the card Forge knows

khaliostr:
A deck can name an alt-art printing by its Scryfall flavor name. FCA #40 is a Lightning Bolt the deck calls "Thrum of the Vestige", and Forge has no card by that name, so the lookup missed and Forge substituted its "this card is not supported" placeholder. The game then played on around a dead card.

This is not a browser-engine bug. Confirmed against the JVM Forge the hosted nodes run, driving the harness with that printing:

```
An unsupported card was requested: "Thrum of the Vestige" from "fca".
Forge could not find this card in the Database.
```

After the fix the same run reports none.

## What changed

- `ManaBrewEngineAdapter` asks the edition what it prints at that collector number, and takes its name only when the flavor name matches the one requested. A mistyped card is never swapped for its neighbour. The printing is still requested by set and number, so the art does not change.
- The Neheb preset was the only shipped deck naming one. It now says Lightning Bolt, with the type line and oracle text that were missing while the name resolved to nothing.

Forge's own flavor-name index cannot answer here: it is built as scripts load, and the harness runs with `LOAD_CARD_SCRIPTS_LAZILY`, so it is empty for a card no game has touched. It stays as a fallback for decks that carry no printing.

492 printings in the current card set carry a flavor name, so any imported deck can hit this.

## Verification

`ManaBrewEngineAdapterTest` covers the flavor name resolving, a mistyped name being left alone, and a known name never being rewritten. Watched failing with the resolver stubbed out.

## Follow-up

The browser build needs one more change, since it frames only the card scripts a deck names: `khaliostr/web-image-spike` carries it (`fix(forge-wasm): frame the script a flavor name really asks for`). Nothing here depends on it.

### DIFF
--- a/forge-harness/src/main/java/forge/harness/host/ManaBrewEngineAdapter.java
+++ b/forge-harness/src/main/java/forge/harness/host/ManaBrewEngineAdapter.java
@@ -10,7 +10,9 @@ import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import forge.ai.LobbyPlayerAi;
+import forge.StaticData;
 import forge.card.CardDb;
+import forge.card.CardEdition;
 import forge.deck.CardPool;
 import forge.deck.Deck;
 import forge.deck.DeckSection;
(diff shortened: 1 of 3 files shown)


fedepoi (approved):

khaliostr:
Landed in #759. The adapter fix and the Neheb preset are on main unchanged.

PR #773, opened 2026-08-25
feat(telemetry): measure the engines that run on players' machines

khaliostr:
The hosted node publishes its decision timings to Prometheus, so we know exactly how the fleet performs. Nothing measures the engines that run on a player's machine. Offline play is the extreme case: no server is in the loop at all, so nothing about it has ever been observable.

That gap matters now. The whole argument for a client-side engine is latency, and every number we have for it comes from a laptop under a test harness.

## What it measures

What the player feels: the gap from sending an answer to the next prompt landing. Engine-agnostic on purpose, because the number is only worth having if the browser engine and the hosted one are measured the same way.

One report per finished game:

```
reportId, engine, clientVersion, platform, format, seats, multiplayer,
durationS, endReason,
turnaround   { n, p50, p90, max }     always
engineThink  { n, p50, p90, max }     when the engine reports its own
byType       [ { type, n, p50, max } ]  capped at 12
```

Aggregates only. No deck, no cards, no opponent, no name. The report id is generated by the client, so a retry cannot double-count a game.

## Where it goes

**Over the relay**, for a game the relay already knows about. `ClientMessage::ReportEngineStats` is recorded as `AnalyticsEvent::EngineStats` beside that room's other analytics, and forwarded to nobody: it is analytics, not state. This also answers a question the relay cannot answer today, which is **which engine actually ran** — a browser hosting a "Manabrew" room on the Forge engine is invisible in the current data.

**To the hub**, for offline play. `POST /api/stats/engine`, unauthenticated because offline play has no session, rate limited per IP with the same limiter the deck play endpoint uses, stored in `engine_play_stats`. Reports queue in localStorage and flush on start and on `online`, the same shape `deckPlayEvidence` already uses, so a game that ends with no connection still counts when there is one. Dropped after a fortnight rather than kept forever.

## Shape and validation

`manabrew_protocol::telemetry::EnginePlayStats` is the single definition: the relay message, the hub endpoint and the client all describe the same thing. It validates itself through `is_plausible()`, and both doors call it. An unparseable id, empty or oversized text, impossible seats, or a game with no decisions is refused.

## Tests

- `engineTelemetry.test.ts`: percentiles, grouping, and that a prompt with no answer behind it is not counted as a turnaround (the opening prompt would otherwise report the whole engine boot).
- `telemetry.rs`: accepts a real report, refuses six shapes a mistake or a hostile client produces.
- `storage.rs`: recorded once per report id, a retry is a no-op, an engine that reports no think time keeps a null.

## Not in this PR

- No Grafana panel yet. This fills the table; charting it is the next step, and for the relay side it is a JSONL/events.db query like the others rather than a new Prometheus series.
- No aggregation into a Prometheus series for browser games. Thousands of clients reporting individually needs relay-side aggregation, and that is worth doing only once there is a reason to watch it live.


khaliostr:
Landed in #759, which also added the browser engine's label on top. This branch no longer carries it, so merging it would drop that.

PR #774, opened 2026-08-25
feat(game): show persistent card choices

### LINKED ISSUE #578: UI - Add extra indication for modes/choices of specific permanents
e.g.
Image

fedepoi:
## Summary

- expose viewer-safe persistent card choices from the Rust and Forge engines
- show the selected value and mana glyph directly inside the battlefield choice chip
- retain existing status labels without adding a redundant bottom-left mana pin
- route exact color decisions through the symbol-first ChooseColor prompt
- add a dedicated free-text card-name prompt instead of serializing the full card database
- cover colors, types, named and selected cards, numbers, modes, and players
- keep secretly chosen numbers private until revealed
- add battlefield debug-card controls for every persistent choice kind, including an all-at-once mode

## Why

Permanent choices were dropped at the game DTO boundary, so cards such as Temple of the Dragon Queen and Roaming Throne gave no lasting battlefield indication. Generic prompt routing also made color choices text-only and made unrestricted card naming send tens of thousands of options.

Fixes #578.

## Test plan

- `yarn lint:all`
- `yarn build:harness`
- `yarn gen:protocol-docs`
- Forge inline-deck smoke including Pithing Needle, Temple of the Dragon Queen, Black Vise, Roaming Throne, and Barrensteppe Siege
- focused Forge smoke for Ominous Lockbox and Alhammarret, High Arbiter
- battlefield debug card: toggle each persistent choice or click All to inspect every variant together
- visual QA intentionally left to the maintainer

## Demo

Not included; visual QA will be completed by the maintainer.


manabrew-rs/crates/manabrew-protocol/src/prompts/choose_card_name.rs
```diff
@@ -0,0 +1,23 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, TS)]
+#[serde(
+    tag = "type",
+    rename_all = "camelCase",
+    rename_all_fields = "camelCase"
+)]
+#[ts(export, export_to = "prompts/chooseCardName.ts")]
+pub enum ChooseCardNameOutput {
+    CardNameDecision { name: String },
+}
```
anacletos: I'd just double check this is called and whether is necessary / even viable to provide a free text string field. When / Which kind of cards would make use of this?

anacletos (changes requested):

manabrew-rs/crates/manabrew-protocol/src/prompts/choose_card_name.rs
fedepoi: Good question! Free text is needed for unrestricted naming effects like Pithing Needle, since the suggestion list is only there to help with discovery. The backend still validates the submitted name case-insensitively against the full set of legal names and canonicalizes it before accepting it.

Constrained effects like Alhammarret use the same flow, but with a smaller set of valid names. I also updated the Forge bot path to always receive up to 50 valid suggestions instead of an empty list.

anacletos (approved):

PR #780, opened 2026-08-28
ci: fail when the protocol docs contradict the source

khaliostr:
## Summary

Adds `scripts/check-docs-drift.mjs` (`yarn check:docs-drift`), run by `protocol-docs.yml`. It fails when a protocol docs page states something the source contradicts.

## Why

#778 makes CI build the docs. Building is not the same as being true. Both defects #777 fixed were in pages that built green and read fine:

- `index.mdx` said "This is **v1** of the protocol" while the crate was 5.2.0.
- `transport.mdx` tabled nine relay envelope kinds when the wire had eleven — `stateDelta` was missing while STATE_DELTA was on by default, so anyone implementing from the table dropped most of what we send.

Both shipped for months. Nothing in the repo could have caught either, and the Astro build never will.

## What it checks

Three rules, each comparing a machine-readable claim in the docs against a machine-readable definition in the source:

| Claim in the docs | Checked against |
| --- | --- |
| the transport page's envelope `kind` table | the `StateEnvelope` union in `src/types/server.ts` |
| `<crate> X.Y.Z` written in prose | that crate's `Cargo.toml` |
| a stated `PROTOCOL_VERSION` integer | the **relay** crate's major |

The third is worth its own rule: `PROTOCOL_VERSION` comes from `manabrew-relay-protocol`, not `manabrew-protocol`. Both sit at major 5 today, so the conflation is invisible until the two crates diverge — it survived a bump inside phase-rs's adapter for exactly that reason.

The union parser resolves `kind: typeof ROOM_RELAY_KIND` members, not just inline string literals, so `roomRelay` counts.

## Test plan

Green on main as-is: `Protocol docs match the source (11 envelope kinds)`.

Reintroduced each original defect and confirmed it is named, then restored:

```
$ # stateDelta removed from the table
    the envelope table disagrees with StateEnvelope in src/types/server.ts
    undocumented: stateDelta

$ # "Pinned upstream: manabrew-protocol 3.0.0"
    claims manabrew-protocol 3.0.0, but that crate is 5.2.0

$ # "The `PROTOCOL_VERSION` is 3 for this release."
    says PROTOCOL_VERSION is 3; the relay crate major is 5
```

`yarn lint:all` clean. Runs in 0.24s, no build step, no network.

## Scope

Deliberately narrow. There is no general rule for "is this documentation true", and a check that fires on prose nobody can satisfy is a check that gets disabled. Anything not of the three forms above is left alone — the rewritten versioning note in #777 states no hardcoded version at all, and this stays silent on it.

Wired into `yarn` and added to `scripts/AGENTS.md` per the repo conventions.


anacletos (approved):
Mmmh approved but wasn't there already check protocol types CI step? 
Should we just check both in there?

PR #800, opened 2026-08-31
feat(relay): count solo games in the lobby and the live gauges

khaliostr:
## Summary

- New `ClientMessage::SetLocalGame`, carrying a `LocalGameKind` enum (today only `Singleplayer`). The relay holds it on the session.
- Players in a game against the AI now show under **Playing** in the lobby player list instead of Available.
- New gauge `manabrew_relay_local_games{kind}`. "Games in progress" and "Concurrent games record" sum it with the room gauges, plus a new panel breaking it out by kind.
- `AuthResult` gained a `features` list so a newer client stays quiet against an older relay.

## Why

Play vs AI runs entirely on the player's machine: no room, no state, nothing on the wire. The relay could never see it, which did not matter while those games ran on the hosted node fleet. #797 moved the engine into the browser and now it matters a lot. Hosted games per hour on prod, 31 Aug, deploy at 12:47:

| hour | 08 | 09 | 10 | 11 | 12 | 13 | 14 | 15 |
| ---- | -- | -- | -- | -- | -- | -- | -- | -- |
| games | 11 | 7 | 11 | 10 | 3 | 1 | 1 | 1 |

`manabrew_relay_rooms{status="in_game"}` fell with them, so "Games in progress" and "Concurrent games record" stopped measuring play and started measuring the multiplayer remnant. Anything sized off them, the node fleet included, is now sized off the wrong number. #799 fixed the after-the-fact tables; nothing was reporting a game while it was running.

### Judgment calls worth reviewing

**Offline play is no longer server-free.** The client opens a relay connection during a solo game purely to announce presence, which is the only way to reach players who never visit the lobby, and it is exactly the connection #797 removed (`hostedAiPlay.ts` used to open it). That was a deliberate choice over the conservative alternative, which would have reported only Lobby visitors and left the metric decaying toward the same blind spot.

**It is a claim, not an observation.** Every other series on these dashboards is something the relay saw. This one is the client's word. It stays out of `manabrew_relay_rooms` and the panel sums the two, so the distinction survives; merging them in the exporter would have made fleet-sizing numbers unfalsifiable. It also counts only connected clients, so read it as a floor.

**Two hazards handled, worth checking I got them right.** `useServerStore.connect` claims the tab session and disconnects whoever holds it, so a background announcement would have kicked the player's own open lobby tab; `probeTabSession` gates it. And an older relay answers an unknown message with a parse error that the client renders as a toast, so `AuthResult.features` gates the send.

Per the follow-up in the issue thread: presence never delays, blocks or fails a game. Nothing in `localGamePresence.ts` is awaited by a caller and nothing throws; a game with no network behaves exactly as it does today.

## Test plan

- `yarn lint:all` clean, including workspace `cargo clippy -D warnings`. Needs `RUSTUP_TOOLCHAIN=1.94.1`.
- `cargo test -p manabrew-relay-protocol -p manabrew-server` and `yarn test` pass (`useScryfallStore.test.ts` fails in any fresh worktree on the `__APP_VERSION__` vite define, pre-existing).
- No tests added, per the repo rule. Worth checking by hand:
  - start Play vs AI with the lobby open in another window: the player moves to Playing, and the other tab is **not** disconnected
  - start Play vs AI with no prior connection: a connection opens and the player appears
  - pull the network, start a game: it plays normally and nothing is announced
  - finish or abandon a game: the player returns to Available
  - reconnect mid-game: the player is still Playing afterwards
  - point a current client at a pre-`features` relay: no error toast, no presence
  - `manabrew_relay_local_games{kind="singleplayer"}` tracks the count and returns to 0

## Demo

No screenshot yet: the change is one line in the lobby player list, which moves a player from the Available section to Playing with the subtitle "Playing solo". Happy to attach one if useful.


manabrew-rs/crates/manabrew-relay-protocol/bindings/deck/index.ts
```diff
@@ -46,7 +46,12 @@ export type Deck = {
   editor?: DeckEditorMetadata;
   coverCardName?: string;
   coverCardFace?: number;
-  playmat?: string;
+  /**
+   * The hub fills this from `playmat_asset_id` and ignores whatever a write
+   * supplies.
+   */
+  playmatUrl?: string;
+  playmatAssetId?: string;
```
anacletos: Stupid question but wasn't this bindings/ file supposed to be .gitignored?  😅

anacletos (dismissed):

manabrew-rs/crates/manabrew-relay-protocol/bindings/deck/index.ts
khaliostr: Not stupid at all, you are right and it was masking real drift.

`manabrew-protocol/bindings/` and `manabrew-hub/bindings/` are both in `.gitignore`; `manabrew-relay-protocol/bindings/` was simply never added. The evidence that it is an oversight rather than a decision:

- nothing imports those files
- in their entire history they have been touched by exactly one commit, dd78380af, which was the analytics dashboards PR (#713) and had nothing to do with protocol bindings
- the 11 lines changing in this PR contain none of my protocol change. `playmatUrl`/`playmatAssetId` are your earlier work; the tracked copies had just gone stale and my `yarn gen:types` caught them up

So the checked-in copy was never the source of truth, it was a lagging snapshot that only updated when someone happened to run the generator.

Fixed in 0283447a: added the directory to `.gitignore` and `git rm --cached` the four files. They are still generated on disk, `yarn gen:types` now produces no tracked churn, and the diff here is back to only the files this PR actually touches. The `protocol-docs` drift gate is scoped to `website/src/generated/protocol-types.json`, so it is unaffected.

anacletos (approved):

PR #807, opened 2026-08-31
feat(tauri): a desktop seat that is actually direct, and a LAN that needs no internet

khaliostr:
> **Stacked on #820**, which carries the transport itself. Based on `khaliostr/iroh-core`, not `main` — review the [compare against that base](https://github.com/witchesofthehill/manabrew/compare/khaliostr/iroh-core...khaliostr/iroh-transport). #814 sits on top of this one.

## Summary

- A desktop binds its own iroh endpoint **in the Tauri shell**, not the webview, so four desktops on one switch reach each other directly instead of being relayed on their own LAN.
- A desktop can host the whole session offline: the embedded relay binds to the network, hosts an iroh relay in the same process, and advertises the room over mDNS.
- The lobby finds a relay on this network transparently. No mode, no toggle, no IP entry: it is connection fallback, and a LAN is a favourable case of the same transport.
- `directSeat.ts` holds the browser and desktop backends behind one seam; nothing above it knows which it got.

## Why

The target case is four desktop clients in one room, sometimes with the internet on and sometimes without. Both should work without anyone being asked to understand transports.

With the internet on, nothing special happens, which is the point: the control plane stays on manabrew.app, `RoomTransport` already carries each peer's local interface addresses, and iroh selects the path across the switch. With it off, one desktop is the relay.

A browser cannot join that. Under `wasm_browser` iroh has no IP transports, so there is no socket for the LAN path, and reaching a relay on that network is blocked separately because an https page may not open `ws://` to a private address. Installing the desktop app is the way out; `docs/TRANSPORT.md` says why the alternative is infrastructure we would have to operate.

Worth stating plainly, and the doc now does: **an unprotected room on a shared network has no gate.** Before, a joiner needed a code off the host's screen. For four people at a kitchen table that is the right trade and the same trust model as every other device on the network; `CreateRoom.password` is there for anyone who wants a closed room.

## The bug this stack created, and the guard against it

`Config::for_hosted_room` turns the direct plane on, but `forge-room` pulled `self-hosted-node` with `default-features = false, features = ["graal-forge"]`. The `iroh` feature landed in #820 after this half was written, so a desktop host got the stub: `DirectPlane::start` returned `None`, no endpoint reached the roster, every seat found `roster.host()` empty, and the whole LAN case silently fell back to the relay.

`local_relay.rs` now asserts `self_hosted_node::DIRECT_PLANE` at compile time under `forge-room`. Dropping the feature again fails the build naming the consequence, rather than shipping a relay.

## Test plan

- `cargo test -p manabrew --features direct-seat` — the desktop seat runs against a real host over a real endpoint.
- `cargo check -p manabrew --features forge-room` with the resource override `scripts/clippy.mjs` uses, both with and without `self-hosted-node/iroh`, to confirm the guard fires on exactly the mistake that shipped.
- `networking-tests::direct_plane_carries_a_hosted_game` — PASS.
- `yarn lint:all` clean.
- Manual: desktop build against staging, joining a hosted room and reaching the host directly.

## Deferred, deliberately

`connect_to_host` on the desktop seat has no timeout while the bot's has two attempts of two seconds; `lan_address()` probes `192.168.1.1` and picks the tunnel under a VPN; `discover_lan_rooms` blocks a runtime thread for up to eight seconds; `direct_seat_start` binds once with the username baked in.


khaliostr:
Superseded by #831, which is the same branch targeting `main` directly. GitHub would not let this one out of its stack (#821), and it had no review history, so a fresh PR was cheaper than experimenting with the stack API.

Issue #811, opened 2026-08-31
Cache card art on desktop, and share it over the LAN


khaliostr:
## Why

A group of four wants to play together on a LAN with **no internet access at all**. One machine goes online first to fetch what the group needs, then everything is unplugged and they play locally.

Everything else in that flow now works or is being built: the desktop can host the relay on the LAN and advertise it over mDNS (#807), identity falls back to unsigned self-minted tokens with no hub configured, and the card database and Forge runtime are bundled in the app. Card **art** is the one thing that still needs the internet, every single time.

## What is missing

**There is no image cache at all.** `fetchImageBlob` in `src/api/scryfall.ts` passes `cache: "no-store"`, there is no `CacheStorage` use anywhere, and nothing is written to disk. Every board draws its art from `cards.scryfall.io` on every load.

There is also no `/scryfall-img/` route on desktop. That proxy is a Caddy route in web deployments (and the vite proxy in dev); the packaged desktop app serves itself from `http://localhost:9527` via `asset_server.rs`, so it falls back to the CDN directly.

### 1. Cache art on disk

Desktop keeps fetched images under the app data directory, keyed by the Scryfall path so the key survives a re-fetch. Then a deliberate pre-seed: **"download art for these decks"**, which walks the selected decklists and fills the cache while the machine is still online. That alone is what most people asking for this actually want, offline or not, and it removes a per-load network dependency from the most common screen in the app.

Worth deciding: a size cap and eviction, and whether the cache is per-deck or global.

### 2. Serve the cache to the LAN

Only one machine should need to have gone online. The host serves its image cache over plain HTTP on the local network and the others read from it, falling back to the CDN when they can reach it.

Two things that make this easier than it sounds, and one that does not:

- The desktop app's own origin is `http://localhost:9527`, so fetching `http://192.168.x.x/...` is same-scheme and raises no mixed-content problem. **This works on desktop and cannot work on web**, where the page is https.
- `scryfallImageProxyUrl` is already the single place that rewrites a CDN url, so pointing it at a LAN host is a small change with one call site.
- `asset_server.rs` deliberately refuses any request not addressed to loopback, as anti-DNS-rebinding. Do **not** relax that. This wants a separate listener, opt-in, running only while the machine is hosting a LAN room.

## Notes

- The host already knows its LAN address and whether it is sharing: `start_local_relay(shareOnLan)` returns `lanHost`.
[...]

PR #814, opened 2026-09-01
feat(tauri): cache card art on disk, and serve it to the LAN

khaliostr:
> **Top of a three-PR stack: #820 → #807 → this.** Based on `khaliostr/iroh-transport`, not `main` — `share_on_lan` / `lan_host` only exist there. Review the [compare against that base](https://github.com/witchesofthehill/manabrew/compare/khaliostr/iroh-transport...khaliostr/art-cache). This branch is what the `deploy-preview` label deploys to staging, so it carries the whole stack; merge it last, or merge the three together.

## Summary

- `image_cache.rs`: card art on disk under the app data dir, keyed by the path below `cards.scryfall.io`, backing the `/scryfall-img/` route the desktop app never had.
- `art_lan_server.rs`: an opt-in second listener serving that cache to the LAN while a room is shared.
- Settings → Cache: download art for my decks, or for every card, with a size estimate, live progress and a Stop.
- Pinned vs opportunistic: a deliberate download is never evicted; art picked up by browsing is dropped oldest-first past a cap.

## Why

Everything else a LAN session needs already works offline. Card art was the one thing that hit the internet every single time, on the most common screen in the app.

The two halves of #811 turned out to be the same mechanism. The issue lists "cache art on disk" and "there is no `/scryfall-img/` route on desktop" separately, but making the cache *be* that route means one thing serves loopback and LAN, and `scryfallImageProxyUrl` keeps its existing shape. Nearest source wins:

```
1. /scryfall-img/<key>        own cache, self-fills from the CDN on a miss
2. http://<lanHost>:<port>/   a machine that already downloaded it
3. cards.scryfall.io          native fetch, as today
```

### The thing most worth reviewing

**Which variants to download follows the battlefield style, and getting this wrong is silent.** The Pixi board does not draw `normal`. `getCardTexture` asks for `border_crop` on the realistic style and `art_crop` on art/frame; `normal` only covers DOM thumbnails and prompts. An earlier version of this branch downloaded `normal` alone and would have produced a full cache and a blank battlefield offline — exactly the failure the issue exists to prevent. `variantsForStyles` now derives the set from `battlefieldCardStyle`, with a checkbox to cover all three.

Sizes, measured over real cards rather than estimated (`content-length` on a random sample):

| set | per card | 38.6k cards |
| --- | -------- | ----------- |
| realistic (`normal` + `border_crop`) | 213 KB | 7.85 GB |
| art / frame (`normal` + `art_crop`) | 190 KB | 6.99 GB |
| every style | 290 KB | 10.70 GB |

Forge lands in the same place for the same reason: `ImageUtil.java:233` requests `version=normal`, plus `art_crop` only where it needs a crop. One size per surface, not every size.

### Other decisions

**"Every card" means `oracle_cards`** — 38,628 cards, one printing each, which is the set a name-keyed engine draws from. Not `unique_artwork` (54,568) or `all_cards`. Scryfall now exposes `jsonl_download_uri`, gzipped JSON Lines, so it streams line by line instead of holding a 200 MB parse. Already-cached keys are skipped, so a cancelled run resumes.

**The LAN listener is deliberately not `asset_server` with its loopback check relaxed.** That check is anti-DNS-rebinding on the origin that serves the app's assets and answers its commands; widening it would be the wrong trade. This is a separate listener with one route, no upstream fetch, and a lifetime bounded by the room. `ImageCache::path_for` is the only thing between a network request and the rest of the disk.

**Serving art is not the same trust decision as serving the relay.** The relay is password-gated; an image cache is world-readable to the subnet. It starts only under `share_on_lan`, which is already deliberate, and the port rides in `LocalRelayInfo.artPort` and an `art` mDNS property.

**No new build cost:** reqwest 0.12/rustls and flate2 were already compiled in the tree.

## Test plan

- `yarn lint:all` clean. `src-tauri` clippy `-D warnings` clean in **both** feature sets — `art_lan_server` is `#[cfg(feature = "forge-room")]` because it is dead code without it.
- 6 tests in `image_cache.rs` cover the traversal guard, pin-survives-eviction, pin-survives-reopen, and the key mappings. Against the repo's no-unit-tests rule; I would keep the traversal one at least, since it guards a network-exposed boundary, but happy to drop them.
- Three vitest suites fail on `@/wasm-net/net`. Verified by stashing this branch's changes that they fail identically on pristine #807: that branch's generated artifact is missing in a fresh worktree, like `src/wasm`.
- **Not exercisable in `tauri dev`**: `start_asset_server` returns `None` there and vite's own `/scryfall-img/` proxy takes the path, so the cache only exists in a packaged build. Needs a real `yarn tauri build` to test.
- Worth checking by hand: download for a deck, pull the network, start a game and confirm art draws on each battlefield style; then a second machine joining a shared LAN room and drawing art it never downloaded.

## Demo

No screenshot yet. The UI is one panel in Settings → Cache: the current style and the variants it needs, an "every style" checkbox, on-disk totals split by deliberate vs incidental, and four buttons (my decks, every card, trim unused, delete all) with size estimates. Happy to attach one.


khaliostr:
Folding this into #807, which now carries the whole stack as a single PR. Every commit here is in that branch already; nothing is lost. Closing so #807 can target `main` directly.

khaliostr:
Superseded by #834, which is this work rebuilt on `main` with no iroh in it. The LAN and art halves never needed the direct data plane: relaying through a box on the same switch already costs nothing, and iroh's win is the WAN case. Decoupling lets this ship without waiting on #831.

PR #820, opened 2026-09-01
feat(net): iroh transport primitives, room rendezvous, and the hosted data plane

khaliostr:
> **Base of a three-PR stack.** #807 (desktop and LAN) sits on this branch, and #814 (card art cache) on that. They are meant to land together; the split is so each can be read.

## Summary

- `manabrew-net`: an iroh endpoint, session framing, a per-seat QUIC channel and a room-scoped gossip topic. Depends only on `manabrew-relay-protocol`, so the wire protocol stays transport-agnostic.
- The relay becomes the rendezvous. `AnnounceTransport` / `RoomTransport` carry a per-room roster, a topic secret, and the deployment's own iroh relay with a token for it. Endpoint ids are attested by the control plane, never trusted from gossip.
- Hosted rooms carry per-seat engine envelopes over the direct plane once a seat proves itself, and fall back to the relay through an explicit barrier rather than a switch.
- A browser seat takes it too, through `manabrew-net-wasm`, lazily imported.
- The deployment hosts its own iroh relay, gated and rate limited. No public infrastructure.
- All of it sits behind a default-off `iroh` cargo feature on `self-hosted-node` and `manabot`.

## Why

Game traffic that could go peer to peer makes two WAN hops through a relay we pay for. This moves it to iroh where the network allows, while the relay keeps everything it is actually trusted for: room creation, discovery, auth, membership, readiness, engine selection, lifecycle and reconnect. Transport is chosen before `GameStarted` and never changes mid-game.

Design note and rationale: `docs/TRANSPORT.md`.

**The relay is gated, and the key is generated, not configured.** `relay.manabrew.app` is a public hostname and an iroh relay forwards for whoever dials it, so `RoomMembersOnly` checks a room-scoped HMAC token the control plane mints and delivers in `RoomTransport`. The signing key is minted per process with `aws_lc_rs::rand::fill` and never leaves it. It deliberately is **not** `server_key`: that value is published in `knownRelays.ts`, ships in every browser bundle, and is on screen in the relay picker, so signing with it would let anyone mint a token and the gate would be decoration. Nothing outside this process verifies a token, so nothing outside it needs the key.

**Nothing binds against a relay url.** A host and a seat both bind relay-less and adopt the relay from `RoomTransport`, the only place the url and the token arrive together. Adoption is *offered* on every broadcast, because a token lives six hours and a room outlives games. Only the first one changes the endpoint's address and triggers a fresh announcement.

**Offered, not performed.** `Endpoint::insert_relay` is not a config swap: it sends `RelayMapChange`, which `UpdateReason::is_major` makes a full net report. The relay re-broadcasts on every join and leave, so a lobby people drift through would re-probe the network on every membership change. Two things stop that: the relay mints a token **once per room** and reuses it until it is within an hour of expiring, so the string is identical across broadcasts, and every peer compares what it holds before inserting. Either alone is useless. Together it is one insert per room plus one before each expiry, and an adoption line in a log then means something happened.

**A superseded seat connection is closed, not forgotten.** Replacing the `GameSender` drops it, but the reader task holds the other half of the guard, so the old QUIC connection would stay open and keep feeding the engine responses under that seat's name.

**Running the relay is not the same as moving traffic onto it.** The fleet keeps `SELF_HOSTED_NODE_IROH` off. What ships here is the mechanism and the fallback, not a rollout.

## Known limits, stated rather than buried

- `Limits::client_rx` caps a **connection**, not a client: `accept_conn_limit` is unimplemented in iroh-relay 1.1, so one token holder can open several and pay the toll on each. What bounds that is the token being room-scoped and expiring in six hours.
- A token is room-scoped but not seat-scoped, so any member of a room can hand theirs to somebody outside it for the rest of its life.
- The gossip replay guard has no time component.
- `iroh` is an unconditional dev-dependency of `self-hosted-node`, because Cargo will not let a dev-dependency be optional. It stays out of the shipped binary but keeps the crate's MSRV at 1.91.

## Test plan

- `cargo test -p manabrew-net -p manabrew-server -p self-hosted-node -p manabot --features self-hosted-node/iroh,manabot/iroh` — 65 pass.
- `networking-tests::direct_plane_carries_a_hosted_game` — a real relay hosting a **gated** iroh relay, a real node configured against it, and a real bot seat. Asserts the bot reaches the host over iroh, the game freezes onto the direct plane, and the capture records which seats the relay could not see. PASS in 7.0s; the node log shows both the node and the bot adopting the relay.
- Two tests reproduce bugs found in review and fail without their fix: a token forged with the published relay password is refused by a live relay, and a superseded connection's `"answer": "injected"` no longer reaches the engine as that seat.
- A gated-relay test in `manabrew-net` pins that a host binding with a url and no token is reachable by nobody, while one adopting with the token is.
- `tests/e2e-ui/direct-plane.mjs` drives a real browser seat onto the plane and reads the wire off a hooked `WebSocket`.
- Live on staging: a hosted game reported `transport="iroh-direct" lan=true rtt_ms=5 setup_ms=52`.
- A full hosted game logs **two** adoption lines, one for the node and one for its bot, both `renewed=false`, across a room that saw several membership changes.
- `yarn lint:all` clean. Feature off vs on: `cargo tree` confirms twelve iroh crates and their QUIC transitives leave the graph; both configurations build, lint and test.


khaliostr:
Folding this into #807, which now carries the whole stack as a single PR. Every commit here is in that branch already; nothing is lost. Closing so #807 can target `main` directly.

PR #831, opened 2026-09-01
feat(net): the direct data plane seam and the relay rendezvous that names it

khaliostr:
Part of #838.

Rebuilt from `main`. This is the transport seam and nothing else. Every seat, every host, and a
relay of our own are follow-ups.

The full version of this branch is kept at `khaliostr/iroh-transport-full` if any of it is wanted
back. #820 and #807 were the same work split for review.

## What this is

- **`manabrew-net`**: an iroh endpoint, session framing and a per-seat QUIC channel. Depends only
  on `manabrew-relay-protocol`, so the wire protocol stays transport-agnostic.
- **The relay becomes the rendezvous.** `AnnounceTransport` and `RoomTransport` carry a per-room
  roster. Endpoint ids are attested by the control plane, never trusted from a peer.

Nothing ships on it yet. No binary depends on the crate, so no image and no CI toolchain moves,
and there are no web, ops or compose changes at all.

## Size

1,305 lines across 19 files, plus 1,506 lines of `Cargo.lock` for the 115 crates in the iroh
tree. The lockfile is more than half of what GitHub reports.

## Opting in

`MANABREW_DIRECT_TRANSPORT` is off. Off, the relay does not advertise `room_transport`, ignores
`AnnounceTransport` and sends no roster, so every room behaves exactly as it does today. It fails
closed.

`MANABREW_IROH_RELAY_URL` is optional. Unset, peers take iroh's own relay defaults. Relays
coordinate hole punching and carry traffic for a pair that cannot be punched through; they
forward QUIC between endpoint keys and never see plaintext. Running one ourselves is a separate
question, and a deployment answers it by setting that variable rather than by shipping new
clients. No client hardcodes a relay, and no address lookup service is configured either way: the
endpoint is built from `presets::Minimal` and the roster is the only source of addressing.

**The roster is the whole admission decision.** A dialer's endpoint id is proven by the QUIC
handshake; the roster is the relay's word on who owns it. A username claimed in `Hello` is worth
nothing on its own. Every username in a roster is filled in from the relay's own session records,
never from a client-supplied field.

**A refused announcement is silent.** Not a room member, or an endpoint id another member already
claimed, and the announcer simply stays on the relay;
`manabrew_relay_transport_announcements_total{kind="rejected"}` is where it shows up. Telling
someone they are not in the room they are sitting in, because a squatter claimed their endpoint
id, is worse than saying nothing.

**A superseded seat connection is closed, not forgotten.** Dropping the `GameSender` is not
enough: the reader holds the other half of the guard, so the old QUIC connection would keep
feeding the engine responses under that seat's name. `GameSender::close` is for this.

**`broadcast_room_transport` drops its room guard before broadcasting.** `broadcast_to_room` takes
its own `rooms.get` and dashmap locks per shard, so holding the guard across the call deadlocks
the thread against itself on every room whose id lands in the same shard.

## What came out of the earlier version

- **Gossip.** The node joined the room topic, announced presence, and the event loop body was
  `debug!(?event, "room gossip")`. The relay's roster was already authoritative. Dropping it took
  510 lines, `iroh-gossip`, `blake3` and the topic secret off the wire.
- **A relay of our own**, its per-room HMAC tokens and the Caddy and compose changes that went
  with it.
- **Every seat and host**: browser, desktop, LAN discovery, the node's direct plane, the bot.
  Each is its own PR.

One thing for whoever writes the browser one. iroh's `build.rs` sets `wasm_browser` for
`wasm32-unknown-unknown`, and under that cfg the IP transports are compiled out. A browser
endpoint has a real address and can host a room, but it cannot hole punch, so browser to browser
over iroh is always relayed. That moves traffic off `manabrew-server` but it is not a latency win
and it is not LAN. A direct data plane between browsers means WebRTC, behind this same
rendezvous.

## Test plan

- `cargo test -p manabrew-net -p manabrew-server`: six tests. The admission check over a real
  loopback connection, including a stranger being refused; a malformed endpoint costing only its
  own member; a frame round-trip; and the room's announcement rules.
- `cargo build --workspace` and `cargo clippy -p manabrew-net -p manabrew-server --all-targets`
  are clean.


khaliostr:
marking as draft while we review architecture and tradeoff together

khaliostr:

PR #832, opened 2026-09-01
refactor: lift the art cache and lan discovery out of the desktop shell

khaliostr:
> **Stacked on #831**, which carries the iroh transport and the desktop art cache. Review the [compare against that base](https://github.com/witchesofthehill/manabrew/compare/khaliostr/iroh-transport...khaliostr/art-server). Draft: the extraction is done, the server side is not.

## Summary

Groundwork for a headless host: one binary on a Linux box that is the relay and holds the card art, while whoever wants to play hosts the table from their own desktop.

- `manabrew-art-cache`: the cache, the `/scryfall-img/` route and the LAN listener, with the Tauri removed.
- `manabrew-lan-discovery`: mDNS advertise and browse, likewise.

## Why

Both lived in `src-tauri`, so nothing in the workspace could reach them. A headless host cannot serve card art from a Tauri crate, and holding the art is the entire reason to want one: it is what every client otherwise downloads separately, 7 to 11 GB each.

The split follows how fast the two halves move. The engine changes daily with parity work and rides the desktop auto-updater; the relay and the art cache are infrastructure that should sit on a box nobody touches. That is also the topology we already run in production, rather than a new shape nobody exercises.

## Two things that changed while moving

**A discovery record now carries a `role`.** A relay is infrastructure worth staying connected to; a room is one table on a desktop. Same fields, different destinations, so the record has to say which. Absent reads as `room`, because every shipped build advertises without it and all of those are rooms.

**`hostname()` reads `/etc/hostname` before the environment.** `HOSTNAME` is a shell variable that neither a systemd unit nor a macOS GUI app has, so every such machine advertised as `manabrew` and collided with the next. Cosmetic when it labelled your own laptop's room; not cosmetic when it is the name a player picks their server by.

Browsing also comes out from behind `forge-room`: a client that only joins has to find the host, and previously it had to be able to host in order to look.

## Still to come on this branch

- `manabrew-server` advertises itself as a relay and serves `/scryfall-img/`, both off unless configured, plus `--download-art`.
- The relay tells clients its art base url, so the client prefers it over the CDN.
- Self-update: manifest poll, minisign verify, swap, exit, reusing `updater.rs`'s existing "defer while a game is live".
- A static musl artifact on release.

## Test plan

- `cargo test -p manabrew-art-cache` — 16 pass, the same tests that were in `src-tauri`, moved unchanged.
- `cargo test -p manabrew-lan-discovery` — 2 new, covering the role default and the hostname.
- `cargo test -p manabrew --features direct-seat` — 3 pass; the shell builds with and without `forge-room`.
- `yarn lint:all` clean.

No behaviour change on the desktop: it keeps the same cache, the same commands and the same LAN sharing, through the extracted crates.


khaliostr:
Superseded by #834, which is this work rebuilt on `main` with no iroh in it. The LAN and art halves never needed the direct data plane: relaying through a box on the same switch already costs nothing, and iroh's win is the WAN case. Decoupling lets this ship without waiting on #831.

PR #834, opened 2026-09-01
feat(relay): a self-hosted relay that holds the card art

khaliostr:
## Summary

A relay you can run yourself: one static Linux binary that is the lobby for a network and holds the card art, so no desktop downloads eight gigabytes of images.

- **`manabrew-art-cache`** — card art on disk, keyed by the path below `cards.scryfall.io`, and a listener that serves it.
- **`manabrew-lan-discovery`** — mDNS advertise and browse. A record says whether it is a `relay` or a `room`.
- **`manabrew-server`** — `MANABREW_ART_DIR` serves `/scryfall-img/`, `MANABREW_LAN_ADVERTISE` answers mDNS, `--download-art` fills the cache and exits.
- **A musl artifact**, a systemd unit and a three-command setup in `ops/selfhost/`.

Supersedes #814 and #832. Independent of #831: none of this needs iroh.

## Why it does not need iroh

The direct data plane turns a WAN game from two hops through Hetzner into one direct connection. On a LAN, relaying through a box on the same switch already costs nothing. `manabrew-art-cache` and `manabrew-lan-discovery` contain zero iroh references, so this ships on its own schedule rather than behind a much larger review.

## Setup

```sh
sudo cp manabrew-server /usr/local/bin/ && sudo systemctl enable --now manabrew-server
sudo -u manabrew MANABREW_ART_DIR=/var/lib/manabrew manabrew-server --download-art
```

Point the desktops at it in Settings and they get the lobby and the art. Games run on whoever hosts the table, from the app that already updates itself, which is why this box needs almost nothing.

## Scope

The desktop half is a follow-up: the `/scryfall-img/` route, the browse command, LAN room sharing, and the lobby preferring a discovered relay so nobody types the address at all. The self-updater is another, and it is inert until the release asset this PR creates exists.

**Everything is off by default.** `MANABREW_ART_DIR` and `MANABREW_LAN_ADVERTISE` are unset unless asked for: the public relay must not quietly become an image CDN.

**`path_for` is the only thing between a network request and the rest of the disk**, since the listener answers whatever the subnet asks. It rejects traversal, absolute paths and every character NTFS refuses.

**mDNS is unsigned**, so anything on a network can offer to be a lobby. Being wrong means an unexpected lobby, never trusting what that host says: identity is still proved to the relay and a private room still needs its password. `ops/selfhost/README.md` says so, and what to do instead on a network you do not control.

**Which art variants get downloaded matters.** The board draws `border_crop` or `art_crop` and never `normal`, so fetching the wrong set leaves it blank offline with a full cache. `--download-art` takes all three because a server cannot know which style each of its players picked.

## Test plan

- `cargo test -p manabrew-art-cache -p manabrew-lan-discovery -p manabrew-server` — 38 pass. The listener answers a real socket for a cached key, 404s what it does not have rather than fetching it, and 404s a traversal; a dropped listener stops answering; the role default; the advertised hostname.
- `yarn lint:all` clean.

**Not verified here:** the musl build runs only in CI, and unlike the xtask asset this graph is not pure Rust (`aws-lc-rs` and `zstd` are C), so it needs `musl-gcc` rather than `rust-lld` alone.


khaliostr:

PR #835, opened 2026-09-01
feat(tauri): use the relay's card art, and find it without being told

khaliostr:
> **Stacked on #834**, which is the server. Review the [compare against that base](https://github.com/witchesofthehill/manabrew/compare/khaliostr/lan-host...khaliostr/lan-desktop).

## Summary

The desktop half: what makes a client use the relay's card art, and find the relay, without anyone typing an address.

- `asset_server` answers `/scryfall-img/` out of this machine's cache, filling from the CDN on a miss.
- `scryfallImageSource` takes the nearest source: here, then a host on this network that already has it, then the CDN.
- `discover_lan_rooms` browses mDNS, and the lobby prefers a discovered `relay` over its configured one.
- `start_local_relay` can share a room on the network, for a group with no server at all.
- Settings gains the card art download panel.

## Why separate from #834

The server stands alone: point the desktops at it in Settings and they get the lobby and the art. This is what removes the "point them at it", and it lives in a different part of the tree.

**The lobby switches automatically.** A discovered `relay` wins over the configured one, decided before the first connection rather than corrected after, so nobody watches it land somewhere and move. Only `role: "relay"` counts; a `room` is one table on a desktop and stays with the unreachable-relay fallback.

The record is **unsigned mDNS**, so anything on a network can offer to be the lobby. That is acceptable because being wrong means an unexpected lobby, never trusting what that host says: identity is still proved to the relay and a private room still needs its password. `src/AGENTS.md` says so next to the behaviour.

**The download panel is mounted this time.** In an earlier shape of this work it was imported by nothing, so it compiled and looked shipped and could never render. Unreferenced code that reads as shipped is worse than absent.

**Which variants get downloaded follows the battlefield style**, because the board draws `border_crop` or `art_crop` and never `normal`. Fetching the wrong set leaves it blank offline with a full cache.

## Test plan

- `cargo test -p manabrew` and `cargo check -p manabrew --features forge-room`.
- `yarn lint:all` clean.
- Manual: two desktops on one network, one hosting, art served from the host.


khaliostr:

PR #836, opened 2026-09-01
feat(relay): let a self-hosted relay update itself

khaliostr:
> **Stacked on #835**, which is stacked on #834. Review the [compare against that base](https://github.com/witchesofthehill/manabrew/compare/khaliostr/lan-desktop...khaliostr/relay-selfupdate).

## Summary

A self-hosted relay keeps itself current, so nobody logs into the box.

## Why

The engine rides the desktop updater and changes daily. A relay in a cupboard does not, and one left behind far enough starts refusing its own clients on `PROTOCOL_VERSION` for no reason any of them can see. That is the real maintenance cost of putting the boring half on a box, and this removes it.

## How

It polls the manifest `self-hosted-node` already watches, and borrows the part of that design worth borrowing: **a stale relay with a game running waits.** A restart under a live table is worse than never updating.

The half that design leaves to a supervisor is here instead, because a stranger's Ubuntu box should not need one: download the published binary, check its minisign signature against the key CI signs with, rename it over this one, exit for systemd. The rename is what makes that safe while running, since the old inode outlives it by the moment it takes to exit.

**Off unless `MANABREW_SELF_UPDATE=1`.** The same binary runs production, which is deployed on purpose and must never pull a version off a manifest by itself. The shipped unit file turns it on; nothing else does.

The trust boundary is one function. `verify` refuses anything the key does not sign, and nothing else stands between a manifest and code running as this service.

One trap found while writing it: `minisign_verify::PublicKey::decode` wants the **whole two-line block**, comment included. The key line alone needs `from_base64`, and passing it to `decode` fails to parse — which would have made every update a silent no-op.

## Test plan

- `cargo test -p manabrew-server self_update` — 3 pass. A version that does not parse is not "behind", because deciding it is would download over this binary. The embedded key decodes. An unsigned download is refused.
- `yarn lint:all` clean.

**Not verified here:** the signing step runs only in CI, and the updater is inert until the release asset #834 creates exists.


khaliostr:

Issue #838, opened 2026-09-01
Epic: peer-to-peer game traffic


khaliostr:
Game traffic between players goes through `manabrew-server`, which makes two WAN hops out of what
is often one network or one city. This epic tracks moving it off the relay where that is possible,
and records where it is not, so the same ground is not re-argued.

With the node fleet retiring (#837), every host is a player's own machine, so this is now the
whole multiplayer data path rather than an optimisation on one corner of it.

## The numbers

313 multiplayer games, 2026-08-20 to 09-01, which is the whole history there is: `client_connected`
telemetry starts on the 20th. Multiplayer is ~10% of games, 24 a day, 2.19 human seats, 27.5
minutes. Counting `pwa` as a browser and resolving each seat by its most recently reported
platform.

| Room | Games | Share | Can be direct via |
|---|---|---|---|
| All browser | 151 | 48.2% | WebRTC only |
| All desktop | 85 | 27.2% | iroh, or the LAN relay when local |
| Mixed | 74 | 23.6% | WebRTC only, and only with work on the host side |
| Unresolved | 3 | 1.0% | |

Caveats: 13 days, 9.3% of players have connected from more than one platform so some seats may be
misattributed, and the host is taken as a room's first human joiner.

## The map

| Pair | Direct? | How | Status |
|---|---|---|---|
| desktop to desktop, same network | yes | embedded relay found over mDNS, one local hop | #834, #835, #836 |
| desktop to desktop, remote | yes | iroh, hole punched | #831, #839, #840 |
| browser to browser | yes | WebRTC | #846 |
| browser host to desktop seat | yes | WebRTC, both ends are webviews | #846 |
| desktop-native host to browser seat | yes, with work | forwarded through the shell | #848 |

## What is settled

**A browser cannot hole punch.** iroh's `build.rs` sets `wasm_browser` for
`wasm32-unknown-unknown` and compiles the IP transports out under it. A browser endpoint can host
a room but every connection is relayed, so browser-to-browser over iroh moves traffic off our
relay without improving latency. WebRTC is the only direct browser transport.

**A LAN needs no iroh.** A seat on the host's network already reaches it in one local hop through
the relay it finds over mDNS, and nothing beats one local hop. The iroh work is for the seat that
is somewhere else.

**iCloud Private Relay defeats the plane.** An iOS seat behind it contributes exactly one
candidate, a proxied reflexive address on a Fastly or Cloudflare egress, and no host candidate, so
[...]

PR #839, opened 2026-09-01
feat(node): carry a seat's engine envelopes over the direct plane

khaliostr:
Stacked on #831. The engine host half of the direct data plane.

`self-hosted-node` offers each seat a QUIC channel and sends that seat's engine envelopes over it
instead of through `BroadcastState`. It is the same crate a desktop app links in-process under
`forge-room`, so this is what makes a desktop-hosted room direct, and
`Config::for_desktop_room` turns it on.

What it wins is a seat that is **not** on the host's network. A seat that is already reaches the
host in a single local hop, through the embedded relay it finds over mDNS, and iroh cannot beat
one local hop. This is for the player in another house, who otherwise takes two WAN hops through
`manabrew.app`.

1,030 lines across 15 files, plus 250 of `Cargo.lock`.

**Falling back is a barrier, not a switch.** While a seat was on the direct plane the relay saw
none of its envelopes, so the host cannot simply resume relay traffic: it owes that seat a full
authoritative state first, or a `stateDelta` patch gets folded onto a base the relay never had.
`SeatTransport::RelayPending` is that debt. It is paid through `on_fallback` into the ordered
outbound queue, before the seat's next envelope, and cleared when the seat answers over the relay,
which is the only signal the host gets that it is reading that path again.

**A superseded connection is closed, not forgotten.** A seat that re-dials leaves its
predecessor's reader running, and that reader holds the other half of the channel guard. Dropping
the `GameSender` is not enough: whoever holds the old connection could keep feeding the engine
responses under that seat's name. There is a test that fails without the close.

**Transport is frozen at `GameStarted`.** Only seats already connected take part, so nothing
migrates mid-game in either direction, and both ends freeze on the same relay message rather than
negotiating.

**The relay is told what it could not see.** `ReportTransport` names the seats that left the relay
data plane and the relay writes it into the game's capture. Without it a capture file is silently
incomplete, which is the same shape of bug as the `stateDelta` measurement trap.

## Relays

The host binds with a configured relay if there is one and iroh's defaults otherwise, so a seat
that cannot reach it directly can still reach it. The control plane can name one later in
`RoomTransport`, and re-adoption is guarded: `insert_relay` schedules a full net report, and the
relay re-broadcasts the roster on every join and leave, so an unchanged relay must not be
re-inserted.

## Off by default, twice

`--features iroh` **and** `SELF_HOSTED_NODE_IROH=1`. Without the feature the crate has no iroh in
its dependency tree at all, not even as a dev-dependency, so no image or CI toolchain moves.
Without the env var a build that has the feature still keeps every seat on the relay. A desktop
room sets it itself.

## Test plan

- `cargo test -p self-hosted-node --features iroh`: five tests over the seat table, no network.
  The superseded connection, the freeze in both directions, the direct to relay-pending to relay
  walk with its re-prime hook, a seat that was never direct owing nothing, and an unchanged relay
  not being re-adopted.
- `cargo build -p self-hosted-node` and `--features iroh`, `cargo clippy --all-targets` on both,
  and `cargo build --workspace` are clean.
- `cargo tree -p self-hosted-node -i iroh` finds nothing with default features.

## Not here

No bot seat: bots run inside forge-wasm now, so `manabot` is not a seat on this path. The desktop
seat is the next PR, and it is what makes an all-desktop room direct end to end. Browser-only
rooms want WebRTC instead, which is #838.


khaliostr:

PR #840, opened 2026-09-01
feat(tauri): a desktop seat that reaches the host directly

khaliostr:
Stacked on #839. The seat half, which closes the loop: with this and the host below it, a desktop
seat plays against a desktop host on a direct QUIC path instead of two WAN hops through the relay.

To be exact about what that wins: a seat on the host's own network already reaches it in a single
local hop, through the embedded relay it finds over mDNS, and nothing beats a single local hop.
This is for the seat in another house.

676 lines across 9 files, and one line of `Cargo.lock`.

## Why the endpoint is native

A desktop binds in the Tauri shell, not in the webview. iroh's `build.rs` sets `wasm_browser` for
`wasm32-unknown-unknown` and compiles the IP transports out under it, so the same seat compiled
into the webview could only ever be relayed. Binding natively is the whole reason a desktop seat
can be direct.

The webview drives it through `direct_seat_*` commands and receives the host's envelopes as a
`direct-seat:envelope` event. Only the seat's own `Response` and `Directive` envelopes move.
Everything else, the entire control plane included, stays on the relay socket the client already
has.

On the web `DirectSeat` binds nothing and returns immediately, so nothing changes there. A
browser room wants WebRTC instead, which is #838.

**The dial is bounded, and the test is the reason.** `direct_seat_roster` is awaited by the
webview, so an unbounded dial hangs the join. The case that does it is a host that is attested but
unreachable: there is a path to try and nothing at the end of it, so the dial neither connects nor
fails fast. `a_dial_that_will_never_land_gives_up_instead_of_hanging_the_join` asserts a ceiling.

**`trySend` is awaited, not raced.** The send crosses into the shell, so the answer only arrives
after the fact. Claiming an envelope the shell then failed to send would drop it on the floor, and
the one most likely to be lost is a prompt response, which leaves the game waiting on that seat
until a resync.

**One bind, not two.** `announce` holds the in-flight promise, because two `RoomTransport`
messages arriving together would otherwise both get past a check that sits before an await. The
second would install a second event listener and every host envelope would arrive twice.

**A seat dials only once it is in the roster it received.** Its own entry proves the relay has
attested it to the host in the same broadcast; dialling before that is refused, correctly.

## Testing

macOS has no WebDriver for WKWebView, so no window can be driven to exercise this. `DesktopSeat`
therefore holds the logic and takes no Tauri types, and the two tests run it against a real host
over a real connection:

- `a_desktop_seat_reaches_a_host_and_carries_envelopes_both_ways` asserts the path is
  `iroh-direct`, that the host accepts the seat under its attested name, and that envelopes cross
  in both directions.
- `a_dial_that_will_never_land_gives_up_instead_of_hanging_the_join`.

`cargo test -p manabrew --features direct-seat --lib` passes both. `cargo check`, clippy, `tsc`
and eslint are clean.

## Not here

LAN play does not need any of this and already exists without it: `manabrew-lan-discovery` plus
the embedded relay. Browser-only rooms want WebRTC, which is #838. Mixed rooms stay on the relay.


khaliostr:

PR #845, opened 2026-09-02
chore(staging): assemble the transport work from #838 for testing

khaliostr:
Integration branch for epic #838. **Do not merge.** It assembles the two review PRs into one build so the transport work can be exercised as one thing, and so CI produces installers for it.

## What is in it

| Line | PR | What |
| --- | --- | --- |
| LAN | #870 | self-hosted relay, card art, mDNS, relay self-update |
| Transport | #871 | iroh direct plane, WebRTC browser plane, desktop host bridge, per-player opt-in |

Plus `main`, kept current. This branch is `main` reset and both PRs merged on top, so it is only ever the sum of those two. When they move, it is re-cut, not patched.

## Why it exists rather than testing each PR

Installers. `publish.yml` only fires on a `v*` tag, so no feature branch produces one. `.github/workflows/desktop-staging.yml` lives **only on this branch**: it is publish.yml's macOS, Windows and Android pair plus the self-hosted relay musl binary, with everything that publishes stripped out (no GitHub Release, no `ops/manifest.json` entry, no Discord). Artifacts live on the workflow run, so an installed app can never be pointed at one by the updater. It triggers on push to `khaliostr/staging-**`.

That workflow does not ride along on the feature PRs, because landing it on `main` would leave a permanent workflow pointed at a branch nobody re-creates. Deployment config is a different matter and ships with the feature: `MANABREW_DIRECT_TRANSPORT`, `MANABREW_ICE_SERVERS` and the STUN-only coturn are on #871's compose files.

## Testing it

The `deploy-preview` label puts this on the staging slot (`staging.manabrew.app`, relay `relay-staging.manabrew.app`), which runs with `MANABREW_DIRECT_TRANSPORT=1` and a STUN server on UDP 3479. One slot, so labelling something else replaces it.

The opt-in is per player: turn on **Settings → Direct connections** on every seat at a table, because one seat without it keeps the whole room on the relay. Then, on the relay's `/metrics`:

```
manabrew_relay_transport_rosters_total{kind="sent"}       climbs once everyone opted in
manabrew_relay_transport_announcements_total{kind="announce"}
manabrew_relay_peer_signals_total{kind="forwarded"}       the browser plane's signalling
manabrew_relay_plane_attempts_total{plane,outcome,pair}   connect rate, by pair kind
```

`withheld` climbing beside a flat `sent` means a table wanted the plane and one seat had not opted in. In the browser console, `[webrtc] … pair=srflx/srflx` is a punched-through pair and `pair=host/host` never left the LAN; the desktop app log shows `seat took the direct plane`.

## Standing cost

Every time #870 or #871 moves, this branch is re-cut from `main` and both are merged again. That is the price of the branch and the reason to close it once the two land.


khaliostr:
Consolidated into #871: LAN and chat are merged into main, #871 is brought up to date and now carries the staging tooling (log overlay + installer workflow) too. One branch. The deploy-preview slot moved to #871.

PR #846, opened 2026-09-02
feat(net): a WebRTC data plane for browser rooms, and relay signalling for it

khaliostr:
The spike from #838. Based on `khaliostr/staging-lan-iroh` (#845) so it can be
exercised against the LAN and iroh stacks together; it needs rebasing onto
`main` or onto #831 to land.

A browser has no IP transports under iroh's `wasm_browser` cfg, so a
browser-hosted room is relayed by construction. WebRTC is the only direct
transport it has, and 48% of multiplayer games are all-browser rooms.

## Relay

`ClientMessage::SignalPeer` and `ServerMessage::PeerSignal` carry an opaque blob
from one room member to a named other. The relay does not read the payload. It
routes on the username, and stamps `from` from its own record of the sending
session, which is the attestation `RoomTransport` already gives the roster. A
blob over `MAX_SIGNAL_BYTES` is dropped: an opaque payload's size is the only
thing the relay can judge it on, and the control plane must not quietly become
a data plane.

Gated on `MANABREW_DIRECT_TRANSPORT` beside `room_transport`. Off, the relay
advertises no `peer_signal` and drops `SignalPeer`, so a client never starts a
negotiation that could not finish. `manabrew_relay_peer_signals_total{kind}`
counts forwarded, or why not.

Seats and the host are addressable. An observer carries no engine envelopes, so
it is not a signalling target even though it may announce an endpoint.

## Client

`WebRtcPlane` holds one `RTCPeerConnection` per peer and one reliable ordered
data channel on each, carrying the same engine envelope set. A seat connects to
the host and to nobody else, because that is the only edge envelopes cross.

Both directions. A browser host serves each seat over its own channel rather
than only receiving their answers, which is where the first iroh draft stopped.

Exactly one end of a pair offers, decided by username order, so two ends cannot
glare. A candidate arriving before the description it would attach to is held,
not dropped: adding it early throws, and it may be the only pair that would have
worked. Both ends freeze on `GameStarted`, the same barrier the iroh path uses,
and fall back to the relay on ICE failure or a closed channel.

## The three design constraints

- **Addressed by username, not peer type.** A desktop seat can use this same
  signalling path.
- **The room's plane comes from what the host advertises.** `TransportEndpoint`
  gains `kinds`; empty means `iroh`, which is what every announcer before the
  field meant. A desktop seat in a browser-hosted room therefore picks WebRTC
  instead of reaching for its native endpoint and finding nothing, and a host
  does not offer to a peer that only speaks iroh.
- **The send seam takes a second sink.** `trySend` returns false and the caller
  puts the envelope on the relay, the same shape `DirectPlane::try_send` has.

## What it measures

Every peer reports its connect outcome, time to open, the median of a short
data-channel ping/pong, and the ICE pair that won as `local/remote` candidate
types. The client's own keepalive round trip to the relay is logged beside it,
which is the number the direct plane has to beat. Timing that Ping is new; the
client sent them but never read the gap.

Probe frames carry a `__probe` discriminator. An engine envelope is always an
object with a `kind`, so the two cannot be confused.

## Spike checklist

- [x] Two tabs, the relay carrying SDP and ICE, an `RTCDataChannel` carrying the
      engine envelopes. Negotiation, both directions, the freeze and every
      fallback are covered by 14 tests against a fake `RTCPeerConnection`.
      Connect rate and RTT are instrumented; the numbers come from running it.
- [ ] `RTCDataChannel` on WebKitGTK, for Linux desktop.
- [ ] Two machines on one switch with the internet off, given Chromium replaces
      host candidates with mDNS names.

The last two need real hardware.

## Not in scope

A desktop-native host serving a browser seat. Its engine is `self-hosted-node`
in-process and owns its own relay socket, so those envelopes never pass through
the webview. Forwarding them through the shell is the cheap way in and
`DirectPlane::try_send` returning false is the seam, as #838 records.

## Checked

`cargo clippy` and `cargo fmt` clean; relay tests pass. `yarn lint`,
`yarn typecheck` and the protocol docs drift check clean. 14 new vitest cases,
142 passing overall. `useScryfallStore.test.ts` fails to collect on
`__APP_VERSION__` locally, the same way it does on plain `main`.


khaliostr:

PR #848, opened 2026-09-02
feat(node): carry a browser seat's envelopes through the shell

khaliostr:
The rest of the mixed room from #838, stacked on #846.

51 of the 74 mixed games were desktop-hosted, so this is the larger half. #846
covered the browser-hosted third for free, because both ends there are
webviews.

## The problem

A desktop host runs its engine in-process, on its own relay session
(`forge-host-<uuid>`), and cannot dial a browser seat. A browser is reachable
over WebRTC and nothing else, and the only thing in that process that can make
a WebRTC connection is the webview sitting beside it. So the envelopes go out
through the shell rather than through a second Rust transport stack, which is
what the epic settled on.

## The seam

`ShellBridge` is a third sink, after the native endpoint and before the relay:

```
engine -> DirectPlane::try_send  (iroh, a desktop seat)
       -> ShellBridge::try_send  (the webview, a browser seat)
       -> the relay              (everything else, always)
```

Four things cross it. Outward: an engine envelope for a seat, and signalling
addressed to the host. Inward: the seats the webview currently reaches,
signalling to send, and a seat's own envelope, which takes the same route into
the engine that a relay `StateUpdate` takes.

## Who holds what

The relay session stays in the node. `ForgeHostBridge` runs a `WebRtcPlane`
under the node's username, so it reads itself as the room's host and every host
rule from #846 applies unchanged, but it never gets a relay session it could
speak for the host with. Signalling arrives on the node's socket, is forwarded
out, and what the webview answers goes back and is sent under the host's own
attested identity.

The freeze stays in the node too. `freeze_for_game` runs after the direct plane
has taken its seats and is told which those were, so no seat is claimed by two
planes. The webview only reports which channels are open, and delivers what it
is handed.

## The window this cannot close

`ShellBridge::try_send` cannot know the send landed: the answer would have to
come back across the shell, and the caller is the room's message loop. A seat
whose channel dies between the webview's last report and that call loses
envelopes.

What repairs it is the debt the direct plane already has. A seat that leaves
this plane is owed a full board before it reads the relay again, paid through
`on_fallback` into the ordered outbound queue and cleared when the seat answers
over the relay. A seat that leaves does not come back for the rest of the game,
in either direction. A test covers that last rule: settling the debt must not
put the seat back on the plane, and neither may the webview reporting it
reachable again.

## One roster, both kinds

A desktop host now announces `kinds: ["iroh", "webrtc"]`, most preferred first,
and a seat takes the first of them it can itself speak (`planeForRoom`). Its
desktop seats land on iroh and its browser seats on WebRTC, out of the same
roster, and neither decides from its own platform. This replaces #846's
"WebRTC wins if the host has it", which would have pulled a desktop seat off
iroh in a desktop-hosted room.

## Checked

`cargo clippy` and `cargo fmt` clean; `src-tauri` builds both with and without
the `forge-room` feature, since `self-hosted-node` is only a dependency under
it. Rust tests pass, including 7 new ones on the bridge. `yarn lint` and
`yarn typecheck` clean, 20 vitest cases on the plane and 148 passing overall.
`useScryfallStore.test.ts` fails to collect on `__APP_VERSION__` locally, the
same way it does on plain `main`.

Still needing real hardware, as on #846: measured numbers from two browsers,
`RTCDataChannel` on WebKitGTK, and two machines on one switch with the internet
off.


khaliostr:

PR #862, opened 2026-09-04
feat(net): report how a direct plane attempt turned out

khaliostr:
Closes the measurement hole that testing #848 exposed. Part of #838.

The feature working is what removes the evidence that it works. The relay times
a seat's round trip from that seat's own traffic, so a seat that leaves the
relay stops being measurable. Game `5a900dc9`, where both seats went direct,
produced **zero** `clientRttMs` samples; the same two players on the relay the
day before recorded 56ms and 504ms.

`ReportTransport` cannot fill it. The host sends it once at `GameStarted` and it
names only the seats that succeeded, so a seat that fails to punch through stays
on the relay and tells nobody it tried. That leaves a connect rate with no
denominator — and the connect rate is what #838 says should decide whether
production runs ICE servers.

So the measuring end reports its own attempts, failures included.

## What it sends

`ReportPlaneQuality` carries outcome, time to open, median channel round trip,
and the same session's round trip to the relay alongside it. Paired
deliberately: a direct RTT on its own cannot say whether it beat the path it
replaced.

## What the relay trusts

Nothing in it. Same discipline as `SignalPeer`:

- the reporter is named from the relay's own record, never from the message
- the peer must be someone the relay placed in the same room
- outcome and plane come from fixed sets, so a client cannot pick this metric's
  cardinality
- timings past ten minutes are dropped rather than averaged in

## Counted once

A connected peer reports twice — when the channel opens, and again with the
round trip. Without a `phase` the counter would score every success twice and
every failure once, inflating the exact rate this exists to measure.

## Buckets

`candidate_pair` is bucketed to `lan`/`punched`/`mixed`/`turn` for the metric
while the capture keeps the exact pair. Both successes measured so far were
`host/host`, which never left the network. A rate that cannot separate those
from punched-through pairs would flatter itself.

## Metrics

```
manabrew_relay_plane_attempts_total{plane,outcome,pair}
manabrew_relay_plane_rtt_ms{plane}
manabrew_relay_plane_relay_rtt_ms{plane}
manabrew_relay_plane_connect_ms{plane}
```

Behind a `plane_quality` feature flag, so an older relay is never sent a message
it would reject as a parse error.

## Scope

Wires the WebRTC plane, which is where the measurement already exists. The iroh
seat measures nothing today and is not covered here.

### DIFF
--- a/manabrew-rs/crates/manabrew-relay-protocol/src/lib.rs
+++ b/manabrew-rs/crates/manabrew-relay-protocol/src/lib.rs
@@ -132,6 +132,67 @@ pub struct SeatTransportReport {
     pub transport: String,
 }
 
+/// One end's account of one attempt to reach a peer off the relay.
+///
+/// Every field past `peer` and `outcome` is the reporter's own measurement, so
+/// the relay treats all of it as a claim: it bounds the numbers, caps the
+/// strings, and records who said it rather than trusting what was said.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct PlaneQualityReport {
+    /// The other end, as the relay attested that username to this session.
+    pub peer: String,
+    /// [`PLANE_OUTCOME_CONNECTED`], [`PLANE_OUTCOME_FAILED`] or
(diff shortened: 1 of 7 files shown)


khaliostr:

PR #864, opened 2026-09-04
fix(engine): expand adapt and monstrosity shorthand into counter params

tazzuu:
## Summary

- fixes the Adapt ability which was previously broken, so that it puts the correct number of counters
- updates the arg parsing for Adapt and Monstrosity since they are similar

## Why

parity for Adapt was breaking

## Test plan

```
yarn parity:test -- --matrix --seeds 42,100,999 --decks red_burn,green_stompy,white_aggro,black_control,comprehensive_test,trigger_expanded,staticability_test
```


anacletos:
Hey @tazzuu welcome to the repo!
We're happy to get this in.

However I should sort of warn you, the parity system is currently not being worked on, as the forge controller isn't quite rule correct yet. Parity work at this stage might be wasted effort until a solid forge oracle is in place. I wouldn't want to have people invest valuable time in it until then.

anacletos (approved):

PR #866, opened 2026-09-04
fix(play): let a browser host concede instead of destroying the table

khaliostr:
A game hosted in a player's own browser had exactly one exit for its host: shut
the engine down under everyone. The relay then forfeited the seat like any
other, left the room in-game with no engine behind it, and the room aged out as
`abandoned`.

That is where browser-hosted multiplayer loses its games. Over 09-02 to 09-04 on
prod:

| host | games | finished | abandoned |
|---|---|---|---|
| node | 40 | 35 (88%) | 1 |
| browser | 23 | 12 (52%) | 9 |

A node-hosted room does not lose those games, because a player quitting there is
just a concession and the engine is still around to declare a winner.

**The host's deliberate exit is now a concession.** With at most one opponent
still playing, `LeaveGameModal` offers Concede: the engine is still in the tab,
so it resolves the game, names the winner, and everyone gets a real game over.
A table that would carry on without the host still has to be torn down, because
there is no result to reach and the engine cannot move to another seat.

**The involuntary exit is now recorded for what it is.** A peer-hosted room that
loses the seat holding its engine gets the same reconnect grace a node host
gets (`cleanup.rs::schedule_engine_host_abort`), then ends as `host_lost` and
resets to its lobby, instead of stranding the survivors in a game nothing can
advance until the room empties.

Not in scope: moving a live game to another seat or to a node. That needs Forge
to hand its state across the wasm boundary, which does not exist yet, and it is
the only thing that would save the games where two or more opponents are left.

Checks: `yarn lint` clean, `cargo clippy -p manabrew-server --lib --bins -D warnings` clean.


manabrew-rs/crates/manabrew-server/src/cleanup.rs
```diff
@@ -163,6 +163,65 @@ pub fn schedule_host_resume_abort(
...
+/// The engine host of a peer-hosted room holds a seat, so its disconnect used
+/// to be handled as an ordinary seat forfeit: the seat went, the room stayed
+/// in-game, and the engine that ran in that player's tab was gone. Guests then
+/// waited out their own timeout, left, and the room aged out as `abandoned` —
+/// a game recorded as if nobody had been there, when in fact the host left.
+///
+/// Give it the vanished-host treatment instead: the same reconnect grace, then
+/// end the game as `HostLost` and hand the room back to its lobby so the
+/// survivors can start another one.
+pub fn schedule_engine_host_abort(state: Arc<ServerState>, room_id: String, player_id: String) {
```
anacletos, line 175: mmh Im not sure I understand why this is needed

khaliostr: The concede button only covers the host who chooses to leave. This covers the host who does not: tab closed, browser crash, laptop shut. The engine dies with the tab either way, and nothing on the client is left to send `EndGame`.

Before this the relay treated that like any other seat: `schedule_seat_forfeit` removed the host's seat after the reconnect grace and left the room `InGame`. The guests sat in a game with no engine behind it until their own timeout, then left, and the room aged out as `abandoned`. That is the 9 of 23 browser-hosted games in the table above.

This is the same treatment `schedule_host_resume_abort` already gives a node host that never comes back: same grace, then end the game as `HostLost` and reset the room so the survivors can start another one. The host's seat is the engine on a peer-hosted room, so losing it is losing the game, not losing a player. It also means the outcome files as `host_lost` instead of `abandoned`, which is what lets us see the loss in the analytics.

anacletos: Can you please not claude back at me it's annoying

anacletos: It doesn't make sense to me to "abort" a client engine.
There is no engine to abort once they are gone.

anacletos: As far as you know, they could have disconnected, which the normal timeout covers

khaliostr: sorry buddy it was automatic :(

khaliostr: i'll look into this in more detail

khaliostr: af as i understand on main the normal disconnect/timeout _forfeits_ the seat. If it's a non-host is no problem, but if a host forfeits the game continues without the host which is UB.

PR #870, opened 2026-09-05
feat(relay): a self-hosted relay that holds the card art, and the desktop that finds it

khaliostr:
Consolidates #834, #835 and #836, which are closed in favour of this one. Independent of the p2p PR (#871): none of this needs iroh. #845 stays open as the staging assembly of both; do not merge it.

## What this is

A relay you can run yourself, and the desktop half that finds it.

- **`manabrew-art-cache`**: card art on disk, keyed by the path below `cards.scryfall.io`, and a listener that serves it.
- **`manabrew-lan-discovery`**: mDNS advertise and browse. A record says whether it is a `relay` or a `room`.
- **`manabrew-server`**: `MANABREW_ART_DIR` serves `/scryfall-img/`, `MANABREW_LAN_ADVERTISE` answers mDNS, `--download-art` fills the cache and exits. `MANABREW_SELF_UPDATE=1` lets it update itself off the release manifest, minisign-verified, and it waits while a game is running.
- **A musl artifact**, a systemd unit and a three-command setup in `ops/selfhost/`.
- **Desktop**: `asset_server` answers `/scryfall-img/` from this machine's cache; `scryfallImageSource` takes the nearest source (here, then a host on this network that has it, then the CDN); the lobby prefers a discovered `relay` over the configured one; `start_local_relay` shares a room on a network with no server at all; Settings gains the card art download panel.

## Why it does not need iroh

On a LAN, relaying through a box on the same switch already costs nothing; the direct plane is for the seat in another house. `manabrew-art-cache` and `manabrew-lan-discovery` contain zero iroh references.

- **Everything is off by default.** `MANABREW_ART_DIR`, `MANABREW_LAN_ADVERTISE` and `MANABREW_SELF_UPDATE` are unset unless asked for: the public relay must not quietly become an image CDN or pull a version off a manifest by itself. The shipped unit file turns the updater on; nothing else does.
- **`path_for` is the only thing between a network request and the rest of the disk.** It rejects traversal, absolute paths and every character NTFS refuses.
- **mDNS is unsigned.** Anything on a network can offer to be a lobby. Being wrong means an unexpected lobby, never trusting what that host says: identity is still proved to the relay and a private room still needs its password. `ops/selfhost/README.md` and `src/AGENTS.md` say so.
- **The updater's trust boundary is one function.** `verify` refuses anything the key does not sign. `minisign_verify::PublicKey::decode` wants the whole two-line block, comment included; the key line alone needs `from_base64`, and getting that wrong makes every update a silent no-op.
- **Which art variants get downloaded follows the battlefield style.** The board draws `border_crop` or `art_crop` and never `normal`; fetching the wrong set leaves it blank offline with a full cache. The server takes all three because it cannot know what each player picked.

## Test plan

- `cargo test -p manabrew-art-cache -p manabrew-lan-discovery -p manabrew-server`: the listener answers a real socket for a cached key, 404s what it does not have rather than fetching, 404s a traversal; a dropped listener stops answering; the role default; the advertised hostname; the updater refuses an unsigned download and a version that does not parse.
- `cargo test -p manabrew` and `cargo check -p manabrew --features forge-room`; `yarn lint:all` clean.
- Manual: two desktops on one network, one hosting, art served from the host.
- **Not verified here:** the musl build and the signing step run only in CI, and the updater is inert until the release asset this PR creates exists.


fedepoi (dismissed):
you have 1 conflict !, overall looking good

PR #871, opened 2026-09-05
feat(net): peer-to-peer game traffic over WebRTC, opt-in per player

khaliostr:
Closes the review split of #838 into one PR. Consolidates #831, #839, #840, #846, #848 and #862, closed in favour of this one. LAN (#870) and chat-and-invites are merged into `main` and this branch is up to date with them; #845 (the old staging assembly) is closed and its staging tooling now lives here.

## What this is

Game traffic between the players in a room goes through `manabrew-server`: two WAN hops out of what is often one network. This lets it go straight between peers instead, behind the relay as rendezvous. Every seat takes the same direct transport, WebRTC: a browser drives it, and a desktop drives it from the webview beside its engine. The relay keeps everything it is trusted for and stays the data plane for every seat that does not, or cannot, leave it. `docs/TRANSPORT.md` has the rules and the numbers.

| Piece | Where |
| --- | --- |
| Rendezvous: `AnnounceTransport`, `RoomTransport`, attested rosters | relay |
| Browser plane: WebRTC data channel, relay signalling (`SignalPeer`) | `src/game/webrtcPlane.ts`, relay |
| Host: `self-hosted-node`, WebRTC driven from the webview | node, `shell_bridge.rs` |
| Desktop host serving a seat through the webview | `shell_bridge.rs`, `forgeHostBridge.ts` |
| Measurement: `ReportTransport`, `ReportPlaneQuality`, plane metrics | relay |

### WebRTC for every seat

This branch was originally an iroh native plane for desktop seats plus WebRTC for browsers. A mixed room broke it: a desktop-native host with one iroh seat and one WebRTC seat never delivered cards or the mulligan prompt. So every seat now takes WebRTC and the iroh plane is removed here, constants and roster fields included. Both transports punch cross-network (measured), so this is a simplification, not a capability loss. The iroh plane is preserved as a separate draft, **#880**, stacked on this branch, if the desktop latency win is ever worth reviving.

Do note, PR line size contains ~1.5K of Cargo.lock (mostly deletions, from dropping the iroh dependency tree).

## Opting in

Two switches, both fail closed.

- **Deployment.** `MANABREW_DIRECT_TRANSPORT` is off in the binary, on in `compose.production.yml` and `compose.staging.yml`. Off, the relay advertises neither `room_transport` nor `peer_signal` and no room can leave the relay.
- **Player.** "Peer to Peer" in Settings, off by default. Ignored on a relay found on the local network: a LAN table stays on that relay for now. Announcing is how a player opts in; a client with it off announces nothing and dials nobody. A desktop hosting a room with it off gets no shell bridge.

**A room upgrades only when every player in it opted in.** The relay enforces it: the roster goes out with a host only once every human seat has announced, and goes out empty otherwise, including the moment a seat that has not opted in sits down. An empty roster hangs up whatever was dialled earlier, and the host freezes only seats the roster names. Bots never count. A leave or a disconnect re-broadcasts, because the leaver may have been the one seat holding the table back. `manabrew_relay_transport_rosters_total{kind="withheld"}` counts tables that wanted the plane and did not get it.

- **The roster is the whole admission decision.** Usernames in it come from the relay's session records, never from a client. Signalling is keyed on that same attested name.
- **A refused announcement is silent.** A squatter must not be able to show a real player an error about the room they sit in. `..._announcements_total{kind="rejected"}` is where it shows.
- **Transport freezes at `GameStarted`** and never migrates mid-game. Falling back is a barrier: the relay saw none of a direct seat's envelopes, so the host owes it a full board first (`SeatTransport::RelayPending`), cleared when the seat answers over the relay.
- **A headless node offers no plane.** WebRTC needs a webview; a fleet node has none, so its rooms stay on the relay. Only browser-hosted and desktop-webview-hosted rooms upgrade.
- **A superseded seat connection is closed, not dropped.** `WebRtcPlane` closes the peer it replaces so a stale channel cannot keep feeding the engine under that seat's name.
- **Signalling is bounded.** The relay routes on username, stamps `from` itself, drops anything over `MAX_SIGNAL_BYTES`, and labels come from fixed sets so a client cannot pick a metric's cardinality.
- **STUN, no TURN.** `compose.production.yml` gains a STUN-only coturn on UDP 3478, host networking on purpose. Staging's moves to 3479: the two slots share the box. TURN is the relay's two hops again and the epic rejected it.

## Follow-ups

- Record in the metrics whether a user (or guest) has the peer-to-peer setting on, independent of whether their room upgraded. Today only the room-level roster counters exist.

## Test plan

- `cargo test -p manabrew-server -p self-hosted-node`: the roster rules, the consent rule, the freeze, the fallback barrier, signalling bounds. No `--features iroh` needed anymore.
- `yarn test`: 25 cases on the browser plane against a fake `RTCPeerConnection`, including the empty-roster hang-up and a host serving each seat over its own channel.
- **Networking regression** (`yarn test:networking`): the transport-agnostic relay scenarios against the real relay and node. The flag failing closed; signalling routed, attested and bounded, checked on the relay's own counters; a room with one seat that never opted in staying on the relay for everyone. The iroh-only game scenarios moved to #880.
- Measured on staging, two home lines on different ISPs: 16ms on the data channel against 62ms through the relay, `srflx/srflx`. Cellular and iCloud Private Relay stay on the relay by design; the docs say why.

---

## The single transport branch (staging bits)

This branch also carries the staging tooling that used to live on #845, so one branch reviews and deploys the staging slot.

### Debug tooling to discount in review

Two blocks here are staging-only scaffolding, gated off by default. Please don't count them against this change:

- **In-app log overlay** (`src/components/dev/DebugLogOverlay.tsx`, ~232 lines) plus its mount in `App.tsx`, gated behind `VITE_STAGING_TOOLS` (set only by the staging installer build). Undefined in every normal build, so it tree-shakes out.
- **Prompt-hop logging** (in `shell_bridge.rs`, `forgeHostBridge.ts`, `webrtcPlane.ts`): one line per prompt at each transport hop, to trace where a start-of-game prompt is lost. Only fires when the direct plane is active, which is opt-in and off in production.
So roughly **250 lines are staging scaffolding**, not part of the transport feature.


src/views/Settings.tsx
```diff
@@ -1053,6 +1053,28 @@ export default function Settings() {
               </PreferenceCard>
             )}
 
+            <PreferenceCard
+              title="Direct connections"
+              description="Send game traffic straight to the other players instead of through the relay. A table only goes direct when every player at it has this on; otherwise everything stays on the relay as before. Your address is shared with the other players in the room."
+            >
```
anacletos: Maybe
```
Peer to Peer
Skip manabrew servers and enable direct connection with the other players at the table. 
This will share your ip-address with people playing with you
and will only activate if all players in a game have this setting enabled.
```

khaliostr: Done, took your wording. Title is "Peer to Peer", the IP and all-players caveats are in the description.

JacopoMadaluni (approved):

PR #887, opened 2026-09-08
fix(engine): report engine crashes instead of calling the game a draw

khaliostr:
## Report

Three Commander games against the AI in a row ended in a "draw" with Coram, the Undertaker on the battlefield. The game was nowhere near a draw.

## What happened

Forge threw inside `CardProperty.cardHasProperty` while applying Coram's `Card.nonLand+ThisTurnEnteredFrom_Library` static to an LKI copy of a spell (`getAlternativeCosts` runs every static against a copy that has no current zone). The harness caught the exception and printed it, `startGameJson` returned, the wasm transport published a final board with no winner, and `GameOverScreen` renders a game over with no winner as "Draw!". Nothing was reported anywhere.

Reproduced on the JVM harness with the reporter's deck, four AI seats: two of three seeds crashed within 20s.

## Changes

- **Forge** (fork PR witchesofthehill/forge#9, pinned at `86d8c9bbf55`): `ThisTurnEnteredFrom` reads `getLastKnownZone()`. All three seeds play to game over after it.
- **Harness**: the session keeps the crash (`getEngineError`); `SabTransport.publishGameOver` sends it to every seat as `{"kind":"error","error":{"code":"engineCrash","message":...}}` ahead of the final board.
- **Client**: `engineCrash` in the game store; the game over screen says "The engine crashed" and shows the exception; the offline play record files `end_reason: engine_error` with `engineError` next to the decks; engine stats file `endReason: error`.
- **Hub**: `offline_play_games.engine_error` (migration 25), so a crash arrives with the decklists needed to reproduce it.

## Not covered

- Node-hosted games: the session records the crash too, but the node's protocol adapter does not forward it yet.
- A JS-level trap in the Web Image (stack overflow) still goes through `worker_error` and the failed-game screen.
- `engineCrash` is not in the Rust `ProtocolErrorCode` enum; the client matches it as a string.

## Order

Merge the fork PR first and keep `manabrew-staging` at the same head.

### DIFF
--- a/manabrew-rs/crates/manabrew-protocol/src/telemetry.rs
+++ b/manabrew-rs/crates/manabrew-protocol/src/telemetry.rs
@@ -137,9 +137,14 @@ pub struct OfflinePlayGame {
     pub format: Option<String>,
     pub engine: String,
     pub starting_life: i32,
-    /// The relay's vocabulary, not the client's: `game_over`, `abandoned`.
+    /// The relay's vocabulary, not the client's: `game_over`, `abandoned`,
+    /// `engine_error`.
     pub end_reason: String,
     pub game_over: bool,
+    /// The engine's description of the crash that ended the game, when one did.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub engine_error: Option<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[ts(optional)]
     pub winner: Option<String>,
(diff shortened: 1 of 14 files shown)


fedepoi:
@khaliostr do we still need this now that builds are stable?

manabrew-rs/crates/manabrew-protocol/src/telemetry.rs
```diff
@@ -149,9 +149,14 @@ pub struct OfflinePlayGame {
     pub format: Option<String>,
     pub engine: String,
     pub starting_life: i32,
-    /// The relay's vocabulary, not the client's: `game_over`, `abandoned`.
+    /// The relay's vocabulary, not the client's: `game_over`, `abandoned`,
+    /// `engine_error`.
```
anacletos, line 153: Friendly reminder this ends up in the docs

khaliostr: nice catch, thanks

PR #895, opened 2026-09-09
feat: texture on tables

JacopoMadaluni:


manabrew-rs/crates/self-hosted-node/src/config.rs
```diff
@@ -146,6 +148,7 @@ impl Config {
     /// reuse the caller's relay connection, host the engine without taking a
     /// seat, no bot. Deck fields are placeholders — unused when `host_plays` and
     /// `bot_enabled` are false.
+    #[allow(clippy::too_many_arguments)]
```
fedepoi, line 151: really?

fedepoi (approved):
looking good

PR #847, opened 2026-09-02
fix(forge-wasm): include named card dependencies

nickpalladino:
## Summary

- Include card scripts referenced by `ChooseFromList` when the Forge WASM asset selector builds a game bundle.
- Preserve Forge's semicolon escape for card names that contain commas.
- Verify Garth's six choices, comma restoration, and unrelated-card pruning at the npm package boundary.

## Why

The browser bundle previously included only cards present in the two decks. Garth One-Eye references six other cards through `ChooseFromList`, so Forge received null card faces while resolving the ability. The interactive session then ended without a winner, and the client displayed that result as a draw.

Fixes #843

## Test plan

- `corepack yarn build:forge-wasm-package --stub-engine`
- `corepack yarn verify:forge-wasm-package`
- `corepack yarn build:wasm`
- `corepack yarn lint:all`
- Manual browser test: started a local Commander game against Forge AI, activated Garth One-Eye, passed priority, and confirmed the chooser appeared without ending the game in a draw.

No parity regression entry was added because this change affects browser asset selection. It does not change Forge Java or Manabrew engine behavior.

PR #899, opened 2026-09-10
fix(forge-wasm): fetch card scripts the boot bundle left out

### LINKED ISSUE #872: Bugfix - Forge wasm can miss scripts on lazy lookup
## Problem
The archive is partially passed to wasm when initializing the engine.
This means that a lazy load that doesn't have the script inside will fail.

This PR attempts a more comprehensive archive init https://github.com/witchesofthehill/manabrew/pull/847
but it's not a complete fix. The complete fix is to have a system for which when the engine doesn't have the file, it will send a message outside the worker to fetch it lazily from the main archive.

This only applies to forge wasm.

khaliostr:
## Summary

- The wasm engine asks the page for a card script its boot bundle left out, over the host seat's buffer, the way it asks a prompt. The page answers from the complete card archive it already holds.
- `forge_card_scripts` in `forge-cardset-archive` answers that lookup with the card and whatever it names; both wasm crates export it.
- Boot and play-time selection recursively include `ChooseFromList`, `CopyFaceFrom`, and `MeldPair` dependencies.
- The archive now carries every Forge card script and every rules/card-data resource staged by the native headless runtime, including binary resources byte-for-byte.
- The seat module gains `answerSeatAssets`, the npm package a `cardScripts` option for hosts that override `assets`, and the fork's `CardStorageReader` the hook (witchesofthehill/forge#10, pinned here).

## Why

#847 taught the bundle to follow `ChooseFromList`, which covers 7 scripts. Scripts that reach for another card by name when an ability resolves are far more common: `Conjure` 188, `Spellbook` 62, `DefinedName` 61, `NamedCard` 34, meld 14. Forge's lazy reader indexes the VFS once at boot, so a file written later is invisible; the lookup has to happen on the reader's miss path, and the worker thread is parked in `Atomics.wait` for the whole game, so the request has to travel over the seat buffer.

In a browser-hosted table the request always uses the local seat, never a buffer relayed to a guest. Each name is asked for once, and a name the archive does not know is remembered as missing.

The asset audit also found that `draft`, `cube`, `setlookup`, and `sealed` were staged by native Forge but absent from the wasm archive. Fourteen binary resources under `defaults` and `effects` were silently skipped. The archive now preserves those files and the package verifier compares all 33,645 card scripts and all 3,351 native runtime assets byte-for-byte against the pinned Forge tree.

Closes #872.

## Test plan

- `yarn lint:all`
- `cargo test -p forge-cardset-archive`
- `node scripts/build-forge-wasm-package.mjs --stub-engine && node scripts/verify-forge-wasm-package.mjs`
  - verifies all 33,645 card scripts and 3,351 native runtime assets byte-for-byte
  - verifies boot and play-time lookups for `ChooseFromList`, `CopyFaceFrom`, and `MeldPair`
- Compile `WasmMain.java` and exercise `writeFramed` with text and binary payloads under JDK 21.
- `scripts/build-forge-wasm.sh` against the pinned fork, then a full browser game on the dev server: an AI deck of 36 Norn's Fetchling and 24 Adarkar Wastes against Izzet Lessons, so no deck holds a Plains. The Fetchling's trigger conjures one on turn 4; the engine logged `fetched the script of "Plains" from the host, 1 in the answer`, exactly one `asset` frame crossed the seat, and the game ran to game over on turn 11.

No parity regression entry: this changes asset delivery to the browser engine, not Forge's rules or the Rust engine.


fedepoi:
Could you add automated regression coverage for the two startup dependency paths exposed while testing #906?

- `CopyFaceFrom`: load Studious First-Year with Rampant Growth absent from the boot bundle. Assert that the missing-card source supplies Rampant Growth and `CardRules.supplyPlaceholderFaces` completes.
- `MeldPair`: construct a one-sided meld card with its counterpart absent from the boot bundle. Assert that lazy fetch supplies the counterpart and `CardFactory.readCard` does not receive a null alternate face.

The current automated selector check covers Garth/`ChooseFromList`, and the manual browser scenario covers `Conjure`, but neither locks these startup paths. They take different routes through `StaticData.attemptToLoadCardInner` and `CardFactory.readCard`, and both caused game-start failures during #906 verification.

PR #900, opened 2026-09-10
fix(play): hold the back button during a live game

### LINKED ISSUE #892: triage - No Way to Rejoin Table
I know it's my fault because I have the back button on my mouse configured by default to actually be a back button. I keep forgetting to switch to my Magic profile on my mouse that disables the back button.

It is really easy to press and I accidentally press it all the time while playing. It would just be nice if we had someway to rejoin the table if and when it does happen, because I'm sure I'm not the only one this happens to or will happen to.

### Engine
**Forge**

---
Reported by **devon7654** in the Issue Tracker Discord forum: https://discord.com/channels/1502665088304419008/1547072114161356830

khaliostr:
## Summary

- Browser back and forward during a live game open the leave dialog instead of leaving the board. `useLiveGameNavigationGuard` is a react-router POP blocker mounted in `Game`; the app's own navigations are not held.
- `Play` no longer relaunches a game that is already running when its history entry is re-entered. The store refused the duplicate and that refusal read as a failed start, so the host ended the table.
- The top bar hides only while the board is on a `/play` route. A live game on any other route keeps a way home.
- `LeaveGameModal` has copy for a guest seat and for a solo game.

## Why

Closes #892. A mouse side button mid-game unmounted the board and hid the nav with it, so the player was stranded. Browser forward then re-mounted `Play`, which relaunched the game and ended the table for everyone. The relay was never the problem: it keeps the seat for the reconnect window.

## Test plan

- Playwright, host plus one bot, Forge in the browser, Community page as the previous history entry.
  - On `main` against staging: back lands on the Community page with no top bar; forward sends `EndGame` and the table is gone.
  - On this branch (local build, staging relay): back keeps `/play` and the board, and shows the dialog. Stay dismisses it, forward is a no-op. Back again and Leave lands in the lobby with one `LeaveRoom` and no duplicate start.
- `yarn lint` clean.

## Demo

Before, after one back press: no board, no top bar.

before

After, same press:

after


fedepoi (approved):
approved but a bit too many comments, can we please be careful with comments in general

PR #912, opened 2026-09-12
fix: safari-wasm and stop proxying images

JacopoMadaluni:
## Summary

* Fix safari, app now runs on `require-corp` rather than `credentialless`
* Stop proxying image data through our reverse proxy
* Host non-cors assets through R2. Scryfall recommends caching this anyway, so we aren't violating any terms by doing this.


.github/workflows/docker-images.yml
```diff
@@ -60,6 +60,7 @@ jobs:
             # every web image, so this is safe for self-hosters too.
             build-args: |
               VITE_SCRYFALL_SYMBOL_BASE=/scryfall-symbols/
+              VITE_SCRYFALL_ASSET_BASE=${{ vars.SCRYFALL_ASSET_BASE }}
```
fedepoi, line 63: why this comes from vars and VITE_SCRYFALL_SYMBOL_BASE does not?

.github/workflows/staging-deploy.yml
```diff
@@ -76,6 +76,7 @@ jobs:
             dockerfile: Dockerfile.web
             build-args: |
               VITE_SCRYFALL_SYMBOL_BASE=/scryfall-symbols/
+              VITE_SCRYFALL_ASSET_BASE=${{ vars.STAGING_SCRYFALL_ASSET_BASE }}
```
fedepoi, line 79: same question here

fedepoi (dismissed):

.github/workflows/docker-images.yml
JacopoMadaluni: mmh cos it doesn't depend on the person who builds the service.
It's hardcoded cos it's just scryfall suffix

.github/workflows/staging-deploy.yml
JacopoMadaluni: Same answer above

fedepoi (approved):

PR #873, opened 2026-09-05
feat(ui): redesign the battlefield

fedepoi:
### Battlefield and player state

- Add focused and overview layouts for multiplayer battlefields.
- Focus opponent fields from the active turn, combat assignments, pointer hover, keyboard shortcuts, and player inspection.
- Keep HUD bounds stable so hover animation and player-panel changes do not move battlefield cards.
- Scale each battlefield to its available height, reserve combat-row space, and contain opponent zone geometry.
- Redesign player HUDs around life, mana, counters, commander damage, combat, control, priority, and turn rules.
- Put library, graveyard, exile, and command zones on the battlefield with counts, top cards, empty states, commander tax, movable tiles, and lock controls.
- Animate card movement between the battlefield and player zones, including lifted and tilted drag feedback.
- Add attacker rows, blocker assignments, combat summaries, lethal damage cues, and a detailed combat breakdown dialog.
- Move turn actions, board settings, game-menu actions, and combat access into the board action cluster with mobile layouts.
- Add player sheets and a global state rail for public zones, player rules, dungeons, day and night, active planes, schemes, teams, and extra turns.

### Prompts and game flow

- Replace the React prompt controller, prompt-specific modals, CSS dice, and game-over screen with one Pixi prompt surface.
- Separate the public prompt orchestrator, shared prompt and drag infrastructure, and modal renderers so action-cluster and modal work no longer collide in one 6,000-line file.
- Render priority, mana payment, mulligan, attackers, blockers, targets, damage order, boolean, list, color, number, card-selection, reveal, reorder, scry, dice, and game-over flows in the board canvas.
- Add responsive full and compact prompt layouts, prompt hide and reopen controls, source-card context, reduced-motion handling, and keyboard and touch activation.
- Add deterministic Pixi dice animation for normal and planar rolls, including multiple players, rounds, ignored rolls, and modified results.
- Rework modal lifecycle around a shared nested stack with focus containment, topmost Escape handling, backdrop dismissal, and transition-safe close actions.
- Add staged loading progress, rotating desktop and touch tips, current shortcut labels, stuck-game diagnostics, and copyable communication logs.

### Card browsing, previews, and input

- Add a shared card browser for visible zones, stack inspection, and card prompts with search, type and color filters, sorting, virtualized grids, keyboard navigation, and preserved scroll state.
- Add per-card realistic and rules views, face switching, persistent inspection state, prompt defaults, and direct legal-card activation while keeping manual moves explicit.
[...]

PR #948, opened 2026-09-18
fix(ui): community loading skeletons and targeting field reveal

fedepoi:
## Summary

- Add stable skeleton sections for Community curated API responses, keep curated content mounted during initial publication loading, and show facet tag skeletons while filter metadata loads — removing the layout shift when late API data inserts above the grid.
- Reveal opponent fields that hold legal targets while a `chooseBoardTargets` prompt is open: a new `targetingFocusIds` focus ranks alongside combat focus in `BoardScene`, so a mid-combat trigger can target permanents on a non-fighting opponent's board (#951).

## Why

Community loaded the primary grid first, then inserted curated sections after three additional API calls completed, producing a visible shift on load.

The battlefield redesign's focused layout pins delimiters to combat participants once attackers exist, and hover refocusing cannot outrank combat focus. A trigger targeting a permanent on a non-participating opponent therefore rendered behind the collapsed veil — invisible and untappable (#951, regression surfaced by #873).

## Test plan

- `yarn tsc --noEmit`, eslint, and prettier clean on all changed files.
- No visual checks or smoke tests run, per request. Manual check for #951: four-player table, declare an attacker vs. opponent A, answer a forced `chooseBoardTargets` prompt with a legal target on opponent B — B's band should expand alongside A's and its card accept the tap-target; after the prompt resolves the delimiters ease back.

## Demo

Not included; visual verification was explicitly skipped.


khaliostr:
@fedepoi the body mentions #951 without a keyword, so it will not close on merge. Could you change "(#951)" to "Fixes #951"? Otherwise I will close it by hand after the merge.

khaliostr (dismissed):

khaliostr (approved):

PR #956, opened 2026-09-19
feat(ai): prompt semantics for manabot (kind, cost, affordable)

khaliostr:
## Closed, superseded by #970

#970 is the same change without the Forge hint lane, the rig, and the `detail`/`api` fields. The hint lane and rig live on `khaliostr/manabot-learned`.

PR #958, opened 2026-09-19
feat(ai): learn manabot's chooseAction scorer from forge's picks

khaliostr:
## Closed, superseded

What shipped from this work:
- #964 and #967: affordability, no attacks into losing blocks, sanity vetoes, land colour demand, card roles table.
- #970: prompt kinds (the clean half of #956).

What did not ship: the learned `chooseAction` scorer. At n=900 on fresh seeds it is worth one point over the rules (36% vs 35%), inside noise. Imitating Forge or humans does not beat Forge; learned combat lost 6 to 10 points. Numbers in the notes.

The branch stays for the bench tooling: Forge hint lane in the harness (`hints: true`, bench only), `captures-to-raw.py` / `manabot-featurize` / `manabot-train` (13.9k prod games, 1.25M human decisions), `capture-blunders.py`, and the blunder counters in `manabot-jvm-game.mjs`.


khaliostr:
guys this is experimental and i hardly follow the underlying ML stuff, mostly claude's work. if the results are decent and the performance compatible with what we needed might be worth a look from someone familiar with this kind of stuff

